PR #3, opened 2020-08-08
Use saved R_HOME value from libR-sys.

dasmoth:
(Depends on https://github.com/extendr/libR-sys/pull/2)

Increases the chances of things working out of the box by avoiding hard-coding a default value of R_HOME, and getting it from libR-sys instead.  Tested on Fedora and MacOS/homebrew.


andy-thomason:
Help with getting this running on many platforms is very welcome.

andy-thomason:
Merged... Lets see if travis is happy :)

dasmoth:
I suspect it won't be without deploying the updated libR-sys first.

PR #10, opened 2020-08-20
Compiling on Windows

CGMossa:
Compiling on windows doesn't work.. Attempting to fix that. 

Added language annotations to code-blocks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ features.
 
 Example:
 
-```
+```rust
 let v = c!(1, 2, 3);
 let l = list!(a=1, b=2);
 print!(paste0!("v=", v, " l=", l));
@@ -32,7 +32,7 @@ and panic-free execution.
 
 Example:
 
-```
+```rust
 let s = r!("hello");
 let i = r!(1);
 let r = r!(1.0);
@@ -42,7 +42,7 @@ Provide iterator support for creation and consumption of r vectors.
 
 Example:
 
-```
+```rust
 let res = (1..=100).iter().collect::<RObj>();
 for x in res.as_i32_slice() {
     println!("{}", x)?;
@@ -53,7 +53,7 @@ Provide a procedural macro to adapt Rust functions to R
 
 Example:
 
-```
+```rust
 #[export_function]
 fn fred(a: i32) -> i32 {
     a + 1
@@ -62,7 +62,7 @@ fn fred(a: i32) -> i32 {
 
 In R:
 
-```
+```r
 
 result <- .Call("fred", 1)
 


CGMossa:
I think this package by dirkschumacher should be used for inspiration
atleast: https://github.com/dirkschumacher/blake3
I had hoped that dirk had managed to put it on CRAN, but that does not seem
to be the case.

Claus Wilke halso have a package that uses rust, but isn't on CRAN either;
see https://clauswilke.com/sinab/#installation

We'll see how all of this turns out...

andy-thomason:
Those do look promising. There is an AppVeyor script in Dirk's package.

I know quite a few people (mostly at pharmas) who use R on windows, so this is quite important.

CGMossa:
It is also incredibly important to me.

I just have to wait for the weekend as I have a tremendous amount of rust
code to write for a paper.

Do we have a discord or similar to just share these small tidbits?

andy-thomason:
Discord channel now open:

https://discord.gg/gNegZu

clauswilke:
Hello, I only just saw this project and this issue/PR. Not sure what the status is, but I'm very excited about any additional effort bringing R and Rust together. Further development of my sinab package depends on this, and I'd rather not have to do everything by myself.

To compile on Windows, the installation instructions that come with sinab have worked for a number of people:
https://clauswilke.com/sinab/

I also have a working appveyor setup: https://github.com/clauswilke/sinab/blob/master/appveyor.yml

You need an appropriate `Makevars.win` file in your R package: https://github.com/clauswilke/sinab/blob/master/src/Makevars.win

To get a package containing Rust onto CRAN, you need to precompile the Windows library and download on the fly as you build the package on Windows on CRAN. See the discussion [here](https://github.com/r-rust/hellorust/issues/6#issuecomment-579770968) and see an example here:
https://github.com/cran/gifski/blob/master/src/Makevars.win
https://github.com/cran/gifski/blob/master/tools/winlibs.R

Note: The gifski package is entirely written in Rust and it's available on CRAN, so this is definitely possible: https://cran.r-project.org/web/packages/gifski/index.html

andy-thomason:
Hi @clauswilke I was hoping to avoid the two-stage **gifski** approach of building a
static library and linking to a DLL by using a Makefile directly, reducing the file count in
examples. Ultimately we should only need a single rust file in addition to the R
package skeleton. But for convenience we may need to do it. The current file count is 11
files.

I don't currently have a windows or mac machine to try a windows build
and hence this has stalled.

I am working (rather slowly at the moment) on a Rust package skeleton generator
that will do this and would appreciate some help.

If anyone can contribute a simple example (less than 20 files) of a windows build on Appveyor,
it would be useful and act as a starting point. Likewise for the Mac.

clauswilke:
Keep in mind though that the two-stage approach is only needed for CRAN submission, because their servers currently lack the cargo toolchain. Otherwise installation from source is pretty straightforward.

A minimal example with a working appveyor build is available here: https://github.com/r-rust/hellorust

One other issue I realized is that there are two distinct compilation scenarios and we need to be clear which one we're talking about: 1. Build an R package that calls Rust code. 2. Build a standalone Rust binary that calls R code. The second scenario is substantially more complex than the first. I'm only talking about the first at this time.

clauswilke:
@CGMossa I'm closing this PR since things now compile on Windows (as far as I understand). If you want to merge the changes to the README, I suggest you make a new PR with a corresponding title. There's other fixes the README needs (e.g., somewhere code and regular text blocks get switched), so such a PR would make a lot of sense.

PR #15, opened 2020-09-10
Sketch in a module and class interface similar to RCpp

andy-thomason:


andy-thomason:
This PR adds support for struct traits in extendr.

PR #20, opened 2020-09-30
Robj from String/&str, and arrays/vectors thereof

TianyiShi2001:
This PR makes it possible to create Robj from `String` or `&[String`.

For example, changing the `hello` example to this would work, which wouldn't work before:

```
#[extendr]
fn hello() -> String {
    "hello".to_owned()
}
```

### DIFF
--- a/extendr-api/src/robj.rs
+++ b/extendr-api/src/robj.rs
@@ -1369,6 +1369,19 @@ impl<'a> From<&'a [&str]> for Robj {
     }
 }
 
+/// Convert a string to an Robj string array object.
+impl From<String> for Robj {
+    fn from(val: String) -> Self {
+        Robj::from(&val as &str)
+    }
+}
+
+impl<'a> From<&'a [String]> for Robj {
+    fn from(vals: &'a [String]) -> Self {
+        Robj::from(vals.into_iter().map(|s| s as &str).collect::<Vec<&str>>())
+    }
+}
+
 /// Convert an integer slice to an integer object.
 impl<'a> From<&'a [i32]> for Robj {
     fn from(vals: &[i32]) -> Self {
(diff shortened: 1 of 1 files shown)


lebensterben:
what's the purpose of `impl<'a> From<&'a [String]> for Robj`?

Suppose we have a`Vec<String>`, it will be coerced to `&[String]`, so `Robj::from::<Vec<String>>` is also implemented.

But your implementation first constructs an Iterator, conversion String to slices, and then collect them into a new vector. 

So that is to say, it construct a new vector of slices out of a vector of string. But is this necessary?

Why not directly having an `Robj` containing `Vec<String>` instead of `Vec<&str>`?

TianyiShi2001:
I realized that I don't need to implement `From`s related to `String` separately from those related to `&str` in the first place. The `AsRef<str>` trait could make it possible to implement both at the same time. However, when I tried:

```rust
impl<S> From<S> for Robj
where
    S: AsRef<str>,
{
    fn from(val: S) -> Self {
	unsafe {
            let val = val.as_ref();
	    // ...
```

I got this error:

```
upstream crates may add a new impl of trait `std::convert::AsRef<str>` for type `*mut libR_sys::SEXPREC` in future versionsrustc(E0119)
```

TianyiShi2001:
Also, as for `&str`-related traits, currently only `From<&'a [&str]>`, i.e. borrowed slices, is implemented, but not its owned counterpart, so I cannot write:

```
#[extendr]
fn hello() -> [&'static str; 2] {
    ["hello", "world"]
}
```

or

```
#[extendr]
fn hello() -> Vec<&'static str> {
    vec!["hello", "world"]
}
```

for example.

andy-thomason:
There is now a "From" for vectors of string-like objects.

We should handle iterators too in returns and input parameters.

fn sum(a: Iterator<Item=f64>) -> f64 { a.sum() }

PR #21, opened 2020-10-01
Fix 'R/vectors' example to run.

dbdahl:


andy-thomason:
This shows that we need CI tests on the examples!

PR #25, opened 2020-10-31
Extendr cran package

andy-thomason:
This is the R code which will be published on CRAN.

It is based loosely on Rcpp to provide familiarity.


clauswilke:
Shouldn't this live in its own repo? All the standard R development toolchains (e.g., `remotes::install_github()`) will only work under this condition. I'd also recommend giving the R package a different name than the crate, otherwise you'll run into problems I think (if only for the fact that the crate and the package can't have their own repo). Maybe rextendr would work as the package name.

andy-thomason:
I'm happy to drop this PR in favour of rextendr.

PR #30, opened 2020-11-19
bump version to 0.1.5 for both extendr-api and extendr-macros.

clauswilke:
We need to bump the version of extendr-macros so we can update it on crates.io. I think it's best to keep the versions of extendr-api and extendr-macros the same at all times, so I've bumped them both to 0.1.5.


clauswilke:
now resolved.

PR #32, opened 2020-11-20
Add support for passing matrices via ArrayView2

hobofan:
I hope that's an acceptable implementation for matrices.

I tried to go a similar route as the conversion to ArrayView1 (via AsTypedSlice), but wasn't really to get something going via that route.

Looking forward to your feedback!

PS: Awesome crate! It's been a joy using it so far, even in it's early stage! :)

### DIFF
--- a/extendr-api/src/robj.rs
+++ b/extendr-api/src/robj.rs
@@ -149,6 +149,30 @@ where
     }
 }
 
+macro_rules! make_array_view_2 {
+    ($type: ty, $fn: tt, $error_str: tt, $($sexp: tt),* ) => {
+        impl<'a> FromRobj<'a> for ArrayView2<'a, $type> {
+            fn from_robj(robj: &'a Robj) -> Result<Self, &'static str> {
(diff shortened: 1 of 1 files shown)


clauswilke:
I have no particular opinion on the implementation (I'll let others comment on that), but I'd like to point out that it's always a good idea to add unit tests for each new feature that gets added.

hobofan:
Good point! Added some tests, similar to the ones of ArrayView1.

clauswilke:
Thanks for adding the tests! I would suggest adding one more test that tests something other than a 1x1 matrix. Maybe 2x3 with different values in the cells, to make sure there's no mixup with rows and columns.

hobofan:
Added a test for that :)

clauswilke:
Thanks!

clauswilke (approved):

andy-thomason (approved):
Excellent. Nice to see eval_string in action in the tests.

PR #33, opened 2020-11-21
Bump macros version to 0.1.3

andy-thomason:


clauswilke:
Andy, had you seen my PR #30? I don't have a super strong opinion on whether the extendr-api and extendr-macros versions should be kept in sync, but I'm quite certain that if we bump the extendr-macros version then we also need to bump extendr-api so that the new dependency makes it to crates.io. This means extendr-api needs to move to 0.1.5, extendr-macros can be either 0.1.3 or 0.1.5.

clauswilke:
now resolved.

PR #34, opened 2020-11-21
Add extendr module to hello example

andy-thomason:
With the new macro system, we can add modules to libraries.

This is currently in the classes example, but not in the hello example.


clauswilke:
Maybe turn it into a separate example for modules? I think it's good to have a minimal example that shows the absolute minimal requirements to use the functionality. By adding modules to the hello example, you'll give people the impression that they have to use modules to use this functionality.

Also, I note that PRs #34 and #35 overlap. Some changes in #34 apply to #35 and vice versa.

andy-thomason:
Dumping this as it is out of date.

PR #37, opened 2020-11-21
Rprint macros

### LINKED ISSUE #36: Add Rprintln macro
It would be great if there was an `Rprintln!()` macro that would function exactly like regular `println!()` but use `Rprintf()` to send the output to R. In addition, it might be useful to add `REprintln!()` to replace `eprintln!()` for errors.

Relevant documentation from Writing R Extensions: https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Printing

dasmoth:
Add a family of printing macros (resolves #36) 

There's an automated test in a separate commit.  I don't know how much value this is really adding, but was curious whether it was even possible, without resorting to extra helper scripts.  The answer is yes, but it's a little finicky.  Note that I had to remove a call to end_r elsewhere in the test suite because -- while R still mostly works after calling this -- its tmp directory no longer exists, so the fifo trick I used in the rprintln! test won't work.

### DIFF
--- a/extendr-api/src/lib.rs
+++ b/extendr-api/src/lib.rs
@@ -311,4 +311,16 @@ mod tests {
             assert_eq!(new_borrowed(wrap__return_f64()), Robj::from(123.));
         }
     }
+
+    #[test]
+    fn r_output_test() {
+        let fifo = lang!("fifo", Robj::from("")).eval().unwrap();
+        let fifo = unsafe { fifo.get() };
+        println!("{:?}", fifo);
+        lang!("sink", fifo).eval_blind();
+        rprintln!("Hello world");
+        lang!("sink").eval_blind();
+        let result : Robj = lang!("readLines", fifo).eval_blind();
+        assert_eq!(result, Robj::from("Hello world"));
+    }
 }
--- a/extendr-api/src/lib.rs
+++ b/extendr-api/src/lib.rs
@@ -83,6 +83,9 @@ pub use libR_sys::R_registerRoutines;
 pub use libR_sys::R_useDynamicSymbols;
 pub use libR_sys::SEXP;
 
+use libR_sys::*;
+use std::ffi::CString;
+
 /// Generic dynamic error type.
 pub type AnyError = Box<dyn std::error::Error + Send + Sync>;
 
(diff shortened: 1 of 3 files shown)


extendr-api/src/lib.rs
```diff
@@ -127,6 +130,20 @@ pub unsafe fn register_call_methods(info: *mut libR_sys::DllInfo, methods: &[Cal
 //     Robj::eval_string(rcode.as_str()).unwrap();
 // }
 
+pub fn print_r_output<T: Into<Vec<u8>>>(s: T) {
+    let cs = CString::new(s).expect("NulError");
```
clauswilke, line 134: Did you mean to write "NullError"? (two l's) See also line 141.

extendr-api/src/rmacros.rs
```diff
@@ -107,3 +107,53 @@ macro_rules! data_frame {
         lang!("data.frame", $($rest)*).eval_blind()
     };
 }
+
+/// Print via the R output stream.
+///
+/// Works like print! but integrates with R and respects
+/// redirection with functions like sink and capture.output
```
clauswilke, line 114: Tidyverse style puts parenthese after function names (`sink()`, `capture.output()`) and also puts them in backticks.

clauswilke: Oh, and similarly for `print!()` I think.

extendr-api/src/rmacros.rs
```diff
@@ -107,3 +107,53 @@ macro_rules! data_frame {
...
+    };
+    ($($rest: tt)*) => {
+        print_r_output(format!($($rest)*));
+    };
+}
+
+/// Print with a newline via the R output stream.
+///
+/// Works like println! but integrates with R and respects
+/// redirection with functions like sink and capture.output
```
clauswilke, line 127: See above regarding backticks and parentheses.

clauswilke (commented):
Thanks a lot! This looks good to me, but I don't fully understand the test issue with `end_r()` you mention. So Andy should take a look also.

I made a few minor comments regarding style.

extendr-api/src/lib.rs
dasmoth: Intentional, since it refers to the presence of ASCII NUL bytes.  Also, matches the usage in the underlying CString API.

clauswilke (approved):
Thanks! Since we're still in the early stages of this project, we should probably lean towards merging PRs quickly, even if there are some issues we're not entirely certain about, just to keep the project momentum going.

andy-thomason (approved):
Looks good. I was thinking of this while writing examples.

andy-thomason:
Out of interest.

end_r is a problem with tests as if you shut R down, it can't be restarted.

We still have problems with concurrent tests as R is strictly single threaded
and will crash if run concurrently in a single process.

dasmoth:
I think the "concurrent testing" issues are a little different, in that R is very non-thread-safe (global variables, I guess).

But even with a single thread running tests, calling end_r is problematic since I'm not sure there's a trustworthy way of knowing which tests are going to be run last.  In my hands, just exiting without calling `end_r` doesn't seem to cause any problems, but might be worth thinking about ways to call it at the end of the test run anyway.

I don't believe Rust has an official "call before exit" API, but I think we could install a bog standard atexit function...

Issue #31, opened 2020-11-19
cargo test fails extendr-api


clauswilke:
When trying to run `cargo test` for `extendr-api`, I get a segfault. Is this OS X specific or does the same problem show up on other platforms? I note that the "hello" example builds and runs fine.

```
clauswilke ~/github/extendr/extendr-api> cargo test
   Compiling extendr-api v0.1.4 (/Users/clauswilke/github/extendr/extendr-api)
    Finished test [unoptimized + debuginfo] target(s) in 1.76s
     Running /Users/clauswilke/github/extendr/target/debug/deps/extendr_api-6486cbe7629caf2f

running 7 tests
error: test failed, to rerun pass '--lib'

Caused by:
  process didn't exit successfully: `/Users/clauswilke/github/extendr/target/debug/deps/extendr_api-6486cbe7629caf2f` (signal: 11, SIGSEGV: invalid memory reference)
```

clauswilke:
Well, it seems to succeed on Travis: https://travis-ci.org/github/extendr/extendr/builds/744744955
So maybe OS X specific.

@CGMossa Does it work for you on Windows?

lebensterben:
can you try:

[profile.test]
codegen-units=1

clauswilke:
@lebensterben Makes no difference.

lebensterben:
Same here:
```
extendr/extendr-api ::  master is  v0.1.4 via  v1.50.0-nightly ~ 13s 
𝛌 cargo test
    Finished test [unoptimized + debuginfo] target(s) in 0.01s
     Running /tmp/extendr/target/debug/deps/extendr_api-8a21d25a9c80c630
/tmp/extendr/target/debug/deps/extendr_api-8a21d25a9c80c630: error while loading shared libraries: libR.so: cannot open shared object file: No such file or directory
error: test failed, to rerun pass '--lib'
```

lebensterben:
After fixing the libR.so missing problem, I get the same SIGSEGV error:
```
cargo test        
    Finished test [unoptimized + debuginfo] target(s) in 0.04s
     Running /tmp/extendr/target/debug/deps/extendr_api-8a21d25a9c80c630

running 7 tests
error: test failed, to rerun pass '--lib'

Caused by:
  process didn't exit successfully: `/tmp/extendr/target/debug/deps/extendr_api-8a21d25a9c80c630` (signal: 11, SIGSEGV: invalid memory reference)
```

clauswilke:
[...]

PR #40, opened 2020-11-23
Output R wrapper file from #[extendr] macros

hobofan:
Given some Rust code:
```rust
#[extendr]
fn hello() -> &'static str {
    \"Hello, this string definitely was created by Rust.\"
}

#[extendr]
fn test( a: &str, b: i64) {
    println!(\"Data sent to Rust: {}, {}\", a, b);
}
```
A wrapper R file containing the following will be generated:
```R
hello <- function() {
  .Call("wrap__hello")
}

test <- function(a, b) {
  invisible(.Call("wrap__test", a, b))
}
```
------

- In each `#[extendr]` macro invocation, the necessary information about the wrapped function is extracted, and written to a global mutex that keeps track of all the functions of previous macro invocations
- At the end of the macro invocation the wrapper code is written to `<TARGET_DIR>/extendr_wrappers.R`

There are some edge cases where the macros might not be reinvoked and stale wrapper code might persist, but for the two most common use cases (unknowing R library consumer, and extension development), those shouldn't really surface.


clauswilke:
Thanks! I hope @andy-thomason has time to review this. It's a bit beyond my Rust skills.

hobofan:
Oh yeah, there aren't really any tests for this right now. In my opinion those would be much better handled in some form of integration test, which I'd tackle after this (as I think #32 would also benefit from that).

extendr-macros/src/output_r.rs
```diff
@@ -118,6 +118,15 @@ fn write_r_wrapper(wrapper_fn: WrapperFn) -> Result<(), Box<dyn Error>> {
 
     let file = File::create(&output_file_path)?;
     let mut writer = BufWriter::new(file);
+    write!(
+        &mut writer,
+        r#"
+===========
+THIS FILE IS AUTOGENERATED VIA #[extendr] MACROS
+DO NOT ATTEMPT TO EDIT IT BY HAND
+===========
```
clauswilke, line 127: Could you add a `#` in front of these four lines? The generated file is meant to be read as R code and so this warning notice needs to be formatted as an R comment.

clauswilke:
A few comments:
1. Something seems to be going wrong in the unit tests. There are a bunch of error messages of the form `Unable to write R wrapper file to target directory`.
2. The autogeneration header will have to be formatted as  R comments (with a `#` at the beginning of the line).
3. Is it possible to provide arguments to the `#[extendr]` directive? I realized that for package development, we'll want to distinguish between package functions and exported functions. Exported functions require special decorations in the R code, e.g.:
```r
#' @rdname hello
#' @export
hello <- function() {
  .Call("wrap__hello")
}
```
And ideally the `@rdname` would be configurable, as it doesn't have to be the same as the function name. So we'd want to write maybe something like `#[extendr(export, hello)]` to generate output with the above decorations. (The `hello` in the `#[extendr]` attribute is the `@rdname`, not the function name.)

hobofan:
1. My first attempts to reproduce it locally failed, but I'll look into it
2. Fixed
3. Yes, macro attributes should be pretty straightforward. I'd leave that for a future PR, together with other things like passing the doc comments through to the wrapper file.

clauswilke:
Deferring macro attributes for later seems reasonable to me. If it is possible to capture and process the doc comments that would be even better. That’s how Rcpp works. See e.g. here:
https://r-pkgs.org/src.html#cpp-man

CGMossa:
I've just tried the workspace-target-folder fix, and I get this issue:

```


running 7 tests
error: test failed, to rerun pass '-p extendr-api --lib'

Caused by:
  process didn't exit successfully: `C:\Users\tpb398\Documents\GitHub\extendr\target\debug\deps\extendr_api-57c11e2b0c4694fa.exe` (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
```

hobofan:
@CGMossa Do you get that error consistently and really only with that commit? Everything I find online for the `STATUS_ACCESS_VIOLATION` is related to FFI, so my first hunch is that it could be something related to some flakey tests that do FFI.

CGMossa:
Right, I think this error is unrelated. I did try to go back before you did your thing, and tried `cargo test` I got this error as well.

I've tested this new commit with the rextendr part and it works...

CGMossa:
I think the access violation error is related to #31, as when I run this: `cargo test -- --nocapture --test-threads=1`, then everything works fine.

andy-thomason:
Hi @hobofan, I have been quite busy with work recently.

I discussed the use of global variables on stack overflow with the Rust language team and
they strongly discouraged me from writing proc macros with side effects.

Before committing to this approach you should be aware of the #[extendr_module] mechanism which collects functions for export.
Check the "classes" example for struct impl export which shows this.

The ultimate plan is to generate the function stubs using the module exporter. There was
some code written to do this - ie. generate a LANG expression for the function definition,
but it would probably be best done in the R package, which is a work in progress.

Rcpp, for example, extracts a list of function names from the Rcpp module definition
and builds the function stubs with R code of the form:

f <- function(...) .Call(wrap__f, ...)

The variadic operator ... copies the formal arg list to the actuals of the callee.

If we were to generate stubs as a text file, we could generate code which is called by
the module to output the R code to stdout. We would only need a single entrypoint
to generate the code:

.Call("__generate_R_bindings")

for example.

Then we can use

Rscript -e "library(mylib)" -e '.Call("__generate_R_bindings")'

as a post-build step.

I'll try to look more closely at your PR, time permitting.

clauswilke:
Andy, not sure if you've followed the work here: https://github.com/clauswilke/rextendr
A first working implementation of the R package is basically done, and it currently uses the approach you describe, with `f <- function(...) .Call(wrap__f, ...)`, and using a bunch of regular expressions to find `f`. This PR was intended to get rid of the regular expressions, but it's not critical. The current approach works.

hobofan:
> I discussed the use of global variables on stack overflow with the Rust language team and
they strongly discouraged me from writing proc macros with side effects.

Yes, I'm aware of the concerns there (and I've seen that SO thread). I've written quite a few proc_macro crates in the past (some of which have also written files into the target directory: [reacty_yew](https://github.com/hobofan/reacty_yew), [jnigen](https://github.com/hobofan/jnigen)), and I've tried out all kinds of shenanigans with them, so I think I have a good feeling for the limitations and problems that come with them. All-in-all, writing out a single file to a well-defined location, in a way that doesn't rely on invocation order of the macros or other implementation details is probably one of less risky side effects to take advantage of. 

> Before committing to this approach you should be aware of the #[extendr_module] mechanism which collects functions for export.
Check the "classes" example for struct impl export which shows this.

Ah good! I had seen `#[extendr_module]` in the source, but wasn't quite sure what it was for, and what R interface it expects. I'll take a look at that.

> If we were to generate stubs as a text file, we could generate code which is called by
the module to output the R code to stdout. We would only need a single entrypoint
to generate the code:
>
> .Call("__generate_R_bindings")

Yes, that is an alternative that I did consider, and we could possibly go with if we really want to avoid writing directly to the target directory. However such a system where code is generated in multiple places and then called by a single entry point is usually much more complex in my experience (I think that's how parts of PyO3 function via [inventory](https://crates.io/crates/inventory]).

> Rscript -e "library(mylib)" -e '.Call("__generate_R_bindings")'

Would we even need to generate R code if we have a single entry point? Could we not directly register the functions into the namespace via the useDynLib registration mechanic?

clauswilke:
> Would we even need to generate R code if we have a single entry point? Could we not directly register the functions into the namespace via the useDynLib registration mechanic?

Yes, we would still need R code, so the R user can write `hello()` instead of `.Call("hello")`.

My personal opinion is that calling the generated Rust library to spit out the R bindings creates a chicken-and-egg problem I don't like (e.g., we can't call `library()` until the bindings are in place), so I'd very much prefer not to do that. I think the options are either to have some Rust macro spit out the R code on compile time or use regular expressions from R to emulate Rust parsing.

clauswilke:
Thinking some more about this, I have to say I really like the user experience of what @hobofan has implemented here. So one perspective would be to merge this and see if we run into any actual problems in real-world usage. We can always disable this feature again and fall back to regular expressions.

andy-thomason:
I'm sorry to be annoying with this PR. You have done some sterling work
here.

Perhaps you could help with the following:

We have code in the #[extendr_module] generator to register
functions with R using the standard DLL mechanism.

[Here](https://github.com/extendr/extendr/blob/master/extendr-macros/src/lib.rs#L334-L354)

It would be nice to add functions to provide metadata, including doc strings, to a function that
could return them to R.

```
extern "C" fn __get_module_metadata() -> Robj {
  ...
  data_frame!(name=..., args=..., doc=...)
}
```

This way we could use the rextendr code to generate functions, NAMESPACE and documentation
for all the functions and methods.

For regular functions, name is just the name called by R.
For methods and traits, the name is qualified: `Fred::new()`

The doc string is available as an attribute of the function but will need
to be translated to the 

As you have started this, I would rather work with you on it than
implemented an alternative. I can provide guidance along the way.

andy-thomason:
Merging to get the work in...

clauswilke:
I've turned the request for future development into a new issue: #69

PR #44, opened 2020-11-25
Setup GH actions

hobofan:
This removes TravisCI and replaces it with GH actions for Linux, macOS and Windows.

There currently is an error during the Windows build, which is why it's commented out until we know how to resolve it. Since the build of `libR-sys` succeeded, I think that the general environment of LLVM for bindgen is set up correctly, though.

```
   --> extendr-api\src\robj.rs:341:67
    |
341 |             let parsed = Robj::from(R_ParseVector(code.get(), -1, status_ptr, R_NilValue));
    |                                                                   ^^^^^^^^^^ expected `i32`, found `u32`
    |
    = note: expected raw pointer `*mut i32`
               found raw pointer `*mut u32`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `extendr-api`.
```


.github/workflows/test.yml
```diff
@@ -0,0 +1,66 @@
+name: Tests
+
+on: [push]
```
clauswilke, line 3: You may also want to run on pull requests. See e.g. here: https://github.com/extendr/rextendr/blob/1ab743759698528a08fec7847df110ed9648249b/.github/workflows/R-CMD-check.yaml#L1-L9

.github/workflows/test.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: cargo test -- --nocapture --test-threads=1
+        env:
+          LIBCLANG_PATH: /usr/lib/llvm-3.9/lib
+
+  test_macos:
+    name: Run tests (macOS)
+    runs-on: macos-latest
+    strategy:
+      matrix:
+        rust-version: [stable, beta, nightly]
```
clauswilke, line 33: Not sure we need to run on all combinations of OSs and rust. Maybe only do stable on macOS and Windows?

.github/workflows/test.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Install dependencies
+        run: brew install llvm r
+      - name: Set dynamic environment variables
+        run: echo "LIBCLANG_PATH=$(brew --prefix llvm)/lib" >> $GITHUB_ENV
+      - name: Build
+        run: cargo build
+      - name: Run tests
+        run: cargo test -- --nocapture --test-threads=1
+
+  # test_windows:
```
clauswilke, line 48: On Windows, I think you need to change the target: `rustup target add x86_64-pc-windows-gnu`

clauswilke (commented):
Looks great! I left a few comments.

dasmoth:
Re: the Windows build... the actual type there should be `enum ParseStatus`.  I guess it's a platform difference whether C-ABI enums are considered to be signed or unsigned?  But since the generated bindings define a Rust `ParseStatus` type, perhaps it's possible to just use that instead, and make things portable that way.

clauswilke:
@dasmoth It does compile on Windows, though, just not with this particular CI setup. So even though the change you suggest would probably make the code more robust, it should also be possible to get things compiled as is. I suspect there's a compiler mismatch, the bindings are generated with one compiler and then cargo uses a different compiler to compiler the library.

@CGMossa has done a lot of digging into getting things compiled on Windows, maybe he can see where the problem is.

clauswilke:
According to the libR-sys installation instructions for Windows here: https://github.com/extendr/libR-sys#windows-specific-instructions, `cargo` needs to be run in a `mingw64` shell. Maybe that's the problem?

clauswilke:
@hobofan One more comment: With switching the CI runs over to GH Actions, you probably should also swap out the badge in the readme file. The new badge link should be something like:
```
[![build
status](https://github.com/extendr/extendr/workflows/test/badge.svg)](https://github.com/extendr/extendr/actions)
```

clauswilke:
@hobofan With your latest commit, does the windows build succeed? If not, I'd suggest you comment it out and turn on actions for pull requests and we merge. Then, going forward, it's easier to test changes to the action setup since it'll be run every time we push to a pull request.

CGMossa:
It is supposed to look like this:

```bash
msys2 -c "C:/Users/tpb398/.cargo/bin/cargo.exe build"
```
I've also used `mingw64` instead of `msys2` directly..

CGMossa:
Okay, we need to install R on Windows here. Either we use msys2 to install it, or do as is done [here](https://github.com/rstudio/shiny-testing-gha-example/blob/single_platform_snapshot/.github/workflows/run-tests.yaml).

hobofan:
So, installing via the `setup-r` action didn't work? Do you have pointer on how to install it with msys2?

hobofan:
@clauswilke I think it's mergeable now without windows (though not much seems to be missing there). I also adjusted the Linux nightly build to be allowed to fail.

andy-thomason:
This looks great. Many thanks to everyone who contributed to this.
I'm happy to merge if no one objects or thinks we should wait for the windows build.

PR #48, opened 2020-11-27
Added Windows to the GHA

CGMossa:
After 21 attempts, here it is. Tested mostly on libR-sys directly.

I've made several discoveries. I'll list them below shortly.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -89,10 +89,10 @@ jobs:
           cargo build
         env:
           LIBCLANG_PATH: C:/msys64/mingw64/bin
-      - name: Run tests
-        run: |
-          echo "C:/msys64/usr/bin" >> $GITHUB_PATH
-          echo "C:/msys64/mingw64/bin" >> $GITHUB_PATH
-          cargo test -- --nocapture --test-threads=1
-        env:
-          LIBCLANG_PATH: C:/msys64/mingw64/bin
+      # - name: Run tests
+      #   run: |
+      #     echo "C:/msys64/usr/bin" >> $GITHUB_PATH
+      #     echo "C:/msys64/mingw64/bin" >> $GITHUB_PATH
+      #     cargo test -- --nocapture --test-threads=1
+      #   env:
+      #     LIBCLANG_PATH: C:/msys64/mingw64/bin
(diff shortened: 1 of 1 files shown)


CGMossa:
Several notes:

* Rustup wasn't configured to have the target-toolchain being `stable-gnu`, thus it was stuck on the default `stable-msvc`. This is what primarily caused the missing `stdio.h`. 
* I don't think that changes to action path is persistent, thus changes to it is done each time.
* There is no need to update, or install msys2 from the ground up; as the default one contains all the tools we need.
* rustfmt and clippy were missing.. 
* msys2 is very specific to configure: inherit path as otherwise it doesn't know about `R`. Select `msystem` to be `mingw64`, otherwise it cannot open or understand the clang libraries..

CGMossa:
Rooms for improvement: It takes 4 mins to install `R`. That's very unreasonable.
Testing takes a long time comparison to building. But that's another story.

CGMossa:
This is building fine, but on windows `tests::r_output_test` is running indefinitely.

hobofan:
> Rustup wasn't configured to have the target-toolchain being stable-gnu, thus it was stuck on the default stable-msvc. This is what primarily caused the missing stdio.h.

Great find! I would have guessed that would default to the x86_64-pc-windows-gnu target on the stable-gnu toolchain, but according to https://github.com/rust-lang/rustup/issues/2568 that's the expected behaviour :/

.github/workflows/test.yml
```diff
@@ -51,29 +51,48 @@ jobs:
...
+          release: false
+          update: false
+          install: >-
+            mingw-w64-x86_64-clang
+            mingw-w64-x86_64-headers-git
+            mingw-w64-x86_64-make
+            mingw-w64-x86_64-pkg-config
+      - uses: r-lib/actions/setup-r@v1
+        with:
+          r-version: 'release'
```
clauswilke, line 78: Can you replicate the R setup you're using here (with `r-lib/actions/setup-r@v1`) to the other operating systems? I think it's preferable to just installing `r-base-dev` because it allows us to specify the R version we're testing. (And ultimately, we should have some tests on `release`, some on `devel`, and maybe some on the previous release.)

clauswilke:
I'm going to merge this since it works. We can address the R version issue separately.

PR #50, opened 2020-11-28
Extendr engine

### LINKED ISSUE #46: Compiled Rust library contains symbols R doesn't like
I have been trying to build an R package that can pass `R CMD check`: https://github.com/clauswilke/helloextendr

This doesn't currently work, because the compiled Rust library contains symbols that R doesn't like:

```
* checking compiled code ... NOTE
File ‘helloextendr/libs/helloextendr.so’:
  Found non-API calls to R: ‘R_CStackLimit’, ‘R_CleanTempDir’,
    ‘R_RunExitFinalizers’, ‘Rf_initialize_R’, ‘setup_Rmainloop’

See ‘Writing portable packages’ in the ‘Writing R Extensions’ manual.
```

I believe these are all functions that are needed to create a Rust binary calling R as a library. So we may need to take these out and put them into a separate crate, maybe called `extendr-standalone` or `extendr-bin`. Unfortunately we need these calls to run `cargo test`, so the tests would have to import that crate I assume.

clauswilke:
Create a separate crate `extendr-engine` that contains the function calls needed to call R from a standalone Rust binary. Closes #46.


clauswilke:
@andy-thomason @dasmoth @CGMossa @hobofan Could I get a few reviews of this PR? It's important for developing R packages with extendr, and it's sufficiently big of a change that I don't want to just merge.

clauswilke:
@andy-thomason I looked into making the engine an optional feature, but after some digging I don't think this is the right approach. There's [no way currently to optionally turn on features](https://github.com/rust-lang/cargo/issues/2911) whenever tests are compiled, so things get rather complex. (You can turn on features via command line, but not from the top of a workspace, so it's just a big mess.)

I propose to merge this PR as is (after fixing conflicts). Other features (such as ndarray support) can be made optional rather than a separate crate, but I think the engine belongs into its own crate.

andy-thomason (approved):

andy-thomason:
The tests are failing because the Bool wrapper is not available in this branch.

extendr-api/src/robj.rs
```diff
@@ -2086,8 +2086,7 @@ impl std::fmt::Debug for StrIter {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
```
andy-thomason, line 2089: We may need to retain this.

PR #52, opened 2020-11-29
Uncomment R_useDynamicSymbols and R_forceSymbols

yutannihilation:
Are there any contexts why these lines are commented out?

https://github.com/extendr/extendr/blob/5ebf0ab1dfc6e9574fe80526dca97ecf8377223b/extendr-api/src/lib.rs#L123-L124

`R_useDynamicSymbols()` needs to be called to avoid this warning:

```
File ‘helloextendr/libs/helloextendr.so’:
  Found no call to: ‘R_useDynamicSymbols’

It is good practice to register native routines and to disable symbol search.
```
(<https://github.com/clauswilke/helloextendr/runs/1466322075?check_suite_focus=true#step:9:89>)

Calling `R_forceSymbols()` is not mandatory (so Rcpp and cpp11 don't use this), but it's recommended to use in combination with `R_useDynamicSymbols()`, so I included this in this here. 

> the `R_forceSymbols` call only allows `.C` etc calls which specify entry points by R objects such as `C_pansari` (and not by character strings)
> 
> ...snip...
> 
> Finally, if `R_init_mypkg` also calls `R_forceSymbols(dll, TRUE)`, only `.Call(C_reg)` will work (and not `.Call("reg")`). This is usually what we want
(<https://cran.r-project.org/doc/manuals/r-patched/R-exts.html#Registering-native-routines>)

### DIFF
--- a/extendr-api/src/lib.rs
+++ b/extendr-api/src/lib.rs
@@ -120,8 +120,8 @@ pub unsafe fn register_call_methods(info: *mut libR_sys::DllInfo, methods: &[Cal
         std::ptr::null(),
         std::ptr::null(),
     );
-    //libR_sys::R_useDynamicSymbols(info, 0);
-    //libR_sys::R_forceSymbols(info, 1);
+    libR_sys::R_useDynamicSymbols(info, 0);
+    libR_sys::R_forceSymbols(info, 1);
 }
 
 // pub fn add_function_to_namespace(namespace: &str, fn_name: &str, wrap_name: &str) {


clauswilke:
This is a question for @andy-thomason, who I believe is still working on the modules interface. Note that this code path isn't even used in my helloextendr project at this time, because `register_call_methods()` is called from `#[extendr_module]`, not from `#[extendr]`, and I'm not using the modules interface at all yet.

https://github.com/extendr/extendr/blob/5ebf0ab1dfc6e9574fe80526dca97ecf8377223b/extendr-macros/src/lib.rs#L351

yutannihilation:
> Note that this code path isn't even used in my helloextendr project at this time

Oh, sorry. Adding this solved my local yet-another hello package copied from this repo, which uses `extendr_module!`, though.

https://github.com/extendr/extendr/blob/5ebf0ab1dfc6e9574fe80526dca97ecf8377223b/extendr-api/examples/R/hello/src/lib.rs#L8-L12

clauswilke:
Ok, it's good to know that this works in the context of `extendr_module!`. Maybe the concern earlier was (or still is) about calling these functions multiple times in case there are multiple macro calls that eventually call this function. Do you know what happens if these functions are called more than once on the same dll. Does that work? Also, could you try exporting a second module for the same package and seeing what that does?

yutannihilation:
I see, thanks for the context.

> Do you know what happens if these functions are called more than once on the same dll. Does that work?

I don't know.

> Also, could you try exporting a second module for the same package and seeing what that does?

It seems to work. After I duplicated `hello` as a different name, it still works and passes the checks.

https://github.com/yutannihilation/hello-extendr/commit/652b305310d52a01b914bc6c1ca2a66d2db63d12

clauswilke:
Ok, I'll go ahead and merge. If it turns out that this causes issues we can deal with them as we discover them.

yutannihilation:
Thanks. I found what `R_useDynamicSymbols()` is does is just setting a boolean, so I bet it won't be a problem no matter how many times we call it. Anyway, let's go ahead!

https://github.com/wch/r-source/blob/a48c6bea1353ce9b79beee0022a19ceeae822162/src/main/Rdynload.c#L250-L256

clauswilke:
Since merging this my package passes all tests, even though I'm not using the module macro.
https://github.com/clauswilke/helloextendr/runs/1476207707#step:9:85

PR #53, opened 2020-11-29
Deploy the dev version of docs to GitHub Pages

yutannihilation:
Since docs.rs serves only the published version, providing dev docs on GitHub Pages is useful. This is not what every Rust crate does, but many packages does e.g. [diesel](https://github.com/diesel-rs/diesel) and [servo](https://github.com/servo/servo). You can view the example here: https://yutannihilation.github.io/extendr/extendr_api/

Some things I'm not sure are:

* Where to place the link to the docs: GitHub Pages doesn't have the badge
* How to generate the index page nicely: it seems rustdoc already has this functionality, but it's not available in stable `cargo doc` (c.f. https://github.com/rust-lang/cargo/issues/8229#issuecomment-633153513). Should I generate the docs by using nightly?

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,12 +15,8 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
-      - name: Install dependencies
-        run: sudo apt-get install r-base-dev llvm-3.9-dev libclang-3.9-dev
       - name: Docs
         run: cargo doc --workspace --no-deps
-        env:
-          LIBCLANG_PATH: /usr/lib/llvm-3.9/lib
       - name: Deploy
         uses: peaceiris/actions-gh-pages@v3
         with:
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,28 @@
+name: Documents
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  docs:
+    name: Deploy the dev version of documents to GitHub Pages
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+      - name: Install dependencies
+        run: sudo apt-get install r-base-dev llvm-3.9-dev libclang-3.9-dev
+      - name: Docs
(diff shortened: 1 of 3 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,28 @@
...
+    name: Deploy the dev version of documents to GitHub Pages
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+      - name: Install dependencies
+        run: sudo apt-get install r-base-dev llvm-3.9-dev libclang-3.9-dev
```
clauswilke, line 19: Do you need the packages `llvm-3.9-dev` and `libclang-3.9-dev`? On ubuntu-20.04, I found I don't need them, and don't need to set `LIBCLANG_PATH` either: https://github.com/clauswilke/helloextendr/blob/main/.github/workflows/R-CMD-check.yaml

(Note that ubuntu-latest is currently 18.04 but will soon switch to 20.04: https://github.com/actions/virtual-environments/issues/1816.)

I think it's better to write action scripts that don't rely on specific versions of llvm/libclang, in particular if they're quite old.

clauswilke: Oh, and when ubuntu-latest switches over to 20.04, the script will break because llvm version 3.9 doesn't exist for 20.04.

yutannihilation: Thanks, I just copied these lines from from the existing YAML and didn't think about the details. Let's remove them.

yutannihilation:
Confirmed the test passes without installing additional system libraries and `LIBCLANG_PATH`.

https://github.com/yutannihilation/extendr/runs/1471710150

yutannihilation:
Ah, since the action on this repo is already setup in the era of 18.04, maybe I need to set `ubuntu-20.04` explicitly at least once?

clauswilke:
Thanks!

clauswilke:
Now that I have merged this I see that the link to the development documentation in the README is a bit ugly. Maybe removed it from between the badges and just write it somewhere in the main body of the text, in a sentence: "The documentation for the latest development version is available here: ..."

yutannihilation:
Ah, true. Sounds good to me. I'll prepare a pull request.

Issue #49, opened 2020-11-28
Different behavior regarding strings in data.frame between R 3.6.3 and R 4.0.3


jbabyhacker:
I am seeing strange behavior with data.frame between R 3.6.3 and R 4.0.3. I have reproduced this on both Ubuntu 20.04.1 and Mac OS 11.0.1. I am using extendr version 0.1.10.

In R 3.6.3 the following code prints out: [[1, 2, 3], [3.0, 4.0, 5.0], [6, 7, 8]]
In R 4.0.3 the following code prints out: [["abcd", "def", "fg"], [3.0, 4.0, 5.0], [6, 7, 8]]

I have spent a few hours at this. Any suggestions on where to look to debug this would be most grateful. 

```Rust
fn main() {
    start_r();
    let x_vec = &["abcd", "def", "fg"][..];
    let y_vec = &[3.0, 4.0, 5.0][..];
    let z_vec = &[6, 7, 8][..];

    let data_frame = data_frame!(x = x_vec.clone(), y = y_vec.clone(), z = z_vec.clone());

    println!("{:?}", data_frame);

    let c = data_frame.list_iter().unwrap().collect_vec();
    if let Some(v) = c[0].str_iter() {
        // Evaluated in R 4.0.3
        // v.collect_vec() is ["abcd", "def", "fg"]
        assert_eq!(v.collect_vec(), x_vec);
    } else {
        // c[0] evaluates to [1,2,3] and is of type i32
        // str_iter() fails in R 3.6.3
    }
    assert_eq!(c[1].as_f64_slice(), Some(y_vec));
    assert_eq!(c[2].as_i32_slice(), Some(z_vec));
    end_r();
}
```

clauswilke:
Are you using a libR-sys compiled for the specific version of R?

jbabyhacker:
Only one version of R was installed at a time. So it is just going to used whatever version is currently installed. I had version 3.6.3 installed, ran the test. Uninstalled it. Then installed 4.0.3, and ran the test.

clauswilke:
It's possible something isn't right in the FFI for R 3.6. That would be a question for @andy-thomason.

Did you run `cargo test`? Did it succeed with both R versions?

jbabyhacker:
[...]

PR #58, opened 2020-12-05
Issue #49: str_iter() for factors

andy-thomason:
This PR adds support for str_iter() in the case where the R object is a factor (ie. is an integer array that
indexes)

See https://www.stat.berkeley.edu/~s133/factors.html

This is a fix for:

https://github.com/extendr/extendr/issues/49

### DIFF
--- a/extendr-api/src/robj.rs
+++ b/extendr-api/src/robj.rs
@@ -318,13 +318,21 @@ impl Robj {
 
     /// Get an iterator over a string vector.
     pub fn str_iter(&self) -> Option<StrIter> {
+        let i = 0;
+        let len = self.len();
         match self.sexptype() {
             STRSXP => unsafe {
-                Some(StrIter {
-                    vector: self.get(),
-                    i: 0,
-                    len: self.len(),
-                })
+                let vector = self.get();
+                Some(StrIter {vector, i, len, levels: R_NilValue})
+            },
+            INTSXP => unsafe {
+                let vector = self.get();
+                let levels = self.getAttrib(&Robj::levelsSymbol());
+                if levels.sexptype() == STRSXP {
+                    Some(StrIter {vector, i, len, levels: levels.get()})
+                } else {
+                    None
+                }
             },
             _ => None,
         }
@@ -518,6 +526,25 @@ impl Robj {
             }
         }
         None
+    }
+
+    /// Get the class attribute as a string iterator if one exists.
+    pub fn class(&self) -> Option<StrIter> {
+        self.getAttrib(&Robj::classSymbol()).str_iter()
+    }
+
+    /// Return true if this class inherits this class.
+    pub fn inherits(&self, classname: &str) -> bool {
+        if let Some(mut iter) = self.class() {
+            iter.find(|&n| n == classname).is_some()
+        } else {
+            false
+        }
+    }
+
+    /// Get the levels attribute as a string iterator if one exists.
+    pub fn levels(&self) -> Option<StrIter> {
+        self.getAttrib(&Robj::levelsSymbol()).str_iter()
     }
 }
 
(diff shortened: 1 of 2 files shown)


extendr-api/src/robj.rs
```diff
@@ -318,13 +318,21 @@ impl Robj {
 
     /// Get an iterator over a string vector.
```
clauswilke, line 319: Maybe add here in the documentation that it also works for factors?

extendr-api/src/robj.rs
```diff
@@ -1686,12 +1714,27 @@ pub struct StrIter {
...
-        unsafe { Self { vector: R_NilValue, i: 0, len: 0 } }
+        unsafe { Self { vector: R_NilValue, i: 0, len: 0, levels: R_NilValue } }
+    }
+}
+
+// Get a string reference from a CHARSXP
+fn str_from_strsxp<'a>(sexp: SEXP, index: isize) -> &'a str {
+    unsafe {
+        if index < 0 || index >= Rf_xlength(sexp) {
+            ""
```
clauswilke, line 1731: Do we have the concept of a missing string (like `NA_character_` in R) in Rust? I'm a bit uncomfortable with turning a missing value into an empty string.

extendr-api/src/robj.rs
```diff
@@ -1868,4 +1916,29 @@ mod tests {
         assert_eq!(p, Robj::from(2));
         Ok(())
     }
+
+    #[test]
+    fn str_iter_test() -> Result<(), AnyError> {
+        start_r();
+        let obj = Robj::from(vec!["a", "b", "c"]);
+        assert_eq!(obj.str_iter().unwrap().collect::<Vec<_>>(), vec!["a", "b", "c"]);
+
```
clauswilke, line 1925: It would be good to also add a test that explicitly tests conversion from a factor. Note that the dataframe example uses factors for R 3.x but not for 4.x, so going forward it doesn't test the factor case.

clauswilke (commented):
I think all of this is going to be super useful. I just pointed out a few places where you could make some improvements.

extendr-api/src/robj.rs
andy-thomason: The correct way to do this is to use an `Option<&str>`. But as the index check should never fail with correctly formed factors, this would be a burden to users. Instead I chose to return something unobtrusive for broken factors.

extendr-api/src/robj.rs
andy-thomason: Added...

extendr-api/src/robj.rs
clauswilke: I understand your concern, but I think we need the `Option` anyways, to properly handle missing values (see #63). I believe the index lookup will fail for missing values, since they are represented as the negative largest integer (if I remember correctly).

yutannihilation:
Sorry for chiming in a bit late here, but I wonder if there are already discussions about whether extendr should support factor or not. It seems either Rcpp or cpp11 support factors, so it's user's responsibility to handle it there (I'm not a serious Rcpp/cpp11 user, so I might be wrong). Of course, we don't need to follow exactly what they do, but, if it would complicate the implementations considerably, it might be a choice not to support factors.

andy-thomason:
I'm happy to try to support as much of the R way of working as possible.

The more advantages we have the better :smile:

clauswilke:
Yes, I see no reason why extendr shouldn't support factors, and I like the way this PR handles them. (Modulo the missing values issue, but we can leave that for later.)

clauswilke (approved):

andy-thomason:
I've added a factor! macro that wraps the R function.

We could, in theory, have factors of other types, but strings are the most common.

This maps to the dictionaries in Arrow.

andy-thomason:
For now, I could return "NA" for out of bound values as this will avoid a breaking change.

If we are happy with this I'll merge it.

clauswilke:
Yes, please go ahead and merge. Let's have the `NA` discussion separately.

yutannihilation:
OK, thanks for clarification! Then I have no concern here.

andy-thomason:
I'll set up a more general issue for Options in iterators.

PR #61, opened 2020-12-05
Implement input converters (from_robj) for Iterators

andy-thomason:
This PR allows you to use iterators as inputs to exported functions.

eg.
```
#[extendr]
fn my_func(a: NumericIter, b: IntegerIter, c:StrIter) {
   let atot = a.sum();
   let btot = b.sum();
   let ctot = c.map(|s| s.parse::<i32>()).sum();
   ...
}
```

### DIFF
--- a/extendr-api/src/robj.rs
+++ b/extendr-api/src/robj.rs
@@ -134,6 +134,26 @@ impl<'a> FromRobj<'a> for Vec<f64> {
     }
 }
 
+macro_rules! impl_iter_from_robj {
+    ($t: ty, $iter_fn: ident) => {
+        impl<'a> FromRobj<'a> for $t {
+            fn from_robj(robj: &'a Robj) -> Result<Self, &'static str> {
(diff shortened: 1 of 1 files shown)


extendr-api/src/robj.rs
```diff
@@ -134,6 +134,26 @@ impl<'a> FromRobj<'a> for Vec<f64> {
...
 }
 
+macro_rules! impl_iter_from_robj {
+    ($t: ty, $iter_fn: ident) => {
+        impl<'a> FromRobj<'a> for $t {
+            fn from_robj(robj: &'a Robj) -> Result<Self, &'static str> {
+                if let Some(v) = robj.$iter_fn() {
+                    Ok(v)
+                } else {
+                    Err("not a vector of strings")
```
clauswilke, line 144: Is this error message correct? Shouldn't it depend on the type for which you're generating an iterator?

clauswilke (approved):
Looks good to me, except the one issue I flagged with the error message.

PR #66, opened 2020-12-08
Return an error when length > 1 of vector is passed to a scaler argument

### LINKED ISSUE #64: Stricter input checking for extendr functions
Related to #63, I think conversion from `Robj` to Rust native types should be stricter and only succeed when the coercion is unique and provably correct. As an example, vector input should not be silently converted into a single value. See reprex for an example. If input is `f64`, not `&[f64]`, then only vectors of length 1 should be allowed as input. 

    library(rextendr)

    rust_function(
      code = "fn one_float(input: f64) -> f64 {
        5.*input
      }",
      patch.crates_io = c(
        'extendr-api = { git = "https://github.com/extendr/extendr" }',
        'extendr-macros = { git = "https://github.com/extendr/extendr" }'
      )
    )

    # correct
    one_float(numeric(0))
    #> Error in one_float(numeric(0)): zero length vector

    # correct
    one_float(numeric(1))
    #> [1] 0

    # incorrect, should complain about vector of length > 1
    one_float(numeric(2))
    #> [1] 0

<sup>Created on 2020-12-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

yutannihilation:
Fix some part of #64

``` r
library(rextendr)

rust_function(
  code = "fn one_float(input: f64) -> f64 {
    5.*input
  }",
  patch.crates_io = c(
    'extendr-api = { path = "/home/yutani/repo/extendr/extendr-api" }',
    'extendr-macros = { git = "https://github.com/extendr/extendr" }'
  )
)
#> build directory: /tmp/RtmpVtIB3g/file7a70ee5a29f
#> Prebuild libR bindings are not available. Run `install_libR_bindings()` to improve future build times.

one_float(numeric(0))
#> Error in one_float(numeric(0)): zero length vector
one_float(numeric(1))
#> [1] 0
one_float(numeric(2))
#> Error in one_float(numeric(2)): >1 length vector
```

<sup>Created on 2020-12-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

(I'm yet to figure out how to write tests for this...)

### DIFF
--- a/extendr-api/src/robj.rs
+++ b/extendr-api/src/robj.rs
@@ -1834,6 +1834,11 @@ mod tests {
 
         let hello = Robj::from("hello");
         assert_eq!(<&str>::from_robj(&hello), Ok("hello"));
+
+        // conversion from a vector to a scalar value
+        assert_eq!(<i32>::from_robj(&Robj::from(vec![].as_slice() as &[i32])), Err("zero length vector"));
+        assert_eq!(<i32>::from_robj(&Robj::from(vec![1].as_slice() as &[i32])), Ok(1));
+        assert_eq!(<i32>::from_robj(&Robj::from(vec![1, 2].as_slice() as &[i32])), Err(">1 length vector"));
     }
     #[test]
     fn test_to_robj() {
--- a/extendr-api/src/robj.rs
+++ b/extendr-api/src/robj.rs
@@ -65,15 +65,15 @@ macro_rules! impl_prim_from_robj {
             fn from_robj(robj: &'a Robj) -> Result<Self, &'static str> {
                 if let Some(v) = robj.as_i32_slice() {
                     match v.len() {
-                        0 => Err("zero length vector"),
+                        0 => Err("Input must be of length 1. Vector of length zero given."),
                         1 => Ok(v[0] as Self),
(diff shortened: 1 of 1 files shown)


clauswilke:
Errors that are sent back to R probably have to be tested from R. This could eventually be part of a test suite that we put into the rextendr package.

yutannihilation:
> have to be tested from R

Makes sense, thanks. I included only minimum tests here, but let's test thoroughly from R in future.

clauswilke:
Maybe we can improve the error messages a bit? How about:
"Vector of length zero given. Expected length 1."
"Vector of length >1 given. Expected length 1."

yutannihilation:
I agree it's nice to improve the error messages. To follow the tidyverse style (c.f. https://style.tidyverse.org/error-messages.html#problem-statement), how about placing the problem state first:

"Input must be length 1. Vector of length ... given."

clauswilke:
Sounds good to me. I suggest one minor edit: "Input must be *of* length 1."

yutannihilation:
Thanks, I'll use it.

clauswilke (approved):

PR #74, opened 2020-12-12
Add bool and HashMap FromRobj impls.

andy-thomason:
This PR adds more input options to extendr functions.
```
// Surprisingly, this one was not there.
fn takes_bool(val: bool) {
}

// Take list(a=1, b=2) and expose as an inexpensive HashMap
fn takes_named_list(val: HashMap<&str, Robj>) {
   let a = val["a"];
   let b = val["b"];
}

// Take list(a=1, b=2) and expose as a long-lived HashMap
fn takes_named_list(val: HashMap<String, Robj>) {
}
```

### DIFF
--- a/extendr-api/src/lib.rs
+++ b/extendr-api/src/lib.rs
@@ -149,9 +149,6 @@ mod tests {
     use super::*;
     use crate as extendr_api;
 
-    // Allow macros in the tests.
-    use crate::*;
-
     use extendr_macros::extendr;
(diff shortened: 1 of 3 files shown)


andy-thomason:
We may need to check that the Robjs in the HashMap<String, Robj> are Owned
or we may end up with undefined behavior.

clauswilke (approved):

PR #77, opened 2020-12-17
Handling of missing values.

andy-thomason:
This PR provides input and output handling for scalar Option<value>.

None represents NA and Some(value) is a regular value.

### DIFF
--- a/extendr-api/src/robj.rs
+++ b/extendr-api/src/robj.rs
@@ -236,6 +236,71 @@ impl<'a> FromRobj<'a> for HashMap<&str, Robj> {
     }
 }
 
+// NA-sensitive integer input handling
+impl<'a> FromRobj<'a> for Option<i32> {
+    fn from_robj(robj: &'a Robj) -> Result<Self, &'static str> {
+        if robj.is_na() {
+            Ok(None)
+        } else if let Some(val) = robj.as_integer() {
+            Ok(Some(val))
+        } else {
+            Err("expected an integer scalar")
+        }
+    }
+}
+
+// NA-sensitive logical input handling
+impl<'a> FromRobj<'a> for Option<bool> {
+    fn from_robj(robj: &'a Robj) -> Result<Self, &'static str> {
+        if robj.is_na() {
+            Ok(None)
+        } else if let Some(val) = robj.as_logical() {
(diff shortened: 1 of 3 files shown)


extendr-api/src/robj.rs
```diff
@@ -1904,6 +1783,20 @@ impl ToVectorValue for &f64 {
...
+impl ToVectorValue for Option<f64> {
+    fn sexptype() -> SEXPTYPE {
+        REALSXP
+    }
+
+    fn to_numeric(&self) -> f64 {
+        if self.is_some() {
+            self.unwrap()
+        } else {
+            unsafe { R_NaReal }
```
clauswilke, line 1795: Here and throughout: I'm not sure this conversion is idiomatic. There is no valid Rust `f64` representation of `NA`. Sure we can return `R_NaReal`, but the calculations with it will be wrong. (This is even worse for integers.) I think the right thing to do is to panic, which then gets caught and converted into an R error. Sure this may feel inconvenient, but it means people have to either properly handle options throughout or accept that their code can't handle `NA`s. That's the only way to ensure results are always correct, in my opinion.

andy-thomason: You are right that calling a rust function with a NA should cause an error.

(rust)
```rust
#[extendr]
my_func(x: f64) {
}
```
with (R)
```R
my_func(NA)
```
Should cause an error.

But this particular code converts an Option<> into a NA if the option is None
and is intended for a return type where you may want to signal one or more NAs.

(rust)
```rust
#[extendr]
one_or_na(x: bool) -> Option<f64> {
  if x {
     Some(1.)
  } else {
     None
  }
}
```
with (R)
```R
my_func(TRUE) -> 1
my_func(FALSE) -> NA
```

It should also work for vectors and iterators containing NAs signaled by options.

(rust)
```rust
#[extendr]
ones_or_nas(x: &[bool]) -> Robj {
  x.into_iter()
    .map(|x| if x { Some(1.) } else { None })
    .collect_robj()
}
```
with (R)
```R
my_func(c(TRUE, FALSE, TRUE)) -> [1, NA, 1]
```

clauswilke: Got it, thanks!

PR #78, opened 2020-12-18
Integration tests

clauswilke:
This is trying to set up an R package that is checked each time GH actions are run.


andy-thomason:
I'm not sure what is causing the rust-nightly failure.

I'm running with 1.50.0-nightly without problems.

We would expect nightly to fail from time-to-time depending on the snapshot.

clauswilke:
@andy-thomason Turns out the problem is R devel, not Rust nightly:
https://github.com/extendr/extendr/runs/1578377705?check_suite_focus=true

clauswilke:
The integration tests work now, but have revealed a problem with extendr on R devel. I'm going to merge them because they're doing their job.

PR #82, opened 2020-12-20
Replace the releasing workflow with cargo-workspaces

### LINKED ISSUE #70: Fix bump.sh and publish.sh for new crate structure
I have added a new crate extendr-engine (#50) to remove the engine part from the default api. However, the scripts `bump.sh` and `publish.sh` don't reflect this change yet.

yutannihilation:
As discussed on #70, I drafted a workflow of releasing.

After some investigation, I found `cargo ws publish` does all the things (because it runs `cargo ws version` before publishing); that is

1. Bump version
2. Commit
3. Tag
4. Push
5. Publish

But, I'm wondering if this is the order we want (Sorry, I just don't know the usual manner of Rust). The steps I wrote here is slightly different:

1. Bump version
2. Commit
3. Tag
4. *Publish*
5. *Push*

I want to hear your thoughts! @clauswilke @andy-thomason 

Fix #70


clauswilke:
This looks good to me in principle. I do think though that the `git push` should come after the version increase, in particular because we will sometimes want to change the version but not publish, e.g. when we change the version to *-alpha.

clauswilke:
I think in the R world we normally wait to tag until a package is published because CRAN likes to reject submissions, but crates.io doesn't have such restrictive acceptance criteria and publishing is unlikely to fail.

yutannihilation:
> e.g. when we change the version to *-alpha.

Ah, I was about to reply to your comment https://github.com/extendr/extendr/issues/70#issuecomment-742709861. You might notice or not, the versioning mechanism is very different from that of R. IIUC, `version` specified with `git` or `path` matters only on the published version of crates and it's ignored elsewhere, according to [the doc](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies):

>  However, crates that use dependencies specified with only a path are not permitted on crates.io. If we wanted to publish our `hello_world` crate, we would need to publish a version of `hello_utils` to crates.io and specify its version in the dependencies line as well:

So, in my understanding, we don't need such a thing as "development version" on Rust, because `cargo update` never mixes dependencies specified with `version` and the ones with `path`. (In contrast, R's dependency mechanism has only version, so we need to be careful not to exceed the future versions)

clauswilke:
I'm pretty sure that we need development versions, because during development we frequently pull in crates via github, and if the version isn't right cargo will complain. For example, if we add a new feature relative to the published crate we should bump the minor version on github, at least to *-alpha, so that other crates can specify that they need a version that has that feature.

yutannihilation:
Hmm, okay, got it. Then, the steps would be

1. Bump version to the release version
2. Commit
3. Tag
4. Push
5. Publish
6. Bump version to the development version
7. Commit
8. Tag
9. Push

And if we don't publish, we do only step 6~9.

clauswilke:
Yes. What I've been struggling with, though, is how frequently do you do steps 6-9 and what development version do you use? If you're currently at 0.1.11, do you go to 0.1.12-alpha, but then if you do major changes you go to 0.2.0-alpha without ever publishing 0.1.12?

clauswilke:
Maybe the alternative is to do steps 6-9 only in preparation of the next release, once we know we have something that can be thought of as the next release candidate.

yutannihilation:
Honestly, I doubt we need to strictly follow the semantic versioning on GitHub. I feel those who depends on the git dependency should specify commit hash, rather than version, and they probably need to communicate with us directly rather than communicating by the version number anyway, if the recent changes are important for them. But this might be just I know very little about Rust's ecosystem.

clauswilke (approved):
Looks good, thanks! Can we delete the two bash scripts that did this job previously?

yutannihilation:
> Can we delete the two bash scripts

Yes, I included the deletion of them.

yutannihilation:
Though I believe this pull request is already ready to merge, I now come to think this can be done automatically by GitHub Actions CI (`custom` allows us to specify arbitrary version name):

``` sh
cargo ws version --force='*' --no-individual-tags custom <release name>
```

But, I cannot come up with what to use for triggering this. If I create a release, it means a new tag is created for it, which means GitHub Actions needs to overwrite the existing tag. Probably it works, but it doesn't look a clean way to me...

clauswilke:
I think doing this manually is fine.

yutannihilation:
I see.

@andy-thomason 
Could you take a look when you have time? Is the description of the workflow clear to you? If this looks OK, I'll merge this.

https://github.com/yutannihilation/extendr/tree/doc/issue-70-cargo-workspace#publishing-the-crate

andy-thomason:
Looks good to me. Go ahead an merge.

The only niggle is that the README is getting less user-friendly.

It may be better to make a PUBLISHING.md or similar and leave the README for users.

yutannihilation:
Thanks, agreed! I'll create `MAINTAINERS_GUIDE.md`.

andy-thomason:
:+1:

clauswilke:
@yutannihilation One issue I just noticed when following these guidelines for publishing libR-sys: If I understand correctly, `git push --tags` pushes the tags but not the commit that changed the version info. So we need a separate `git push` also.

yutannihilation:
Oh, good catch... I'll fit it.

PR #84, opened 2020-12-20
Make ndarray optional

### LINKED ISSUE #72: Make ndarray optional
Thought I'd open up a separate issue for this, as it has come up in comments repeatedly (and is in the Roadmap of the README).

This would probably a good first step towards feature flags (which is also a concern in #60), which I guess also would need to get some support in `rextendr`.

yutannihilation:
An attempt to fix #72.

This pull request separates the ndarray-related implementations (i.e. matrix) to a separate module and use it optionally.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -147,15 +147,15 @@ jobs:
       
       - name: Build
         run: |
-          cargo build --manifest-path extendr-api/Cargo.toml --features libR-sys/use-bindgen --features ndarray
-          cargo build --manifest-path extendr-engine/Cargo.toml --features libR-sys/use-bindgen
+          cargo build --manifest-path extendr-api/Cargo.toml --features tests-all
+          cargo build --manifest-path extendr-engine/Cargo.toml --features tests-all
 
       - name: Run tests
         run: |
           export R_HOME=`R -s -e 'cat(R.home())'`
           export LD_LIBRARY_PATH=${R_HOME}/lib
-          cargo test --manifest-path extendr-engine/Cargo.toml --features libR-sys/use-bindgen -- --nocapture --test-threads=1
-          cargo test --manifest-path extendr-api/Cargo.toml --features libR-sys/use-bindgen -- --nocapture --test-threads=1
-          cargo test --manifest-path extendr-api/Cargo.toml --features libR-sys/use-bindgen --features ndarray -- --nocapture --test-threads=1
+          cargo test --manifest-path extendr-engine/Cargo.toml --features tests-all -- --nocapture --test-threads=1
+          cargo test --manifest-path extendr-api/Cargo.toml --features tests-all -- --nocapture --test-threads=1
+          cargo test --manifest-path extendr-api/Cargo.toml --features tests-wo-ndarray -- --nocapture --test-threads=1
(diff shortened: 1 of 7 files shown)


.github/workflows/test.yml
```diff
@@ -146,5 +146,6 @@ jobs:
           export LD_LIBRARY_PATH=${R_HOME}/lib
           cargo test --manifest-path extendr-engine/Cargo.toml --features libR-sys/use-bindgen -- --nocapture --test-threads=1
           cargo test --manifest-path extendr-api/Cargo.toml --features libR-sys/use-bindgen -- --nocapture --test-threads=1
+          cargo test --manifest-path extendr-api/Cargo.toml --features libR-sys/use-bindgen --features ndarray -- --nocapture --test-threads=1
```
clauswilke, line 149: I've been wondering whether we should add a feature `tests-all` and add `ndarray` and `libR-sys/use-bindgen` (if possible, not sure if it works for dependent libs) to that feature. Then the test code never needs to change as we add new features, and we also don't get a super long argument as we add a larger set of features.

clauswilke: To be clear: I would add the `tests-all` feature to all three crates, so the tests for extendr-api, extendr-engine, etc look all the same, even if those crates have different actual features they offer.

clauswilke (approved):

andy-thomason (approved):

.github/workflows/test.yml
yutannihilation: Thanks for the idea, I added `tests-all`. I found we also need another flavor for testing without ndarray. I'm not sure if this is more handy than listing individual features. What do you think?

extendr-api/Cargo.toml
```diff
@@ -12,5 +12,13 @@ repository = "https://github.com/extendr/extendr"
...
+ndarray = { version = "0.13.1", optional = true }
 
+[features]
+default = []
+
+# All features needed for testing
+tests-all = ["ndarray", "libR-sys/use-bindgen"]
+
+# All features needed for testing, minus ndarray
+tests-wo-ndarray = ["libR-sys/use-bindgen"]
```
clauswilke, line 24: I think the questions is how we envision this in the future, when there are multiple optional modules. Presumably we wouldn't test all possible combinations of presence and absence but simply all optional modules present and all absent. This would suggest something like `tests-minimal`.

Alternatively, it may be a good idea to add a `use-bindgen` feature so people can turn it on without having to know that it is actually a feature of libR-sys.

.github/workflows/test.yml
clauswilke: See my other comment.

extendr-api/Cargo.toml
yutannihilation: Makes sense. Let's use tests-minimal.

andy-thomason: Are we good to go with this merge?

yutannihilation: Yes, I believe this is ready. What I was a bit wondering is whether I should also add this, 

> Alternatively, it may be a good idea to add a `use-bindgen` feature so people can turn it on without having to know that it is actually a feature of libR-sys.

but let's discuss it when it's needed. Let's merge this.

Issue #67, opened 2020-12-08
Remove/update r_output_test for Windows compatibility


Ilia-Kosenkov:
I stumbled upon this issue when trying to run tests on Windows.
The tests can actually be executed once linked against dev version of `libR-sys`, including PR [\#12](https://github.com/extendr/libR-sys/pull/12) in that repo.

I observed that the last out of 8 tests, `r_output_test`, simply freezes while being executed.
This was confirmed at least on two separate Windows machines.
I was able to pinpoint this issue to bugged `fifo` function in `R` (documentation [says](https://stat.ethz.ch/R-manual/R-devel/library/base/html/connections.html) it is supported under Windows, yet it does not work as expected) .

<details>
<summary>
R repro
</summary>

```R
fi <- fifo("")
sink(fi)
cat("Hello world!") # This call freezes R process
sink()
```

</details>

So this test can be either removed or replaced with something like this:
<details>
<summary>
r_output_test
</summary>

```rust
 #[test]
    fn r_output_test() {
        let con_target = "test_con";
        let txt_con = lang!("textConnection", Robj::from(con_target) , open = Robj::from("w"))
            .eval().unwrap();
        let con = unsafe { txt_con.get() };
        lang!("sink", txt_con).eval_blind();
        rprintln!("Hello world");
        lang!("sink").eval_blind();
        let result : Robj = lang!("get", Robj::from(con_target)).eval_blind();
        assert_eq!(result, Robj::from("Hello world"));
        lang!("close", con).eval_blind();
    }
```

</details>

Ilia-Kosenkov:
<details>
<summary>
(cargo test  -- --nocapture --test-threads=1) *>&1 > test.log
</summary>

```
   Compiling winapi-x86_64-pc-windows-gnu v0.4.0
   Compiling winapi v0.3.9
   Compiling proc-macro2 v1.0.24
   Compiling memchr v2.3.4
   Compiling unicode-xid v0.2.1
   Compiling autocfg v1.0.1
   Compiling cc v1.0.66
   Compiling version_check v0.9.2
   Compiling glob v0.3.0
   Compiling libc v0.2.81
   Compiling lazy_static v1.4.0
   Compiling log v0.4.11
   Compiling bitflags v1.2.1
   Compiling unicode-width v0.1.8
   Compiling cfg-if v0.1.10
   Compiling quick-error v1.2.3
   Compiling regex-syntax v0.6.21
   Compiling strsim v0.8.0
[...]

PR #86, opened 2020-12-21
Replaced 'fifo' with 'textConnection' in 'r_ouput_test'

Ilia-Kosenkov:
Addresses an issue (#67) with `fifo` on `Windows`, which appears to hang indefinitely.
`fifo` is replaced with `textConnection`, but the overall logic of the test is preserved.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -61,6 +61,15 @@ jobs:
           export LD_LIBRARY_PATH=${R_HOME}/lib
           cargo test -- --nocapture --test-threads=1
 
+      # Installing 'msys2' on Windows only
+      - uses: msys2/setup-msys2@v2
+        if: runner.os == 'Windows'
+        with:
+          msystem: MINGW64
+          path-type: inherit
+          release: false
+          update: false
+
       - name: Build (Windows)
         if: runner.os == 'Windows'
         run: cargo build --target x86_64-pc-windows-gnu
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -75,16 +75,12 @@ jobs:
       - name: Build (Windows)
         if: runner.os == 'Windows'
         run: cargo build --target x86_64-pc-windows-gnu
-        env:
-          LIBCLANG_PATH: C:\msys64\mingw64\bin
-        shell: msys2
-      
+        shell: msys2 {0}
+
       - name: Run tests (Windows)
         if: runner.os == 'Windows'
         run: cargo test --target x86_64-pc-windows-gnu -- --nocapture --test-threads=1
-        env:
-          LIBCLANG_PATH: C:\msys64\mingw64\bin
-        shell: msys2
+        shell: msys2 {0}
 
       - name: Query dependencies for integration testing
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,7 +53,7 @@ jobs:
           path-type: inherit
           release: false
           update: false
-          
+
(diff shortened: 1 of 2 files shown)


Ilia-Kosenkov:
Tests working fine on `Windows` if `PATH` is amended before execution (see [`extendr/libR-sys`](https://github.com/extendr/libR-sys) desctiption for details) 

<details>
<summary>Test on Windows</summary>

- **x64**
    ```PowerShell
    $env:PATH="$env:R_HOME\bin\x64;$env:PATH" ; 
    cargo +stable-x86_64-pc-windows-msvc test --target=x86_64-pc-windows-gnu -- --nocapture --test-threads=1
    ```
    
- **x32**
    ```PowerShell
    $env:PATH="$env:R_HOME\bin\i386;$env:PATH" ; 
    cargo +stable-i686-pc-windows-msvc test --target=i686-pc-windows-gnu -- --nocapture --test-threads=1
    ```

</details>

clauswilke:
Thanks, looks great! Could you turn on the tests in the GH Action script also?

https://github.com/extendr/extendr/blob/d27585f43f900c49f7a805be6ce3cd83a196ab6d/.github/workflows/test.yml#L68-L73

clauswilke:
For some reason the Windows tests fail with a linker error even though the build succeeds. Looks like it's picking up the wrong toolchain again. This is so frustrating.

Ilia-Kosenkov:
@clauswilke,
I think I see the problem -- it links to `mingw` from `rtools` because we are missing `msys2` for windows.
Is it possible to add `uses` with an `if` condition?

.github/workflows/test.yml
```diff
@@ -61,22 +71,18 @@ jobs:
...
-          path-type: inherit
-          release: false
-          update: false
 
       - name: Build (Windows)
         if: runner.os == 'Windows'
         run: cargo build --target x86_64-pc-windows-gnu
-
+        env:
+          LIBCLANG_PATH: C:\msys64\mingw64\bin
```
clauswilke, line 79: We shouldn't need `LIBCLANG_PATH` here, as this build doesn't generate bindings. The linking problem is that the wrong linker is being used, the one from Rtools rather than the one shipped by rustup. The same linking problem arises when the default host is `*-gnu` rather than `*-msvc`, but that doesn't seem to be the problem here.

.github/workflows/test.yml
```diff
@@ -61,16 +71,20 @@ jobs:
...
-
-## Tests don't currently succeed on Windows
-## The following test hangs: tests::r_output_test
-## See: https://github.com/extendr/extendr/issues/67
-#      - name: Run tests (Windows)
-#        if: runner.os == 'Windows'
-#        run: cargo test -- --nocapture --test-threads=1
+        env:
+          LIBCLANG_PATH: C:\msys64\mingw64\bin
+        shell: msys2
```
clauswilke, line 80: The syntax is `msys2 {0}` for some reason.

clauswilke: See e.g. here: https://github.com/extendr/libR-sys/blob/6ff5c6c496ed38143599bd1801b2c7999c765173/.github/workflows/test.yml#L79

Ilia-Kosenkov: Yes I saw that; new commit incoming. The project builds fine on my local machine, I hope it is indeed `msys2` issue.

clauswilke (approved):
Thanks!

Ilia-Kosenkov:
@clauswilke,
All seems to be fine

PR #87, opened 2020-12-21
Better documentation 1

andy-thomason:
A large collection of examples in the crate documentation all tested.

This should significantly improve the start-up cost for new users.


extendr-api/src/lib.rs
```diff
@@ -6,59 +6,58 @@
...
+//! See [Robj] for much of the content of this crate.
+//! [Robj] provides a safe wrapper for the R object type.
 //! 
-//! This library is just being born, but goals are:
+//! Use attributes and macros to export to R.
+//! ```
+//! use extendr_api::*;
+//! // Export a function or impl to R.
+//! #[extendr]
+//! fn fred(a: i32) -> i32 {
```
clauswilke, line 17: Maybe give a more meaningful function name? `add_one`?

extendr-api/src/lib.rs
```diff
@@ -6,59 +6,58 @@
...
+//!     a + 1
+//! }
 //!
-//! Implement common R functions such as c() and print()
+//! // define exports using extendr_module 
+//! extendr_module! {
+//!    mod mymodule;
+//!    fn fred;
```
clauswilke, line 24: See above.

extendr-api/src/lib.rs
```diff
@@ -6,59 +6,58 @@
...
+//! ```
 //!
-//! Example:
+//! In R:
 //!
 //! ```ignore
-//! let v = c!(1, 2, 3);
-//! let l = list!(a=1, b=2);
-//! print!(v, l);
+//! result <- fred(1)
```
clauswilke, line 32: Does this now work without having to define explicit R wrapper functions? Is that taken care of by the modules somehow? That wasn't clear to me.

extendr-api/src/lib.rs
```diff
@@ -6,59 +6,58 @@
...
 //!
-//! Provide a wrapper for r objects.
+//! Common R functions such as c() and print() are available as macros:
 //!
-//! Example:
+//! ```
+//! use extendr_api::*;        // Put API in scope.
+//! extendr_engine::start_r(); // Start test environment.
+//! let v = c!(1., 2., 3.);
+//! let l = list!(a=1, b=2);
```
clauswilke, line 41: Tidyverse R code style would place spaces around `=`: `a = 1, b = 2`

extendr-api/src/lib.rs
```diff
@@ -6,59 +6,58 @@
...
-//!
-//! In R:
-//!
-//! ```ignore
-//!
-//! result <- .Call("fred", 1)
-//!
+//! use extendr_api::*;        // Put API in scope.
+//! extendr_engine::start_r(); // Start test environment.
+//! let res = (1..=100).collect_robj();
```
clauswilke, line 59: Should this be `1..100`?

extendr-api/src/robj.rs
```diff
@@ -22,7 +23,77 @@ use super::numeric_is_na;
...
+/// assert_eq!(a, c);
+/// assert_eq!(a, d);
+///
+/// // Different ways of making boolean scalar TRUE.
+/// let a : Robj = true.into();
+/// let b = r!(TRUE);
+/// assert_eq!(a, b);
+///
+/// // Create a named list
+/// let a = list!(a=1, b="x");
```
clauswilke, line 48: Again, spaces around `=`

extendr-api/src/robj.rs
```diff
@@ -22,7 +23,77 @@ use super::numeric_is_na;
...
+/// assert_eq!(a.len(), 2);
+///
+/// // Use an iterator (like 1:10)
+/// let a = (1..=10).collect_robj();
```
clauswilke, line 52: `1..10`?

extendr-api/src/robj.rs
```diff
@@ -22,7 +23,77 @@ use super::numeric_is_na;
...
+/// assert_eq!(a.len(), 10);
+///
+/// // Use an iterator (like (1:10)[(1:10) %% 3 == 0])
+/// let a = (1..=10).filter(|v| v % 3 == 0).collect_robj();
```
clauswilke, line 56: `1..10`?

extendr-api/src/robj.rs
```diff
@@ -22,7 +23,77 @@ use super::numeric_is_na;
...
+/// assert_eq!(b, vec![1., 2., 3., 4.]);
+/// ```
+///
+/// Convert to/from Rust [std::collections::HashMap].
+///
+/// ```
+/// use extendr_api::*;        // Put API in scope.
+/// extendr_engine::start_r(); // Start test environment.
+///
+/// let abc = list!(a=1, b="x", c=vec![1, 2]);
```
clauswilke, line 90: spaces around `=`.

clauswilke (commented):
Looks good so far. Some minor issues, and some typos that I think are responsible for the breaking tests.

extendr-api/src/lib.rs
yutannihilation: The direct interpretation of `1:100` is `(1..=100)`, so I think this is appropriate here. But, probably we should also show `1..100` and explain it doesn't include `100` (This was one of the things I felt puzzling when I was a very beginner of Rust...).

clauswilke: I see. Shows how little I know about Rust still.

extendr-api/src/lib.rs
yutannihilation: Maybe we should follow rustfmt rather than tidyverse style as this is Rust code?

clauswilke:
Actually, the test errors are in the `#[extendr]` macro:

```
error: custom attribute panicked
 --> src/lib.rs:16:1
  |
test src/lib.rs - (line 13) ... FAILED
6 | #[extendr]
test src/lib.rs - (line 31) ... ignored
  | ^^^^^^^^^^
  |
  = help: message: Unable to write R wrapper file to target directory: Os { code: 2, kind: NotFound, message: "No such file or directory" }

error[E0425]: cannot find function `init__fred` in this scope
  --> src/lib.rs:24:7
   |
14 |    fn fred;    
   |       ^^^^ not found in this scope

error: aborting due to 2 previous errors
```
https://github.com/extendr/extendr/pull/87/checks?check_run_id=1591605832#step:9:54

extendr-api/src/lib.rs
clauswilke: Sounds reasonable. Is the convention no spaces under rustfmt?

yutannihilation: Seems so.

``` console
❯ echo 'fn f() {let l = list!(a=1, b=2);}' | rustfmt
fn f() {
    let l = list!(a = 1, b = 2);
}
```

yutannihilation: It seems we need spaces here anyway, sorry.

``` console
❯ echo 'fn f() {let l = list!(a=1, b=2);}' | rustfmt
fn f() {
    let l = list!(a = 1, b = 2);
}
```

andy-thomason:
The test errors are caused by the R wrapper code misfiring in doc tests. We need to separate metadata from
the wrapper code in a subsequent PR.

For now, I'll disable the doc test.

andy-thomason:
I've added simpler Robj::from (or r!) conversions to include arrays.

You should now be able to return an array (a fixed size list of values) from a function
or assign it to an object.

```
let my_list : Robj = list!(a = [1, 2, 3], b = [1., 2., 3.]);
```

andy-thomason:
Question: should I call f64 values "numeric" or "real".

```
robj.as_real_slice()
vs
robj.as_numeric_slice()
```

andy-thomason:
We now have dozens of example doc tests, more consistent names for functions and hopefully a broad spectrum
of support for many more ways of working.

Writing docs has the advantage that you need to fill in the gaps!

Check the docs by running `cargo doc --target-dir /tmp/`

Please prompt me if anything is missing.

andy-thomason:
Note the r!() and R!() macros.

The first generates an R object from rust.

The second generates an R object from R source code.

I hope to parameterise the R!() macro in future.

yutannihilation:
> Question: should I call f64 values "numeric" or "real".

I guess we should follow R's terminology here; it should be "numeric" if we allow conversion from/to `INTSXP` (and `LGLSXP`?), otherwise "real."

c.f. 
* `isReal()`: https://github.com/wch/r-source/blob/90a3e7d8560e145ef01e1339127c6d67a9deba24/src/include/Rinternals.h#L1821
* `isNumeric()`: https://github.com/wch/r-source/blob/90a3e7d8560e145ef01e1339127c6d67a9deba24/src/include/Rinlinedfuns.h#L961-L972

clauswilke:
Overall looks great! One place where I see room for improvement, in particular for macros, is that it would help to add 1--2 sentences explaining in more detail what each macro does and what its arguments are, before showing the examples. Some macros are already there (e.g. `rprint!`) or are fine as is (e.g., `factor!`), but others need more detail.

A couple of specific comments:
- I'm still not sure what the `lang!` macro does. If its purpose is to create a call, then I think this should be stated explicitly: "This macro can be used to create a function call. The first argument is the name of the function (provided as a string), and subsequent arguments are taken as arguments to the function. Arguments can be named, as in R. To evaluate the function call, use `.eval()` or `.eval_blind()`."
- For the `R!` macro, it would be helpful to add something like the following: "The argument to this macro is arbitrary R code. Note that the code does not need to appear in quotes, but you do have to add a semicolon after every R statement. The macro takes the code and evaluates it with `eval()`, so the return value is of type `Result<Robj, AnyError>`."
- I think having an `r!` macro and an `R!` macro is going to be a source of confusion. Maybe the `r!` macro could be renamed into `ro!` or `rob!` or `rfrom!`?
- I agree with @yutannihilation regarding real/numeric.

andy-thomason:
The lang! macro creates an R language object which may be a call, and operator or other things.

We could call it "make_call!" and drop the eval_blind() (instead use eval().unwrap())

I'm going to rename the f64 functions to "real" from "numeric" then.
We haven't looked at `complex` in any detail, and I've left this alone for now.
There are a few complex number crates such as `num_complex`.

clauswilke:
@andy-thomason One thing that just occurred to me: Both `lang!` and now `call!` probably can't handle function calls to namespaces. Normally you have to locate the namespace first and then you can use `Rf_findFun()` to find the function call in the name space.

As an example, see e.g. here:
https://github.com/clauswilke/sinab/blob/955d38cfb1dc1833433548ff638baf315daa4a4d/src/r_callbacks.c#L18-L19
where the `get_namespace()` function is defined here:
https://github.com/clauswilke/sinab/blob/955d38cfb1dc1833433548ff638baf315daa4a4d/src/r_callbacks.c#L3-L12

Maybe add macros `lang_namespace!` and `call_namespace!` that take a namespace as their second argument? Or modify `lang!` so it can take an R object as input and add a macro `find_fun!` that can locate a function in a namespace?

andy-thomason:
I've been avoiding thinking about environments and namespaces so far, but that suggestion is a good one.

I didn't know about `Rf_findFun`, but I'll read about it tomorrow.

I've switched to talking about "integer", "real", "logical" and "raw" for vectors.

We may make all the macros require `unwrap()` or `?` and should write some examples.

I may also add a `Mutex` mechanism, so we can make multi-threaded R calls and run the
tests in parallel.

clauswilke:
Yeah, it's pretty straightforward. There's `Rf_findVar()` to find any symbol and `Rf_findFun()` to only find functions. See e.g. here: https://github.com/hadley/r-internals/blob/master/environments.md

It's a good idea to use `Rf_findFun()` to construct a call, so if there's no such function you get a more informative error message.

andy-thomason:
@clauswilke We should probably merge this and take the pain of the big change!

clauswilke:
Yes, I agree. Better to take several small steps. In particular since the squash commit will compress all of this into one a single commit.

PR #88, opened 2020-12-23
bumping libR-sys to 0.2.0 on crates.io

clauswilke:
libR-sys 0.2.0 is now on crates.io.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,4 +7,4 @@ members = [
 ]
 
 [patch.crates-io]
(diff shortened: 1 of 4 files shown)


clauswilke:
The integration tests fail because they pull extendr-api from github which doesn't have the required changes yet. I don't see a way to avoid this problem. (Specifying the extendr-api location via relative path doesn't work, because R check moves the test folder around.) So I'll merge and hope for the best. If all goes well, the test on the merge commit will already pass.

PR #90, opened 2020-12-23
Update GitHub Actions

Ilia-Kosenkov:
In line with `extendr/libR-sys` GitHub Actions update, I made an attempt to improve `extendr`'s CI.
The goal is to be able to control the setup using matrix configuration and share as many steps as possible between OSs.

The problem is Windows builds were testing only `x86_64` binaries, but R integration tests were executed for both `x86_64` and `i386`. Because Rust does not support multi-targeting  (yet,  https://github.com/rust-lang/cargo/issues/8176), I simulate multiple targets for `cargo`. It does not affect other builds, but Windows builds are controlled through `rust-version`:
- if `rust-version` does not include architecture id, `cargo build`/`cargo test` and R integration tests are executed for both architectures
- if `rust-version` includes architecture specifier (`x86_64`/`i686`), then `cargo build`/`cargo test` are executed for one target only, and R integration tests are launched from R process with correct bitness and an extra `--no-multiarch` flag.

Currently, architecture-specific Wndows builds are disabled and commented out, but can be re-enabled to e.g. test 64-bit only build.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,30 +28,16 @@ jobs:
           # - {os: windows-latest, r: 'devel', rust-version: 'stable-msvc'}
           # - {os: windows-latest, r: 'oldrel', rust-version: 'stable-msvc'}
 
-          # Windows jobs with x86_64 Rust toolchain build for x86_64 only
-          # R integration tests are run with 64-bit R version and '--no-multiarch' flag
-          # - {os: windows-latest, r: 'release', rust-version: 'stable-x86_64-pc-windows-msvc'}
-          # - {os: windows-latest, r: 'release', rust-version: 'nightly-x86_64-pc-windows-msvc'}
-          # - {os: windows-latest, r: 'devel', rust-version: 'stable-x86_64-pc-windows-msvc'}
-          # - {os: windows-latest, r: 'oldrel', rust-version: 'stable-x86_64-pc-windows-msvc'}
-
-          # Windows jobs with i686 Rust toolchain build for i686 only
-          # R integration tests are run with 32-bit R version and '--no-multiarch' flag
-          # - {os: windows-latest, r: 'release', rust-version: 'stable-i686-pc-windows-msvc'}
-          # - {os: windows-latest, r: 'release', rust-version: 'nightly-i686-pc-windows-msvc'}
-          # - {os: windows-latest, r: 'devel', rust-version: 'stable-i686-pc-windows-msvc'}
-          # - {os: windows-latest, r: 'oldrel', rust-version: 'stable-i686-pc-windows-msvc'}
-
           - {os: macOS-latest,   r: 'release', rust-version: 'stable'}
           # - {os: macOS-latest,   r: 'release', rust-version: 'nightly'}
           # - {os: macOS-latest,   r: 'devel', rust-version: 'stable'}
           # - {os: macOS-latest,   r: 'oldrel', rust-version: 'stable'}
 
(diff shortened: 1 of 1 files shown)


clauswilke:
I'm not sure this is going in the right direction. A couple of comments:

1. On CRAN, Windows packages are tested for both architectures at the same time. Therefore, we should do the same here.
2. We should do every effort to limit the number of CI minutes we need, even if we're not paying for them. First, it's the polite thing to do, and second GitHub sometimes throttles runs when their system is under heavy load, so the more CI runs we have and the longer they take the more annoying things can become.
3. Because of point 2, I wouldn't want to test every combination of OS and R version. We need to do this for libR-sys to generate bindings, but we don't need to do it here. We could test 3 R versions for 1 OS only, e.g. Ubuntu.
4. Also because of point 2, I'd rather keep all the Windows tests in a single run. Setting up R and Rust takes up the majority of the time on these runs, while testing on two architectures is rather fast. This also aligns with point 1: test on all Windows architectures at once.

Ilia-Kosenkov:
@clauswilke,
We do test for both architectures of Windows at the same time when running R integration tests, but rust binaries were not tested -- only one architecture, hardcoded 
https://github.com/extendr/extendr/blob/cecc67f37a425b97fc441854bee678501d247eed/.github/workflows/test.yml#L75-L83

To your points 2-4: this can be easily done by changing the matrix configuration -- I did not change any steps that were in the original script, only polished it slightly and unified Build/Test steps.
For instance, if you want to preserve GHA minutes, leave only `os::r-release::rust-stable`, three builds for the first job, and four builds for the second job because Windows deals with two architectures at the same time.

clauswilke:
For your first comment: Ah, I think I see it now, for Windows you're doing both architectures with precomputed bindings but only one when using bindgen. Does it make sense to do both architectures at once when using bindgen also, or does that cause trouble? I guess it's not possible to generate bindings on the fly for multiple architectures on Windows? At a minimum, I think it would be good to add a few comments explaining what is happening in various places.

For your second comment: Could you pare it down the way you suggest, so I can take a look?

Ilia-Kosenkov:
@clauswilke,
I think I can do the same trick that I did for the first job and force it to execute each cargo step twice.
Let me see what I can do. I will also reduce the number of different builds we run.

Ilia-Kosenkov:
@clauswilke,
It seems to work. You can inspect the output of e.g. Winows' `cargo test` to see that it in fact runs twice for each command, targeting different architectures.

I believe this trick of doing more than one Rust target can help us later to cross-compile for M1 (if that will be the case).
The script will be simplified a little bit once `rustc` starts to accept multiple `--target` attributes.

clauswilke:
This generally looks good. Could you remove the entire commented-out configurations for x86_64 only and for i686 only and simplify the R CMD check code back to what it was before?

And also, we probably want turn on (on ubuntu only) R release/Rust nightly, R oldrel/Rust stable, R devel/Rust stable.

Thanks!

clauswilke:
Sorry, one more item: Since you're now using for loops to loop over multiple targets, it's probably a good idea to add an `echo` statement inside each loop printing out what target we're working on.

clauswilke:
One last comment: Does your strategy for R devel avoid the need for setting `LD_LIBRARY_PATH`? If yes, then could you delete the comment about needing it?

Ilia-Kosenkov:
It still sets `LD_LIBRARY_PATH` for all linux machines (see Configure linux step), as it was in the original script. I did not test it without this variable though, but I suspect that `r-devel` requires it.

Ilia-Kosenkov:
Checked in my fork - r-devel definitely needs this variable set.

clauswilke:
Sounds good, thanks!

clauswilke (approved):

PR #91, opened 2020-12-24
Add integration tests: scalar inputs/outputs

clauswilke:
I'm adding a bunch of systematic tests to the integration test framework.


clauswilke:
@andy-thomason It's very strange, the module macro works fine on all platforms except Windows, but manual registration of functions from C works on all platforms.

This works everywhere *except* Windows (see tests here: https://github.com/extendr/extendr/actions/runs/441652467):
https://github.com/extendr/extendr/blob/b9ffd05b72eef173e09415f7e4ebe33ae3522a60/tests/extendrtests/src/entrypoint.c#L1-L5
https://github.com/extendr/extendr/blob/b9ffd05b72eef173e09415f7e4ebe33ae3522a60/tests/extendrtests/src/rust/src/lib.rs#L20-L28

This works everywhere (see tests here: https://github.com/extendr/extendr/actions/runs/441722384):
https://github.com/extendr/extendr/blob/b4ee9f71125f95ed9eac001570f86c0865eb7217/tests/extendrtests/src/entrypoint.c#L9-L33

clauswilke:
@Ilia-Kosenkov If you'd like to dig into another Windows-specific issue, you could see if you can figure out why routine registration works from C but not from Rust. I don't see where the code is different, but there must be a difference somewhere.

The C registration code is here:
https://github.com/extendr/extendr/blob/b4ee9f71125f95ed9eac001570f86c0865eb7217/tests/extendrtests/src/entrypoint.c

The corresponding Rust registration code is generated in the `extendr_module` macro here:
https://github.com/extendr/extendr/blob/db5a7e2c6b4b363aacd9246e259e61049b3d8596/extendr-macros/src/lib.rs#L495-L512
and then here:
https://github.com/extendr/extendr/blob/db5a7e2c6b4b363aacd9246e259e61049b3d8596/extendr-api/src/lib.rs#L102-L127

Ilia-Kosenkov:
Let's see what I can do, currently I have no idea why this happens.

Ilia-Kosenkov:
@clauswilke,
Ok, there is something. Basically, `i686` dll is empty -- no exports.
The tests succeed if I run from 64-bit process with `--no-multiarch`.
Have no idea why though, will try to figure it out.

clauswilke:
@Ilia-Kosenkov We've had problems in the past with linkers on various architectures being overly aggressive and removing symbols generated by Rust, because they weren't used at all in the C part that R generates. That's why we added code such as the following:
https://github.com/extendr/extendr/blob/b9ffd05b72eef173e09415f7e4ebe33ae3522a60/tests/extendrtests/src/entrypoint.c#L1-L5

Maybe that's not sufficient to trick the linker on Windows. The way to test this is to retain more and more of the C routine registration code, maybe the entire definition of the `CallEntries` array, and see whether at some point the tests succeed.

clauswilke:
It's also possible that we need to include `Rinternals.h` to obtain an object the linker knows how to handle. I'll experiment with some of these ideas via GH Actions also.

clauswilke:
FYI, I'm also seeing now that the tests succeed on the x64 architecture, see e.g. here:
https://github.com/extendr/extendr/runs/1606086085?check_suite_focus=true#step:14:220

```
** running tests for arch 'x64' ...

  Running 'testthat.R'

 OK
```

Somehow I had missed that previously.

clauswilke:
Very strange. If I leave both C and Rust routine registration code in and add `_decoy` to the C routine registration function name things don't work, but if I add `_decoy` to the Rust routine registration code instead things work:
https://github.com/extendr/extendr/pull/91/commits/07aca48825b113d1a9d89bbb53deb0150b43bf59

It seems like on Windows i686 if the C code doesn't contain a registration function of the right name things don't work. But how does the linker even know what the right name is? How can it distinguish between `R_init_extendrtests` and `R_init_extendrtests_decoy`?

Ilia-Kosenkov:
I also got this out of my log, don't know if it helps
```script
File 'extendrtests/libs/x64/extendrtests.dll':
    Found no calls to: 'R_registerRoutines', 'R_useDynamicSymbols'
```

clauswilke:
Ok, here is the solution: We need to forward routine registration from C to Rust, something like this:
```
void R_init_extendrtests_extendr(DllInfo *dll);

void R_init_extendrtests(DllInfo *dll) {
  R_init_extendrtests_extendr(dll);
}
```
In a way, this also elegantly solves the issue of empty symbols, because we don't need to add any other dummy code to prevent them.

The postfix `_extendr` could be added automatically to the function name by `extendr_module!` so the user doesn't have to think about it.

@andy-thomason If this looks fine to you I'll prepare a PR to modify `extendr_module!` accordingly.

clauswilke:
I made the separate PR to fix routine registration. It's #93.

Issue #89, opened 2020-12-23
Integration tests create GitHub merge chicken-and-egg problem


clauswilke:
The integration tests contain a separate R package that uses extendr just like any stand-alone package would. It pulls the latest extendr sources from GitHub. This creates a chicken-and-egg problem if the integration tests depend on some new extendr feature that hasn't been pushed to the main branch on GitHub yet.

I ran into this when I bumped libR-sys to version 0.2.0. You can see that the latest commit before the merge failed the CI tests: https://github.com/extendr/extendr/pull/88/commits/42b5ccb15b418a645201151eed3f327c6ab75382. However, the actual merge into master had all tests succeed: https://github.com/extendr/extendr/commit/cecc67f37a425b97fc441854bee678501d247eed.

I don't currently see a way to fix this. The offending lines in the code are here:
https://github.com/extendr/extendr/blob/cecc67f37a425b97fc441854bee678501d247eed/tests/extendrtests/src/rust/Cargo.toml#L19-L21
We might think to simply use relative paths within the repo here, but that won't work because the entire `extendrtests` folder gets copied into a temp directory when the integration tests are run, and thus relative paths break.

I'm not sure there's anything that can be done, so this issue may primarily serve as an explanation and reminder of what is happening.

clauswilke:
Here's a workflow that is a bit cumbersome but should at least work without having to merge failing tests. It requires two steps.

First, in you PR, in the three lines I highlighted above, point the extendr sources to the fork/branch that contains the new feature, e.g.:
```
 extendr-api = { git = "https://github.com/clauswilke/extendr", branch = "new-feature" } 
```
This should then allow the tests to succeed. Merge this PR. **Important: Do not delete the branch after merging!** It's still being used.

Second, make another PR that reverts the extendr sources back to the main extendr repo, and, if all tests pass, merge. Now, if you want, you can delete the branches you created for both PRs.

andy-thomason:
You should be able to use many different sources of packages together,
In theory, **cargo publish** should modify the `Cargo.toml ` to leave just the package.

I usually use a standard layout of the project:

```
/home/myuser/extendr/extendr
/home/myuser/extendr/libR-sys
```

Which allows us to use relative paths, github refs and crates:
```
 libR-sys = { path = "../../libR-sys", version="xx.yy.zz" } 
```

This way it should build using the path in development and the crate for production.

Of course, without the workspace, we rely on this directory structure.

clauswilke:
[...]

PR #93, opened 2020-12-24
Fix module registration.

### LINKED ISSUE #75: Module registration currently doesn't work
In trying to compile an R package using extendr on Windows, I have found that the module registration doesn't work. I believe the problem is in this line:
https://github.com/extendr/extendr/blob/41772722a01062b064320c98d687e3eae889790a/extendr-macros/src/lib.rs#L347

The function to be called should be named `R_init_<modname>`, not `R_init_lib<modname>`. See e.g. here:
https://github.com/r-rust/hellorust/blob/0b53c737ead980f58fe9584fbf513f15fdfe0582/src/wrapper.c#L30

When doing manual, C-style registration the checks work:
https://github.com/extendr/helloextendr/runs/1561836399?check_suite_focus=true

When doing the current module registration the Windows check doesn't work, the package compiles but then the exported function isn't found:
https://github.com/extendr/helloextendr/runs/1561757762?check_suite_focus=true#step:11:251

For some reason, Linux and macOS don't care whether the functions are properly registered or not.

clauswilke:
Fix module registration by forwarding the registration call from C to Rust. Apparently this is needed on Windows i686. See discussion here: #91 

This closes #75.

### DIFF
--- a/tests/extendrtests/src/rust/Cargo.toml
+++ b/tests/extendrtests/src/rust/Cargo.toml
@@ -16,7 +16,7 @@ crate-type = ["staticlib"]
 extendr-api = "*"
 
 [patch.crates-io]
-extendr-api = { git = "https://github.com/extendr/extendr" }
-extendr-engine = { git = "https://github.com/extendr/extendr" }
-extendr-macros = { git = "https://github.com/extendr/extendr" }
+extendr-api = { git = "https://github.com/clauswilke/extendr", branch = "reg-forwarding" }
+extendr-engine = { git = "https://github.com/clauswilke/extendr", branch = "reg-forwarding" }
+extendr-macros = { git = "https://github.com/clauswilke/extendr", branch = "reg-forwarding" }
 #libR-sys = { git = "https://github.com/extendr/libR-sys" }
--- a/extendr-api/examples/R/data/src/entrypoint.c
+++ b/extendr-api/examples/R/data/src/entrypoint.c
@@ -1,11 +1,9 @@
+// We need to forward routine registration from C to Rust 
+// to avoid the linker removing the static library.
 
+void R_init_data_extendr(void *dll);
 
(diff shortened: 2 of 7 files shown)


andy-thomason:
Worth noting that the `extendr_module!` macro works pretty well now.

andy-thomason:
Ah. I'm a bit behind the times. There is a naming issue with the DLL on windows that you are fixing.

clauswilke:
Yes, correct. On Windows the linker strips all symbols from the generated library unless there's a function called `R_init_<pkgname>()` in the C file. This PR fixes this problem.

tests/extendrtests/src/rust/Cargo.toml
```diff
@@ -16,7 +16,7 @@ crate-type = ["staticlib"]
 extendr-api = "*"
 
 [patch.crates-io]
-extendr-api = { git = "https://github.com/extendr/extendr" }
-extendr-engine = { git = "https://github.com/extendr/extendr" }
-extendr-macros = { git = "https://github.com/extendr/extendr" }
+extendr-api = { git = "https://github.com/clauswilke/extendr", branch = "reg-forwarding" }
+extendr-engine = { git = "https://github.com/clauswilke/extendr", branch = "reg-forwarding" }
+extendr-macros = { git = "https://github.com/clauswilke/extendr", branch = "reg-forwarding" }
```
hobofan, line 21: @clauswilke Should this have been merged? Looks like a temporary workaround.

clauswilke: It's necessary, see #89. I'll fix with #91 as soon as GH Actions complete.

hobofan: 👍 Ah okay, hadn't seen that issue yet.

PR #97, opened 2020-12-28
Minor fix on the guide to bump version

yutannihilation:
Address https://github.com/extendr/extendr/pull/82#issuecomment-750396078

### DIFF
--- a/MAINTAINERS_GUIDE.md
+++ b/MAINTAINERS_GUIDE.md
@@ -88,7 +88,8 @@ cargo ws version --force='*' --no-individual-tags --no-git-push patch
 # review the changes
 git show HEAD^
 
-# push all the changes
+# push all the commits and tags
+git push
 git push --tags
 
 # publish without modifying the current version


andy-thomason:
How is this going? I'm trying to get as much as possible in for 0.2.0.

Did your cargo ws experiments work out?

yutannihilation:
Yes, you can just execute this one line to publish all the crates with `minor` version increment:

``` sh
cargo ws publish --force='*' --no-individual-tags minor
```

and then, you can bump the versions to the development version (`-alpha.0`) by

``` sh
cargo ws version --force='*' --no-individual-tags --pre-id alpha prerelease
```

What this pull request fixes is the manual steps to do this just in case above doesn't work. I don't have crates to publish so I cannot test `cargo ws publish`, but I confirmed `cargo ws version` works very well (although I needed some time to figure out how to use versatile options...).

yutannihilation:
I'm merging this as this is really a minor change, but please file a new issue if there's anything unclear to you about cargo-workspaces. I'm happy to investigate further :)

clauswilke:
I followed these instructions when I published libR-sys 0.2.0 to crates.io and it worked well.

One other remark: I am now convinced that bumping the version to a development version right after publishing **is not** the right approach. Each time we bump the version with `cargo ws version`, we create tags in the git history, and it doesn't make sense to tag one commit with two different versions, or even tag two consecutive commits with two different versions. We should only bump to a development version number (if at all) once some meaningful additional development has happened.

For context, see the releases for libR-sys on GitHub:
https://github.com/extendr/libR-sys/releases 
The last two tags were automatically created by `cargo ws version`. We would want all of these releases to be meaningful versions.

yutannihilation:
Hmm, that's strange. `--no-individual-tags` should skip these tags for individual crates (e.g. ` libR-sys@0.2.0`). Should I open a new issue and investigate?

clauswilke:
I probably forgot to add `--no-individual-tags` when I ran the code. I'm about to release a new version of libR-sys, will pay attention to this and see what happens.

clauswilke:
Or rather, after re-reading the documentation and thinking a bit about it: Since libR-sys is only a single crate, I didn't think this was relevant, but apparently it is.

clauswilke:
However, I do think we should remove the instructions for bumping the version immediately to pre-release, because that will create two different version tags on subsequent commits, and that doesn't make much sense.

yutannihilation:
Sure, I'll tweak the instructions, but let me clarify a bit. We can commit and push without tags by specifying `--no-git-tag`. Is this the behavior you want? Or, do you want to stop using pre-release versions at all (I feel pre-release version is not necessary in general).


> Since libR-sys is only a single crate, I didn't think this was relevant, but apparently it is.

Not sure if this is relevant, but I agree we don't need to use cargo-workspaces in this case.

clauswilke:
> Not sure if this is relevant, but I agree we don't need to use cargo-workspaces in this case.

What I was trying to say was the opposite. `cargo ws version` is still very helpful for individual crates, but it needs `--no-individual-tags` even when it's only a single crate.

yutannihilation:
Ah, got it.

clauswilke:
I just did the latest release of libR-sys following the instructions here (with `--no-individual-tags`) and this time I didn't get the extra tag, so things seem to work as intended:
https://github.com/extendr/libR-sys/releases

However, in the instructions, should `git show HEAD^` be `git show HEAD`? With the `^` I don't see the commit that added the tag.

yutannihilation:
Thanks for testing! Probably I meant `git diff HEAD^`, sorry.

PR #103, opened 2020-12-29
Make calls to R thread safe.

andy-thomason:
Until now, running in multi-threaded environments has been not possible for **extendr**.

With this PR, we introduce a function `single_threaded` which wraps calls to R which
would allocate or write data.

We assume that functions in R that only read are thread safe, but this may not be true and
if in doubt consult the R source-code.

I can't guarantee that my amateur attempts at making a recursive mutex using thread local variables
and atomics are perfect, but they do seem to pass the tests.

I am now getting test runs to pass without the `--test-threads=1`, but a few more checks and runs would
be necessary to make this routine without random failures.

### DIFF
--- a/extendr-api/0001-Add-single_threaded-and-wrap-R-api-writes.patch
+++ b/extendr-api/0001-Add-single_threaded-and-wrap-R-api-writes.patch
@@ -1,0 +1,710 @@
+From 2e8907f4eb47af5634f4973478159915a076cc2a Mon Sep 17 00:00:00 2001
+From: andy-thomason <andy@andythomason.com>
+Date: Tue, 29 Dec 2020 19:27:34 +0000
+Subject: [PATCH] Add "single_threaded" and wrap R api writes.
+
+---
+ extendr-api/src/lang.rs            |  23 +++--
+ extendr-api/src/lib.rs             |  25 +----
+ extendr-api/src/robj/into_robj.rs  | 141 ++++++++++++++++-------------
+ extendr-api/src/robj/mod.rs        |  34 +++----
+ extendr-api/src/robj/rinternals.rs |  36 ++++----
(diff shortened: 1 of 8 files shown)


andy-thomason:
Note that this is based on the substantial changes in the berrer docs PR.

clauswilke:
I think it's best to assume any call to the R API needs to be wrapped. The R documentation says repeatedly that you shouldn't assume any particular API functions don't allocate memory/call the GC.

From https://cran.r-project.org/doc/manuals/R-exts.html#Garbage-Collection:

> ... in general we do not know (nor want to know) what is hiding behind the R macros and functions we use, and any of them might cause memory to be allocated, hence garbage collection and hence our object `ab` to be removed. It is usually wise to err on the side of caution and assume that any of the R macros and functions might remove the object.

andy-thomason:
I would draw the line at NA tests and other trivial reads. It could get costly to wrap everything
as well as dangerous (as RF_error and panics are going to be bad).

It would be nice to hook the allocator and disable it when not in a single_threaded() block.

clauswilke:
Ah, I see, yes.

andy-thomason:
I need to tidy this up once we merge #87

andy-thomason:
I've just discovered the joy of git format-patch and git apply.

```
git reset --hard <your change>
git format-patch -1 (generates patch file in current dir)
git reset --hard <change from master branch>
git apply <patch file>
git push -f
```

And behold! our change is on master.

clauswilke:
> I've just discovered the joy of git format-patch and git apply.

FYI, when you're working with pull requests, you can always just merge in master. Any ugly merge commits disappear in the final squash merge.

clauswilke:
Btw., looks like you accidentally committed the patch:
https://github.com/extendr/extendr/blob/master/extendr-api/0001-Add-single_threaded-and-wrap-R-api-writes.patch

PR #105, opened 2020-12-30
Environments, object access and wrappers

andy-thomason:
This is an overhaul of wrappers including new environment support.

This also cleans up some protection issues.

Raw vectors now have a distinct `Raw()` wrapper instead of the `&[u8]` conversion before.

Unnamed pair lists have a wrapper.

I plan to do named pair lists and "new" lists also.

I plan to add access functions such as "dollar", "subset" and "index" for $, [ and [[.


clauswilke:
This is mostly beyond me but I've looked it over and it seems reasonable. A few more comments in places might be useful.

andy-thomason:
Thanks for having a look anyway!

I'm trying to use traits as much as possible which makes it a bit more
complex to read.

For example, using AsRef<str> allows you to use &str, &&str, String,
&&String and so on
as a string parameter.

IntoIterator is a fairly profound thing to support as almost all the Rust
collections generate
it.

I wanted to be able to generate almost any R type, but environments have
proven difficult.
The `R_NewEnv` function seems to be missing from my bindings and so I'm
using a `call!`
instead.

I'm mindful of the namespacing issue. I may allow non-string things in the
call! macro
or something to generate language objects for "mypackage::fred" and
"::mypackage::fred::jim::harry".
(ie. Lang(&[double_colon_symbol(), r!(Symbol("mypackage)),
r!(Symbol("fred"))])

We could do a fast path to known symbols via Symbol, also which bypasses
the conversions:
(ie. Lang(&[Symbol("::"), Symbol("mypackage"), Symbol("fred")])

Some wrappers take integer, real and logical parameters.

Env{parent: global_env(), names: &["a", "b"], values: &[1, 2]}

Obviously it doesn't work if you mix the types:

Env{parent: global_env(), names: &["a", "b"], values: &[1, 1.0]}

You can also use iterators:

Env{parent: global_env(), names: &["a", "b"], values: 1..3}

clauswilke:
> The `R_NewEnv` function seems to be missing from my bindings and so I'm using a `call!` instead.

In which header is this included? We don't yet cover all headers in libR-sys and we could include more.

> I'm mindful of the namespacing issue. 

I find the various options you propose all a bit cumbersome. Why not create a macro or struct that can separately take a namespace argument and a function argument, e.g. `lang_namespace!("lm", "stats", y ~ x)` for `stats::lm(y ~ x)`? R doesn't have nested namespaces, so we only ever have one occurrence of `::` in each expression.

andy-thomason:
Likewise, we could make call! look for :: and parse it.

call!("stats::lm", y.tilda("x"))

clauswilke:
If this can be done at compile time rather than runtime that would be great!

andy-thomason:
We are getting close to an R parser at that point!

As a procedural macro, we can do pretty much anything.

clauswilke:
I don't understand why `R_NewEnv()` is missing. All the functions around it in the header file are included in the bindings, and I don't see anything about it that is special.

clauswilke:
Aha. `R_NewEnv()` is new in R 4.1. Maybe a good reason to support version-specific compilation. I'll look into enabling this in libR-sys.

https://github.com/wch/r-source/commit/8a322e786068ec2e23324e61821ff666d0b6e329

clauswilke:
I merged the changelog in case you wanted to add a line or two about these efforts:

https://github.com/extendr/extendr/blob/master/CHANGELOG.md

PR #107, opened 2021-01-02
More environment related features.

andy-thomason:
Support for adding and finding variables in environments.

```
global_var()
set_global_var()
env.local()
env.set_local()
env.find_var()
env.find_func()
env.ls() 
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -130,7 +130,7 @@ jobs:
         run: |
           foreach($target in ($env:BUILD_TARGETS).Split(',')) {
             echo $target
-            cargo test $(if($target -ne 'default') {"--target=$target"} ) -- --nocapture --test-threads=1
+            cargo test $(if($target -ne 'default') {"--target=$target"} ) -- --nocapture
           }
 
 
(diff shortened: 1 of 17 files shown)


clauswilke (approved):
LGTM.

I see that you're now running tests multi-threaded in Visual Studio. Do you want to change all the tests in Github Actions to multi-threaded? You'd have to remove `test-threads=1` at these locations:
https://github.com/extendr/extendr/blob/d7bc7f9f048720513efac9dd53f44413afe6cb69/.github/workflows/test.yml#L133
https://github.com/extendr/extendr/blob/d7bc7f9f048720513efac9dd53f44413afe6cb69/.github/workflows/test.yml#L259
https://github.com/extendr/extendr/blob/d7bc7f9f048720513efac9dd53f44413afe6cb69/.github/workflows/test.yml#L263
https://github.com/extendr/extendr/blob/d7bc7f9f048720513efac9dd53f44413afe6cb69/.github/workflows/test.yml#L267
https://github.com/extendr/extendr/blob/d7bc7f9f048720513efac9dd53f44413afe6cb69/.github/workflows/test.yml#L271

extendr-api/src/robj/functions.rs
```diff
@@ -1,6 +1,86 @@
...
+/// Create a new, empty environment parented on global_env()
+/// Use the Env{} wrapper for more detail.
+/// ```
+/// use extendr_api::*;
+/// extendr_engine::start_r();
+/// let env = new_env();
+/// assert_eq!(env.len(), 0);
+/// ```
+pub fn new_env() -> Robj {
+    call!("new.env", TRUE, global_env(), 29).unwrap()
```
clauswilke, line 13: Maybe add a comment here that `29` is the default value for that parameter in `new.env()`.

extendr-api/src/robj/functions.rs
```diff
@@ -1,6 +1,86 @@
...
+}
+
+/// Get a global function from global_env() and ancestors.
+/// ```
+/// use extendr_api::*;
+/// extendr_engine::start_r();
+/// let ls = global_function(sym!(ls)).unwrap();
+/// assert_eq!(ls.is_function(), true);
+/// // Note that the following will throw an R error:
+/// // let iris = global_function(sym!(iris));
```
clauswilke, line 37: Can you explain in the comment why the error? Is it because `iris` is not a function?

andy-thomason:
I'll try to make that clearer...

It seems that everything is implemented as promises in the base
environment. I don't yet know the correct procedure for unpacking them. For
now I'm using eval.

clauswilke:
Ah yes, datasets are lazy-loaded, so `iris` doesn't exist until you try to access it for the first time.

extendr-api/src/robj/functions.rs
andy-thomason: They could have chosen 42, but it is not prime!
I'll add a comment.

29 seems like a sensible default. We could make a make_sized_env() which attempts to
pick a better set of parmeters.

For example, small envs are probably better off as a unhashed.

andy-thomason:
If we are feeling brave we could drop the --test-threads in the workflows. I'm bound to
have missed many stupid things, but it is good to test.

andy-thomason:
With this change, we should be able to make functions that return Results.

```
#[extendr]
fn my_func() -> Result<f64> {
    Ok(1.0) // Changing this to Err("fail".into()) will cause a panic on return and hence and R error.
}
```

PR #110, opened 2021-01-06
Add vector, matrix and matrix3d wrappers.

andy-thomason:
This is a placeholder for Extendr-native array support.

The aim is to provide basic array functionality and defer to other optional libraries if more complexity is required.


clauswilke (approved):
LGTM

andy-thomason:
I had to decide between 1-based indexing a and 0-based. We may need to offer the option as ndarray do.

Issue #111, opened 2021-01-07
extendr-api cannot be compiled when ndarray is enabled


yutannihilation:
As I commented on https://github.com/extendr/rextendr/issues/10#issuecomment-755387234, I see this error on my local, but this doesn't happen on CI runs. Does anyone see the same errors? (I'm using Rust 1.49 on Ubuntu 20.04).

``` console
❯ cargo test --manifest-path extendr-api/Cargo.toml --features ndarray
   Compiling libR-sys v0.2.0
   Compiling extendr-engine v0.1.11 (/home/yutani/repo/extendr/extendr-engine)
   Compiling extendr-api v0.1.11 (/home/yutani/repo/extendr/extendr-api)
error[E0106]: missing lifetime specifier
  --> extendr-api/src/robj_ndarray.rs:14:24
   |
14 |     Robj: AsTypedSlice<T>,
   |                        ^ expected named lifetime parameter
   |
help: consider using the `'a` lifetime
   |
14 |     Robj: AsTypedSlice<'a, T>,
   |                        ^^^

error[E0106]: missing lifetime specifier
  --> extendr-api/src/robj_ndarray.rs:14:24
   |
14 |     Robj: AsTypedSlice<T>,
   |                        ^ expected named lifetime parameter
   |
help: consider using the `'a` lifetime
   |
14 |     Robj: AsTypedSlice<'a, T>,
   |                        ^^^

error[E0310]: the parameter type `T` may not live long enough
  --> extendr-api/src/robj_ndarray.rs:17:31
   |
12 | impl<'a, T> FromRobj<'a> for ArrayView1<'a, T>
   |          - help: consider adding an explicit lifetime bound...: `T: 'static`
...
17 |         if let Some(v) = robj.as_typed_slice() {
   |                               ^^^^^^^^^^^^^^ ...so that the reference type `&[T]` does not outlive the data it points at

error: aborting due to 2 previous errors

Some errors have detailed explanations: E0106, E0310.
For more information about an error, try `rustc --explain E0106`.
error: could not compile `extendr-api`
```

clauswilke:
I get the same error on Rust 1.48.

clauswilke:
And if I apply the fix suggested by the compiler (`Robj: AsTypedSlice<'a, T>`) the error goes away.

If this error doesn't arise on CI runs are we sure that this code actually gets compiled and tested during CI?

yutannihilation:
Fixed by #118

PR #112, opened 2021-01-07
Generate ndarray-related documentations

yutannihilation:
As suggested by https://github.com/extendr/rextendr/issues/10#issuecomment-755374666, this PR adds `--all-features`, though this won't compile on my local (#111)... Can I just merge this and see what will happen?

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -16,7 +16,7 @@ jobs:
         with:
           toolchain: stable
       - name: Docs
-        run: cargo doc --workspace --no-deps
+        run: cargo doc --workspace --no-deps --all-features
       - name: Deploy
         uses: peaceiris/actions-gh-pages@v3
         with:


clauswilke:
See https://github.com/extendr/extendr/issues/111#issuecomment-755829105. I think you should apply the fix suggested by the compiler, even if the problem doesn't arise in CI.

clauswilke:
It looks to me like we're only testing default features on CI. Maybe this changed when we rewrote the CI scripts? I thought we were testing different feature combinations, but maybe I'm misremembering.

https://github.com/extendr/extendr/blob/7d247b0f029bc5d0c06595558404a36f73a504f8/.github/workflows/test.yml#L133

clauswilke:
And there's another problem: #114 

(Everything else in extendr is now supposed to be thread-safe, so this should be fixed if possible.)

yutannihilation:
Alas, it's tested and failed, but the CI won't fail...

c.f. https://github.com/extendr/extendr/runs/1659948167#step:8:45

It's tested on another matrix here:

https://github.com/extendr/extendr/blob/7d247b0f029bc5d0c06595558404a36f73a504f8/.github/workflows/test.yml#L255-L272

clauswilke:
Oh, Ok, so the test errors out but we aren't notified. Maybe that's a case for @Ilia-Kosenkov.

clauswilke:
I'm not very good with shell scripting but I think the problem is that the return codes from the various `cargo test` invocations are not collected. Normally you'd do something like `command 1 && command 2` to test two separate commands for success. Here we'd have to collect the return values and and them together somehow.

Ilia-Kosenkov:
I will try to fix it later today (or as soon as I can) , I had a similar issue in `libR-sys`.

yutannihilation:
Oh, thanks for taking a look. So, probably we need these lines?

https://github.com/extendr/libR-sys/blob/5ca8d402ff0f56c67544d87ec58fed2d1adbe50c/.github/workflows/test.yml#L140-L143

Ilia-Kosenkov:
@yutannihilation ,
You are correct, this allows to fail fast as soon as one of the builds / tests fail.
`$?` evaluates to either true or false depending on the exit code of last program, and `$LASTEXITCODE` is self-explanatory.

If you would like to implement this, feel free to do so.

yutannihilation:
Thanks, then let me try this!

yutannihilation:
Now I confirmed this will generate `ArrayView1` and `ArrayView2` on the page of `Trait extendr_api::FromRobj`. Let me merge this.

yutannihilation:
Yay!

https://extendr.github.io/extendr/extendr_api/trait.FromRobj.html#impl-FromRobj%3C%27a%3E-for-ArrayView1%3C%27a%2C%20T%3E

PR #115, opened 2021-01-07
Let CI fail when the build/test fails

yutannihilation:
As discussed on https://github.com/extendr/extendr/pull/112#issuecomment-755836852, we need to check the status (`$?`) and throw an error manually here.


Ilia-Kosenkov:
I see that all bindgen builds fail with some rust errors and the CI is now correctly configured. I suggest to merge it as is if there are no objections.

Ilia-Kosenkov (approved):

clauswilke:
Yes, go ahead.

yutannihilation:
Sure.

PR #119, opened 2021-01-07
More integration tests

andy-thomason:
Making sure there are conversions to and from common types.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -6,7 +6,7 @@
         {
             "label": "test",
             "type": "shell",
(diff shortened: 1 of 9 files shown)


clauswilke:
@andy-thomason The Rust code for integration tests should go here:
https://github.com/extendr/extendr/blob/master/tests/extendrtests/src/rust/src/lib.rs

Then you can test calling it from R here:
https://github.com/extendr/extendr/blob/master/tests/extendrtests/tests/testthat/test-type-conversion.R

andy-thomason:
It looks like my guess may have been true that `AtomicU64` has bugs on 32 bit architecture.

PR #121, opened 2021-01-08
Implement unwind protection by converting R errors to panics.

andy-thomason:
Wrap an R function such as Rf_findFunction and convert errors and panics into results.
```
use extendr_api::*;
test! {
   let res = catch_r_error(|| unsafe {
       throw_r_error("bad things!")
       std::ptr::null_mut()
   });
   assert_eq!(res.is_ok(), true);
}
```

### DIFF
--- a/extendr-api/src/error.rs
+++ b/extendr-api/src/error.rs
@@ -24,6 +24,7 @@ pub fn unwrap_or_throw<T>(r: std::result::Result<T, &'static str>) -> T {
 
 #[derive(Debug, PartialEq)]
 pub enum Error {
+    Panic{},
     NotFound,
     NotAVectorType,
     ParseError,
--- a/extendr-api/src/lib.rs
+++ b/extendr-api/src/lib.rs
@@ -209,7 +209,7 @@ pub use lang::*;
 pub use logical::*;
 pub use rmacros::*;
 pub use robj::*;
-pub use thread_safety::{handle_panic, single_threaded};
+pub use thread_safety::{handle_panic, single_threaded, throw_r_error, catch_r_error};
 pub use wrapper::*;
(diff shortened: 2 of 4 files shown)


clauswilke (approved):
does it make sense to add a function that combines this with the mutex into a single safe wrapper for R api calls? I presume we'll have to use this all over the place eventually.

andy-thomason:
I'll have to do some timings to see how much overhead it has.

The single_threaded should only be an atomic access penalty, but I'm not
sure
about the R unwind protection.

clauswilke:
Yeah, I didn't think this through carefully. Since this is protection against catastrophic errors, I think a better approach is to wrap each extendr function into this, so it's executed only once per call into extendr.

Any users that don't write extendr functions but instead write Rust standalones calling R could be expected to think about how they want to handle catastrophic R errors. They could wrap all of their entire code into one single `catch_r_error()` call.

andy-thomason:
The main problem is that some of the R functions, like `Rf_findFun` throw
errors in quite benign circumstances.

I would be very happy to wrap *all* R function calls in protection and use
Rust alternatives where possible.

It is very hard to make R multithread safe, for example, but much easier
for rust data structures where we want to
actively encourage users to use parallelism.

This is partly the motive for using the rust wrappers, rather than the
cpp++/Rcpp method of wrapping a SEXP
in a specialised class wrapper. `&[f64]` can be safely multithreaded, but
SEXP/Robj can't.

clauswilke:
Can `catch_r_error()` be nested? Can you wrap an entire call to a Rust function into it, and then separately wrap R functions that are particularly prone to throw an error, such as `Rf_findFun()`?

PR #124, opened 2021-01-13
Unify iterators

andy-thomason:
This PR fills some gaps in the set of iterators to cover environments
and to make all string iterators use the `is_na()` methods rather than `Option<&str>`.

The environment method `as_env()` now returns the parent and an iterator over elements
of  the environment.

The `PartialEq` operator only compares pointers for environments, so `enva == envb` will
just compare SEXP pointers,

The integer and real iterators return actual i32 and f64 values instead of references which
will make it easier to write stats functions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,9 @@
 # CLion support files
 **/.idea
 .vscode/tasks.json
+
+# Rstudio temps.
+.RData
+.Rhistory
+extendr.Rproj
+tests/extendrtests_0.1.0.tar.gz
--- a/extendr-api/src/lib.rs
+++ b/extendr-api/src/lib.rs
@@ -209,7 +209,7 @@ pub use lang::*;
 pub use logical::*;
 pub use rmacros::*;
 pub use robj::*;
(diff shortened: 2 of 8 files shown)


clauswilke:
There's a stack overflow error on Windows:
https://github.com/extendr/extendr/pull/124/checks?check_run_id=1697541518#step:8:92

The error on macOS seems to be related to the R setup, not your code.

andy-thomason:
The windows error is in the `rprintln` tests which use pipes. I'm not sure how well this is supported on windows.

It may be worth putting a `[cfg]` guard around the test.

andy-thomason:
We do not increase the stack size for windows, it seems:

```
            if cfg!(not(target_os = "windows")) {
                R_CStackLimit = usize::max_value();
            }
```

Is there a reason for this. Maybe a more conservative value would work, such as 32MB?

yutannihilation:
Here's a discussion about the stack size and the error:

https://github.com/extendr/extendr/issues/129#issuecomment-761196797

Issue #123, opened 2021-01-13
Generating wrapper functions directly from Rust is the wrong approach after all


clauswilke:
We need OS and build-specific modifications to the wrapper functions, and making this modifications from Rust at compile time doesn't make sense. See: https://github.com/extendr/rextendr/issues/21#issuecomment-759415572

Instead of generating wrapper functions, we should just generate a csv file or similar that contains the info required to put together the wrapper functions.

Relevant extendr-macros code is here: https://github.com/extendr/extendr/blob/master/extendr-macros/src/output_r.rs

clauswilke:
cc @hobofan

Ilia-Kosenkov:
Ideally (not sure if possible) it could be nice to have a feature-guard.
In `rextendr` the `Cargo.toml` file is written on the fly, so any features can be enabled for `extendr-macros`.

In a normal scenario `extendr-macros` will work as it does now. With the feature enabled, instead of writing `extendr_wrapper.R`, we output some `.csv` database that we can later consume.
This will not change public API (i.e. how macros are used in rust code), but will allow us to tweak the output and make better derived products.

clauswilke:
There's really no reason not to just write both all the time.

CGMossa:
I still believe that generating R wrapper functions from Rust is the right approach. I've been tinkering with it,
and found that a few modifcations to `extendr-macros` enables adding the `PACKAGE` document.

I'm writing the steps here, but will make a PR once I get os-detection correct. 

## Steps

In `write_r_wrapper` add

```rust
    let package_name = std::env::var("CARGO_CRATE_NAME")?;
```

In `to_r_wrapper` write:

```rust
    fn to_r_wrapper(&self, package_name: String) -> String { // change signature
    
    
            ".Call(\"{name}\"{leading_comma}{args}, PACKAGE = \"{package_name}\")", // amend the formatting
            
            
            package_name = package_name, // at the end of the format!-call

andy-thomason:
I will eventually get off my backside and write the metadata generator.

This will give you a list of impls and functions that you need to generate the .Call.

Note that with the new registration, you can't use "warp__rust_inc" but must use the symbol.

clauswilke:
@CGMossa @andy-thomason You're both missing a few things.

1. There are two separate use cases, (i) writing an R package containing Rust code and (ii) using `rust_source()` from rextendr. They need different wrapper functions.

[...]

PR #126, opened 2021-01-14
package_name_to_exported_wrapper

CGMossa:


CGMossa:
This is untested on macos/linux. 
It works on Windows. 

I hope either a that I can pull in some CI testing, or something tests this with rextendr on those platforms.

clauswilke:
This won't work. See here: https://github.com/extendr/extendr/issues/123#issuecomment-760383469

CGMossa:
I've thought about it, and now see what the problem is: The wrapper in R should be agnostic to whether the platform used was
windows, linux or mac. 

I would like to still suggest that that code is still part of the exported wrapper code, as the crate-name is the same, and 
one only needs to detect what platform the current R instance is on. That code can be amended to the wrapper exporter code in Rust.

I'll do that here as to experiment with it.

clauswilke:
Honestly, this is not that useful. If you want to work on this part of the code, it would be more helpful to add a code path that outputs the info required to reconstruct wrappers into a csv file. (Function names, argument names, return/no return, docstrings if available.)

andy-thomason:
I would welcome you to take a look at the metadata branch. This may make R code generation a lot easier.

clauswilke:
Yes, but see: https://github.com/extendr/extendr/pull/128#issuecomment-761631432

I think we'll still have to generate output during the macro expansion phase. Otherwise we're back to regular expressions.

andy-thomason:
I've taken this on board for the metadata wrapper builder.

I may remove the old wrapper generator as it has been subsumed into the metadata variety.

Would you mind if I closed this PR?

clauswilke:
The old wrapper generator is still being used by rextendr, though. Please don't remove it until we have ported rextendr. Relevant issue: https://github.com/extendr/rextendr/issues/28

clauswilke:
But yes, this PR can be closed I think.

andy-thomason:
Closing as this been incorporated in the metadata.

Issue #125, opened 2021-01-14
Sanitize wrapper function parameter names in R


Ilia-Kosenkov:
Currently `extendr-macros` maps parameter names from Rust to R code 1-to-1.
However, valid Rust parameter name `!=` valid R parameter name.
The simplest example is underscore, as in `_x`, whcih is valid Rust name for discarded parameter, but not valid in R.

Reprex highlights the problem

``` r
rextendr::rust_function("fn rust_test(_x : f64) {}")
#> build directory: C:\Users\[redacted]\AppData\Local\Temp\Rtmp6PuVXN\file3eb07fb729fa
#> Error in source(r_path, local = env): C:\Users\[redacted]\AppData\Local\Temp\Rtmp6PuVXN\file3eb07fb729fa/target/extendr_wrappers.R:7:23: unexpected input
#> 6:         
#> 7: rust_test <- function(_
#>                          ^
```

<sup>Created on 2021-01-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

`extendr_wrappers.R` contains the following code
```r

#===========
#THIS FILE IS AUTOGENERATED VIA #[extendr] MACROS
#DO NOT ATTEMPT TO EDIT IT BY HAND
#===========
        
rust_test <- function(_x) {
    invisible(.Call("wrap__rust_test", _x))
}
```

This can be easily resolved if problematic parameter names a wrapped in backticks in R, like `_x`.

PR #128, opened 2021-01-14
Basic function and struct metadata.

andy-thomason:
This PR provides metadata that will eventually replace the init__ functions.

The metadata is accessible from Rust or R and can be used to register functions,
build call and class wrappers and generate documentation.

This is the new `cargo expand` on the hello example:

```
#![feature(prelude_import)]
#[prelude_import]
use std::prelude::v1::*;
#[macro_use]
extern crate std;
use extendr_api::*;
fn hello() -> &'static str {
    "hello"
}
#[no_mangle]
#[allow(non_snake_case)]
pub extern "C" fn wrap__hello() -> extendr_api::SEXP {
    unsafe {
        use extendr_api::FromRobj;
        extendr_api::handle_panic("hello paniced.\u{0}", || {
            extendr_api::Robj::from(hello()).get()
        })
    }
}
#[allow(non_snake_case)]
fn init__hello(info: *mut extendr_api::DllInfo, call_methods: &mut Vec<extendr_api::CallMethod>) {
    call_methods.push(extendr_api::CallMethod {
        call_symbol: std::ffi::CString::new("wrap__hello").unwrap(),
        func_ptr: wrap__hello as *const u8,
        num_args: 0i32,
    })
}
#[allow(non_snake_case)]
fn meta__hello(metadata: &mut Vec<extendr_api::metadata::Func>) {
    let mut args = ::alloc::vec::Vec::new();
    metadata.push(extendr_api::metadata::Func {
        doc: "",
        name: "hello",
        args: args,
        return_type: "-> & \'static str",
        func_ptr: wrap__hello as *const u8,
    })
}
#[no_mangle]
#[allow(non_snake_case)]
pub fn get_hello_metadata() -> extendr_api::metadata::Metadata {
    let mut functions = Vec::new();
    let mut structs = Vec::new();
    meta__hello(&mut functions);
    extendr_api::metadata::Metadata {
        doc: "",
        functions,
        structs,
    }
}
#[no_mangle]
#[allow(non_snake_case)]
pub extern "C" fn wrap__get_hello_metadata() -> SEXP {
    unsafe { Robj::from(get_hello_metadata()).get() }
}
#[no_mangle]
#[allow(non_snake_case)]
pub extern "C" fn R_init_hello_extendr(info: *mut extendr_api::DllInfo) {
    let mut call_methods = Vec::new();
    init__hello(info, &mut call_methods);
    unsafe { extendr_api::register_call_methods(info, call_methods.as_ref()) };
}
```


andy-thomason:
Next step is to add some tests and remove the old `init_` functions and pass the metadata directly to `register_call_methods`.

clauswilke:
Any idea what's going on with the failing Windows test? All the recent PRs have the same problem, so it doesn't seem to be specific to this PR.

extendr-api/src/metadata.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub name: &'static str,
+    pub methods: Vec<Func>,
+}
+
+/// Module metadata.
+pub struct Metadata {
+    pub doc: &'static str,
+    pub functions: Vec<Func>,
+    pub structs: Vec<Struct>,
+}
```
clauswilke, line 36: I think the module name should also be reported.

clauswilke (approved):
test failures seem to be intermittent and can be ignored.

andy-thomason:
Metadata is now working from the Rust side.

In theory, you should be able to `.Call(wrap__get_my_module_metadata)` from the R side
to get an R named list with functions and impls.

I've tightened up the `#[extendr] impl` rules for now. Only impl <typename> is allowed.

The metadata contains doc strings which can be used to generate R documentation.
These are direct copies of the Rust doc strings and so may need some conversion.

The metadata can also be used to generate class wrappers in whatever flavour we like.

The list of impls contains list of methods which could be bound to an environment.

We could then do:

person <- new("Person")
person$name()

or use s3 classes:

person <- news3("Person")
name(person)

Of course, we would have to generate "new" and "news3".

clauswilke:
Just to be clear: I think this is good to have in the code, but by itself it doesn't solve the various problem for rextendr. I would have to go back to writing regexes to parse Rust code to use these functions:

1. `rust_function()`: For this use case, I cannot know the name of the function without somehow parsing the input source code. I also don't want to require the user to add a module call to the code, so I would somehow have to generate it, and again that requires knowing the name of the function.

2. `rust_source()`: For this use case, I could require the user to add a module call, but then I don't know the name of the module without parsing the input source code, or having the user provide the name in a separate argument, which is awkward.

Both of these problems are avoided if the `extendr` macros write out data as they go, as is currently implemented.

clauswilke:
Would it be possible to add a function that always has the same name (regardless of the module name) to discover the name(s) of any modules that have been defined? I believe functions in different dynamic libraries can have the same name, and we can call them by explicitly providing the name of the dynamic library in `.Call()`.

andy-thomason:
The problem is that the procedural macros are not allowed to have any shared state and **should** not do I/O. We are doing this at the moment, but it is not guaranteed to work forever. For example, the compiler could run the in parallel.

I could make the metadata function just "get_module_metadata" and then you can extract the
metadata with a call to the module.

I'll do an example of this today.

I am especially keen to let us wrap libraries that are not designed for use with R without extra work.

We could also add a "generate_wrappers" function exported to R.

clauswilke:
I think this would work. It removes the need for regex parsing in `rust_source()`, and I think it's fine to regex in `rust_function()`, as that's a must more constrained problem.

andy-thomason:
The function registration now uses metadata instead of the old "init__" functions.

andy-thomason:
We can now generate wrapper code from R and even execute it directly.

```
eval(parse(text = make_extendrtests_wrappers(use_symbols = FALSE)))
```

True use_symbols will generate code for use in libraries, if set to false
it will use strings.

In the example above, we use strings, so we can call the code interactively.

clauswilke:
I'm not entirely convinced it makes sense to generate wrapper code from Rust, as it needs to be adjusted to the specific needs of the R use case. E.g., it needs the correct library name when used from rextendr, as is done here:
https://github.com/extendr/rextendr/blob/a4253f33e028d141501f5bb83846dd34233241ed/R/source.R#L168-L173

clauswilke (commented):
To get the integration tests to pass, you need to temporarily point cargo to your github branch here:
https://github.com/extendr/extendr/blob/7d302589176015503e286ac12ce2dfcdcb1fbae1/tests/extendrtests/src/rust/Cargo.toml#L19-L21

extendr-api/src/lib.rs
```diff
@@ -622,4 +650,26 @@ mod tests {
         assert_eq!("NA".is_na(), false);
         assert_eq!(na_str().is_na(), true);
     }
+    
+    
+    #[test]
+    fn metadata_test() {
```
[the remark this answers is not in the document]
andy-thomason: There is a `#[cfg]` further up.

We should move all the tests to their own modules. I'm quite unhappy about the complexity
of the project layout now.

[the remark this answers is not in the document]
clauswilke: Somehow testing code made it into the final library though: 
https://github.com/extendr/extendr/runs/1717739787#step:13:106

```
File ‘extendrtests/libs/extendrtests.so’:
  Found non-API calls to R: ‘R_CStackLimit’, ‘R_CleanTempDir’,
    ‘R_RunExitFinalizers’, ‘Rf_initialize_R’, ‘setup_Rmainloop’

```
This normally happens when you compile an `init_r()` call into your library.

clauswilke:
Sorry, I should have been more precise. This won't work on GH Actions:
https://github.com/extendr/extendr/blob/a77d863907a11ae21e410d18fb713ef1ee5bc03a/tests/extendrtests/src/rust/Cargo.toml#L19-L21
These relative paths don't exist there.

You need to write something like:
```
extendr-api = { git = "https://github.com/extendr/extendr", branch = "metadata"}
extendr-engine = { git = "https://github.com/extendr/extendr", branch = "metadata" }
extendr-macros = { git = "https://github.com/extendr/extendr", branch = "metadata" }
```
See: https://github.com/extendr/extendr/issues/89

andy-thomason:
That is a shame. I did this earlier...

clauswilke:
Yeah, it basically works now. You just have a few minor issues remaining in the integration tests:

1. Undocumented objects:
https://github.com/extendr/extendr/pull/128/checks?check_run_id=1718200375#step:13:75

2. Testing code somehow ended up in the compiled library:
3. https://github.com/extendr/extendr/pull/128/checks?check_run_id=1718200375#step:13:93

tests/extendrtests/R/wrappers.R
```diff
@@ -1,8 +1,19 @@
...
+
+#' Get module metadata as a list.
+#' 
+#' @rdname wrappers
+#' @export
+get_extendrtests_metadata <- function() .Call(wrap__get_extendrtests_metadata)
+
+#' Build wrappers 
+#' 
+#' @rdname wrappers
```
clauswilke, line 17: This will need a line such as the following to keep R happy:
```
#' @param use_symbols ... some explanation here ...
```

clauswilke:
To get rid of the undocumented objects error, you'll have to rebuild the documentation with `devtools::document()`. But you'll also have to document the `use_symbols` argument (see other comment I made).

clauswilke:
It still requires running `devtools::document()` in the extendrtests package to rebuild the `.Rd` files. Otherwise R will continue to complain about undocumented symbols.

And some testing code still makes it into the compiled library and needs a conditional compilation guard.

clauswilke:
Please see PR #135. Relative paths work after all in extendrtests Cargo.toml, they just need to be a bit different from what one might expect naively.

clauswilke:
Once you're ready, if problems with integration tests remain, I'm happy to go in and fix up this branch. Just let me know.

andy-thomason:
Thanks, Claus.

Your wisdom is appreciated.

clauswilke:
@andy-thomason Could you make the metadata functions public, at least in the crate? See here:
https://github.com/extendr/extendr/issues/140#issuecomment-763559061

cc @dfalbel

extendr-api/src/metadata.rs
```diff
@@ -99,13 +93,24 @@ fn write_r_wrapper(w: &mut Vec<u8>, func: &Func, package_name: &str, use_symbols
...
+
+fn write_function_wrapper(w: &mut Vec<u8>, func: &Func, package_name: &str, use_symbols: bool) -> std::io::Result<()> {
+    if func.hidden {
+        return Ok(());
+    }
+
+    write_doc(w, func.doc)?;
+
+    let args = func.args.iter().map(|arg| arg.name).collect::<Vec<_>>().join(", ");
```
Ilia-Kosenkov, line 106: Because of #125, is it possible to input an additional check here?
From Rust guidelines (if I understand correctly) https://doc.rust-lang.org/reference/identifiers.html,
valid parameter names can start with the underscore `_` (which is not useful, but allowed).
If such parameter name is emitted into R, the resulting R code is invalid, as R does not allow identifiers to start with `_`.
Can we replace map with a function that matches first character and if it is `_`, wraps name into backquotes, like ``` `_y` ```.

Here is my [improvised attempt](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=736766f593c7651f144ad387707df182).

andy-thomason:
So nearly  there!

Thanks to @dfalbel I have just one problem to solve with the class wrappers.

The registration settings allow both symbols (packages) and strings (interactive)

MyClass$new() returns a pointer as expected with class "MyClass"

```
> library(extendrtests)
> p <- MyClass$new()
> p
<pointer: 0x559681d60490>
attr(,"class")
[1] "MyClass"
```
The dollar overload:
```
#' @export
`$.MyClass` <- function (self, name) { func <- MyClass[[name]]; environment(func) <- environment(); func }
```

Converts p$a into a modified function that steals "self" from p.

Now `p$a()` calls the function with self set.

I'm guessing copy-on-write de-duplicates the function.

It all works using metadata generated code:

```
MyClass <- new.env(parent = emptyenv())

MyClass$new <- function() .Call(wrap__MyClass__new)
MyClass$set_a <- function(x) .Call(wrap__MyClass__set_a, self, x)
MyClass$a <- function() .Call(wrap__MyClass__a, self)
MyClass$me <- function() .Call(wrap__MyClass__me, self)
#' @export
`$.MyClass` <- function (self, name) { func <- MyClass[[name]]; environment(func) <- environment(); func }
```

The only remaining problem is getting the dollar operator to export.

clauswilke:
Does this solve your issue exporting `$`?

https://stackoverflow.com/a/29639694/4975218

andy-thomason:
I just had to run the "document" action in R studio and it worked!

Thanks, Claus.

tests/extendrtests/src/rust/Cargo.toml
```diff
@@ -16,7 +16,13 @@ crate-type = ["staticlib"]
...
+
+# development verion
+#extendr-api = { path = "../../../../extendr-api/"}
+#extendr-engine = { path = "../../../../extendr-engine/"}
+#extendr-macros = { path = "../../../../extendr-macros/"}
+
+# R CMD check
+extendr-api = { git = "https://github.com/extendr/extendr", branch = "metadata"}
+extendr-engine = { git = "https://github.com/extendr/extendr", branch = "metadata" }
+extendr-macros = { git = "https://github.com/extendr/extendr", branch = "metadata" }
```
clauswilke, line 28: Could you use these relative paths:
https://github.com/extendr/extendr/blob/10d0ab8a64ab4fafb125d69cf555f08b5b090593/tests/extendrtests/src/rust/Cargo.toml#L18-L21

You can then check locally as described here (and GH Actions will work also):
https://github.com/extendr/extendr/blob/master/tests/extendrtests/README.md#running-tests-locally

clauswilke:
I'm glad things are moving forward. The integration tests almost pass, only a few minor issues remaining.

In particular, if you ran "document" there should be an updated `.Rd` file in this folder:
https://github.com/extendr/extendr/tree/master/tests/extendrtests/man
I think it's missing from the commits.

tests/extendrtests/R/extendrtests.R
```diff
@@ -1,7 +1,44 @@
...
 #' @useDynLib extendrtests, .registration = TRUE
-NULL
\ No newline at end of file
+NULL
+
+# Generated by
+#
+# cat(.Call("wrap__make_extendrtests_wrappers", use_symbols=TRUE))
+
+# '@export
```
clauswilke, line 10: Notice that the `'` is in the wrong location. You have a space between `#` and `'`, `# '`, instead of `#' `. This means the `@export` directive is ignored. The tests pass because testthat has access to unexported functions also.

extendr-api/src/metadata.rs
```diff
@@ -162,20 +161,31 @@ fn write_method_wrapper(w: &mut Vec<u8>, func: &Func, package_name: &str, use_sy
...
+            writeln!(w, "# '@export")?;
             writeln!(w, "{} <- new.env(parent = emptyenv())\n", imp.name)?;
                         
             for func in &imp.methods {
+                // write_doc(& mut w, func.doc)?;
                 write_method_wrapper(&mut w, func, self.name, use_symbols, imp.name)?;
             }
 
-            writeln!(w, "#' @export")?;
+            writeln!(w, "# '@export")?;
```
clauswilke, line 188: See comment below. Was this change on purpose? You're basically disabling the export.

tests/extendrtests/R/extendrtests.R
andy-thomason: Just as I thought it was all working!

Issue #129, opened 2021-01-15
Windows tests are failing


clauswilke:
Several of the recent PRs have Windows tests failing with a stack overflow:
```
Error: Error: C stack usage  2101032 is too close to the limit
error: test failed, to rerun pass '--lib'
Error: i686-pc-windows-gnu
```
For example: https://github.com/extendr/extendr/pull/128/checks?check_run_id=1705514593#step:8:95

Since this seems to be happening with all recent PRs, I suspect it's not related to them but some other cause.

clauswilke:
For example, there is no good reason why #126 should introduce a new stack overflow.

clauswilke:
@CGMossa @Ilia-Kosenkov Any ideas?

clauswilke:
Maybe it's happening because we're no longer running tests single-threaded?

Ilia-Kosenkov:
@clauswilke,
I am unable to reproduce it on my local machine.
Right now the issue is transient, so we first need to reliably reproduce it.
I will try running it on CI to see if I can pinpoint the problem.

yutannihilation:
I'm curious if the error also happens with the release version of R. If not, it might be related to some recent changes on R-devel.

Btw, (probably this is unrelaetd,) I couldn't pass tests of i686 on my local (x86_64 succeeds). I think I've followed the instruction on libR-sys's README and mimicked what GHA does. Am I doing something wrong...?

``` console
> $env:MSYS_ROOT="C:\msys64\"
> $env:LIBCLANG_PATH="$env:MSYS_ROOT\mingw64\bin"
> $env:PATH="$env:MSYS_ROOT\usr\bin;$env:MSYS_ROOT\mingw32\bin;$env:MSYS_ROOT\mingw64\bin;$env:PATH"
> cargo test --manifest-path extendr-api/Cargo.toml --target i686-pc-windows-gnu --features tests-all -- --no-capture
   Compiling libR-sys v0.2.1
   Compiling extendr-engine v0.1.11 (C:\path\to\extendr\extendr-engine)
   Compiling extendr-api v0.1.11 (C:\path\to\extendr\extendr-api)
    Finished test [unoptimized + debuginfo] target(s) in 18.26s
     Running target\i686-pc-windows-gnu\debug\deps\extendr_api-c8bf11d0bcd2920b.exe
error: test failed, to rerun pass '--lib'

Caused by:
[...]

PR #130, opened 2021-01-15
Better & more informative logging and error handling

Ilia-Kosenkov:
After digging into #129 I noticed that the current output of GH Actions is not really friendly for debugging purposes, especially when multiple commands are executed for several architectures in one step.

Using GHA syntax `"::group::"` I slightly improved how logs are displayed (see Check results especially for Windows bindgen).
I also added missing error handling to Build & Test steps, which previously could swallow some build/test errors.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -120,8 +120,14 @@ jobs:
       - name: Build
         run: |
           foreach($target in ($env:BUILD_TARGETS).Split(',')) {
-            echo $target
+            echo "::group::Build for $target target"
             cargo build $(if($target -ne 'default') {"--target=$target"} ) 
+            echo "::endgroup::"
+
+            if (!$?) { 
+              echo "::error::Build for $target target" ;
+              throw "Last exit code $LASTEXITCODE"
+            }
           }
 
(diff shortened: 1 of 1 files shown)


clauswilke (approved):

.github/workflows/test.yml
```diff
@@ -120,17 +120,29 @@ jobs:
       - name: Build
         run: |
           foreach($target in ($env:BUILD_TARGETS).Split(',')) {
-            echo $target
+            echo "::group::Build for $target target"
             cargo build $(if($target -ne 'default') {"--target=$target"} ) 
+            echo "::endgroup::"
```
yutannihilation, line 125: Should this line be after the next `if` branch? `echo` always succeeds, which probably means we cannot detect build/test errors.

clauswilke: I think you're right. Thanks for catching this!

yutannihilation: I just found this while running the tests that should fail :)

https://github.com/extendr/extendr/issues/129#issuecomment-761350724

Sure, I'll create a pull request for this.

PR #134, opened 2021-01-18
use make_name instead of quote to get the impl nm

### LINKED ISSUE #131: Exporting classes with lifetime annotation
I have a `impl` that I'd like to export that includes lifetime annotations, eg:

```
struct Person<'a> {
    pub name: &'a String,
}

#[extendr]
impl<'a> Person<'a> {
    fn new() -> Self {
        Self { name: &"".to_string() }
    }

    fn set_name(&mut self, name: &str) {
        self.name = &name.to_string();
    }

    fn name(&self) -> &str {
        self.name.as_str()
    }
}
```

Compilation of this file fails with:

```
custom attribute panicked

help: message: `"init__Person  \'a __new"` is not a valid identifier
```

It seems to me that this happens because this only happens a invalid function name is generated here:

https://github.com/extendr/extendr/blob/master/extendr-macros/src/lib.rs#L319-L321

Do you think it's fixable by tweaking this line to strip those invalid characters? I'd be happy to contribute a PR if this sounds like a reasonable solution.

dfalbel:
This PR allows to export `impl`s that have lifetime annotations (to fix #131) like the one below:

```
struct Person<'a> {
    pub name: &'a String,
}

#[extendr]
impl<'a> Person<'a> {
    fn new() -> Self {
        Self { name: &"".to_string() }
    }
}
```

See discussion in #131

I am happy to add tests but I couldn't find out where I should add them, if in the R package or somewhere in the `extendr-macros` directory.

### DIFF
--- a/extendr-macros/src/lib.rs
+++ b/extendr-macros/src/lib.rs
@@ -283,6 +283,35 @@ fn generate_wrappers(_opts: &ExtendrOptions, wrappers: &mut Vec<ItemFn>, prefix:
     ));
 }
 
+fn make_name (ty: &syn::Type) -> String {
+
+    // c.f. https://docs.rs/syn/1.0.58/syn/enum.Type.html#variants
+    let path: &syn::Path = match ty {
+        syn::Type::Path(p) => {
+            // <Type as Trait>::function(...), which should not appear in impl
+            if p.qself.is_some() {
+                panic!("Something is wrong");
+            }
+
+            &p.path
+        }
+        _ => {
+            panic!("Unknown type")
+        }
+    };
+
+    // ::Foo
+    if path.leading_colon.is_some() {
+        panic!("There shouldn't be leading colons?")
+    }
+
+    if path.segments.len() > 1 {
+        panic!("Namespaced impl are not supported and connot be exported.")
+    }
+
+    path.segments[0].ident.to_string()
+}
+
 /// Handle trait implementations.
 ///
 /// Example:
@@ -317,7 +346,7 @@ fn generate_wrappers(_opts: &ExtendrOptions, wrappers: &mut Vec<ItemFn>, prefix:
 fn extendr_impl(mut item_impl: ItemImpl) -> TokenStream {
     let opts = ExtendrOptions {};
     let self_ty = item_impl.self_ty.as_ref();
-    let self_ty_name = quote! {#self_ty}.to_string();
+    let self_ty_name = make_name(self_ty);
     let prefix = format!("{}__", self_ty_name);
     let mut method_init_names = Vec::new();
     let mut wrappers = Vec::new();
--- a/extendr-macros/src/lib.rs
+++ b/extendr-macros/src/lib.rs
@@ -283,6 +283,8 @@ fn generate_wrappers(_opts: &ExtendrOptions, wrappers: &mut Vec<ItemFn>, prefix:
     ));
 }
 
+// Extracts the `impl` name from a `syn::Type`. It's used to make names for 
+// the C init wrappers.
 fn make_name (ty: &syn::Type) -> String {
 
     // c.f. https://docs.rs/syn/1.0.58/syn/enum.Type.html#variants
@@ -290,19 +292,19 @@ fn make_name (ty: &syn::Type) -> String {
         syn::Type::Path(p) => {
             // <Type as Trait>::function(...), which should not appear in impl
             if p.qself.is_some() {
-                panic!("Something is wrong");
+                panic!("This doesn't look to be an `impl`.");
             }
 
             &p.path
         }
         _ => {
-            panic!("Unknown type")
+            panic!("Unknown type.")
         }
     };
 
     // ::Foo
     if path.leading_colon.is_some() {
-        panic!("There shouldn't be leading colons?")
+        panic!("There shouldn't be leading colons.")
     }
 
     if path.segments.len() > 1 {
--- a/extendr-macros/src/lib.rs
+++ b/extendr-macros/src/lib.rs
@@ -306,7 +306,7 @@ fn make_name (ty: &syn::Type) -> String {
     }
 
     if path.segments.len() > 1 {
-        panic!("Namespaced impl are not supported and connot be exported.")
+        panic!("Namespaced impl are not supported and cannot be exported.")
     }
 
     path.segments[0].ident.to_string()


extendr-macros/src/lib.rs
```diff
@@ -283,6 +283,35 @@ fn generate_wrappers(_opts: &ExtendrOptions, wrappers: &mut Vec<ItemFn>, prefix:
     ));
 }
 
+fn make_name (ty: &syn::Type) -> String {
```
clauswilke, line 286: Could you add a brief comment here explaining what the function does?

extendr-macros/src/lib.rs
```diff
@@ -283,6 +283,35 @@ fn generate_wrappers(_opts: &ExtendrOptions, wrappers: &mut Vec<ItemFn>, prefix:
...
+            &p.path
+        }
+        _ => {
+            panic!("Unknown type")
+        }
+    };
+
+    // ::Foo
+    if path.leading_colon.is_some() {
+        panic!("There shouldn't be leading colons?")
```
clauswilke, line 305: Please remove the question mark at the end of the panic message.

clauswilke (commented):
I think this is good in principle. A couple of comments:

- Please review the panic messages and see if you can make them a bit more informative. In particular, "something is wrong" is not a great message.

- I think there should be a simple test in extendr-macros testing this function directly.

- If you also want to add an integration test, that would be great. We don't have a lot of those yet, so any new test there adds a lot of different aspects of the project.

yutannihilation:
(For the panic messages, I'm sorry my example was too casual...)

>     * I think there should be a simple test in extendr-macros testing this function directly.

I have no experience of writing tests for procedural macros, but, googling around, it doesn't seem to be very simple because procedural macros cannot be used within the crate itself. The choices I found are:

* Create a dedicated crate for testing (c.f. https://blog.logrocket.com/procedural-macros-in-rust/)
* Use [`proc_macro2`](https://docs.rs/proc-macro2/1.0.24/proc_macro2/)

clauswilke:
Regarding testing:

- Isn't `make_name()` just a regular Rust function that can be tested directly with some input and associated output?
- I think `syn` already pulls in `proc_macro2`, so there wouldn't be any penalty from using it directly. This should be checked, though.
- Tests of macros could also go into `extendr-api`.

dfalbel:
I'll work on tests in the coming days... Thanks for the feedback!

If I were to do that:

> Isn't make_name() just a regular Rust function that can be tested directly with some input and associated output?

Where should I write this test?

clauswilke:
> Where should I write this test?

The Rust convention is that unit tests go into the same source file into a test module:
https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-module-and-cfgtest

dfalbel:
Turns out that while writing tests I figured it doesn't work for the desired use case, ie. exporting `impl` with lifetime annotations because there many other parts in the extendr macro that would need to be changed, for example the generated trait for `From` doesn't declare the lifetime annotations.

```
impl From<Animal<'a>> for Robj {
    fn from(value: Animal<'a>) -> Self {
        unsafe {
            let ptr = Box::into_raw(Box::new(value));
            let res =
                Robj::makeExternalPtr(ptr, Robj::from("Animal"),
                                      Robj::from(()));
            res.registerCFinalizer(Some(__finalize__Animal));
            res
        }
    }
}
```

We would probably also need to handle lifetime 'conflicts' when the lifetime annotation already exists in the trait, eg:

```
impl <'a> extendr_api::FromRobj<'a> for &mut Animal<'a> {
    fn from_robj(robj: &'a Robj) -> std::result::Result<Self, &'static str> {
        if robj.check_external_ptr("Animal") {
            Ok(unsafe {
                   std::mem::transmute(robj.externalPtrAddr::<Animal<'a>>())
               })
        } else { Err("expected Animal") }
    }
}
```

I don't have knowledge and time to implement this... So, If you think that it's still valuable to switch from `quote` to this, I would be happy to catch and return a nice error message when a lifetime annotated `impl` is exported. Otherwise we can close this PR. - And I'm sorry for spending your time on it :(

clauswilke:
I think any future work on this will require something like the `make_name` function, so in my opinion we should finish off this PR and merge it. Separately, if you could open an issue explaining what you have found and what is still missing, we at least have a framework for how to continue working on this.

yutannihilation:
> Isn't `make_name()` just a regular Rust function that can be tested directly with some input and associated output?

Ah, yes. Maybe I was thinking too complicated, sorry for the noise.

andy-thomason:
Thanks for looking at this.

If you give me a test case, I can have a look at this this evening.

In the metadata branch, which is a significant rewrite of the macro code, I
have explicitly limited impls to untemplated ones prior to doing
a specialisation.

The (unwritten admittedly) plan was to allow specialisations in the module
declaration:

Something like

struct Fred<T> {
x: T
}

#[extendr]
impl<T> Fred {
}

extendr_module! {
mod mymod;
impl Fred<i32> as FredInt;
impl Fred<f64> as FredReal;
}

This would also apply to lifetimes, which are part of the generic params
and constants
which are a work in progress in Rust.

An example of this in hand generated code is:

impl<'a> From<IntegerIter<'a>> for Robj {
/// Convert an integer iterator into a vector.
fn from(val: IntegerIter) -> Self {
val.collect_robj()
}
}

dfalbel:
Oh I see! It would be cool to be able to export specializations too. 
I think the minimal example would be something like this:

```
struct Animal<'a> {
    pub name: &'a str,
}

#[extendr]
impl<'a> Animal<'a> {
    fn new() -> Self {
        Self { name: "" }
    }

    fn set_name(&mut self, name: &'a str) {
        self.name = name;
    }

    fn name(&self) -> &str {
        self.name
    }
}
```

andy-thomason:
I've made a function  that converts many kinds of type to an R-friendly string.

This converts many types to a single name.

& Fred -> Fred
&& Fred -> Fred
& mut Fred -> Fred
& 'a Fred<'t> -> Fred

etc.

@Ilia-Kosenkov  suggested we make a function for identifiers also.

andy-thomason:
I've deferred the lifetime and template expansion issue to later.

We could use `type AnimalStatic = Animal<'static>

dfalbel:
Ok! Let's close this PR, we can find this code if we need that in the future.

PR #135, opened 2021-01-18
Using local extendr instead of Github's 

### LINKED ISSUE #89: Integration tests create GitHub merge chicken-and-egg problem

dfalbel:
This tries to fix #89 .
It's not pretty, but should work... Let me know what do you think.


clauswilke:
I'm not convinced. This seems to work by accident rather than due to some underlying principle.

Can you still run checks on the package locally, by opening it in RStudio and clicking on "Check package"?

dfalbel:
This works because the package is checked in this path: ` /home/runner/work/extendr/extendr/`.
I can't run locally because RStudio will use the package directory instead of `extendr` directory for checking, but I think this is fixable by running the rcmdcheck command from inside tests/extratests instead of the project's root here:

https://github.com/extendr/extendr/blob/master/.github/workflows/test.yml#L177

and tweaking the paths again.

dfalbel:
Actually this can't be fixed is RStudio, because it runs `devtools::check()` and this will always use a `tempdir()` but. It would work if we instead of clicking Check in Rstudio, typed `devtools::check(check_dir = ".")` or something similar.

clauswilke:
Ok, if you think this is somewhat reliable and workable, could you document this in the readme here:
https://github.com/extendr/extendr/blob/master/tests/extendrtests/README.md
Including how to run the tests locally?

dfalbel:
@clauswilke updated docs for running locally and the workflow so we can run both easily. What do you think?

tests/extendrtests/.gitignore
```diff
@@ -37,3 +37,6 @@ vignettes/*.pdf
 
 # R Environment Variables
 .Renviron
+
+# check dir
+check/
```
clauswilke, line 42: Could you add a newline here?

tests/extendrtests/README.md
```diff
@@ -10,4 +10,18 @@ The Rust code that is part of the package is located here: https://github.com/ex
...
+The Cargo.toml file hardcodes the relative path of the `extendr` libraries, thus
+in order to check this project locally you need to run:
+
+```r
+devtools::check(check_dir = "check")
+```
+
+This is necessary so the relative path in Cargo.toml points to the correct
+location. **Note**: Clicking 'Check' in RStudio will not work because it runs
+`devtools::check()` with no arguments and the check are then run in a temporary
```
clauswilke, line 26: and the *checks* are then run

tests/extendrtests/README.md
```diff
@@ -10,4 +10,18 @@ The Rust code that is part of the package is located here: https://github.com/ex
...
\ No newline at end of file
+The test functions that verify that the wrapper and Rust functions work correctly are located here: https://github.com/extendr/extendr/blob/master/tests/extendrtests/tests/testthat/test-wrappers.R
+
+## Running tests locally
+
+The Cargo.toml file hardcodes the relative path of the `extendr` libraries, thus
+in order to check this project locally you need to run:
+
+```r
+devtools::check(check_dir = "check")
```
Ilia-Kosenkov, line 21: Perhaps it should be `rcmdcheck::rcmdcheck` as we use it on CI?

dfalbel:
I just resolved the comments.

tests/extendrtests/README.md
```diff
@@ -18,10 +18,10 @@ The Cargo.toml file hardcodes the relative path of the `extendr` libraries, thus
 in order to check this project locally you need to run:
 
 ```r
-devtools::check(check_dir = "check")
+rcmdcheck::check(check_dir = "check")
```
Ilia-Kosenkov, line 21: I am afraid it should be `rcmdcheck::rcmdcheck`
https://github.com/extendr/extendr/blob/702b24ba9e871010779cf28906fedffc3b34802a/.github/workflows/test.yml#L177

tests/extendrtests/README.md
```diff
@@ -18,10 +18,10 @@ The Cargo.toml file hardcodes the relative path of the `extendr` libraries, thus
...
 ```
 
 This is necessary so the relative path in Cargo.toml points to the correct
 location. **Note**: Clicking 'Check' in RStudio will not work because it runs
-`devtools::check()` with no arguments and the check are then run in a temporary
+`devtools::check()` with no arguments and the checks are then run in a temporary
```
Ilia-Kosenkov, line 26: Also, if you change code fragment above to `rcmdcheck::rcmdcheck`, do it also here.

tests/extendrtests/README.md
dfalbel: ahh right! missed that! Should be fixed now!

tests/extendrtests/README.md
dfalbel: RStudio actually calls `devtools::check()` so I'm leaving that but added parenthesis explaining. `devtools::check` is really a wrapper for `rcmdcheck::rcmdcheck` so there shouldn't be any differences in the end.

clauswilke (approved):
Thanks, looks good to me. @andy-thomason will be happy about being able to use relative paths.

tests/extendrtests/README.md
Ilia-Kosenkov: @dfalbel,
My mistake, you are correct.

Ilia-Kosenkov (approved):
Looks good, I hope it will simplify testing of new PRs.

clauswilke:
It looks like this commit was accidentally reverted in the main branch. However, this made me realize one issue with this commit: It not only disables the "check" option in RStudio, it also disables the "Install and Restart" option. Is there a one-liner command to build and install the package?

clauswilke:
Maybe this can be resolved by running the checks in a folder parallel to extendrtests, so the relative paths remain the same? Something like `../check` instead of `check`?

PR #136, opened 2021-01-19
From vec string

dfalbel:
This PR allows automatic conversion of R character vectors to Rust's `Vec<String>`.
Here are a few asserts from tests:

```
fn char_vec(x: Vec<String>) -> Vec<String> {x}
```

```
expect_equal(.Call(wrap__char_vec, c("hello", "world")), c("hello", "world"))
expect_error(.Call(wrap__char_vec, .45), "not a character vector")
expect_error(.Call(wrap__char_vec, 15L), "not a character vector")
expect_error(.Call(wrap__char_vec, TRUE), "not a character vector")
expect_error(.Call(wrap__char_vec, NA_character_), "Input must not be NA")
expect_error(.Call(wrap__char_vec, c("hello", NA)), "Vector cannot contain NA's")
```

What do you think?


clauswilke:
Looks good to me in principle, but @andy-thomason should take a look. In particular, do we reject `NA`s even though we have a special Rust string that can represent them? (I feel the answer should be yes, as otherwise it's inconsistent with other variable types, but I'm not 100% sure.)

dfalbel:
Yeah, I wasn't sure about that too. 
IMHO, users should explicitly allow NA's by using a `Vec<Option<String>>` type. It would look more idiomatic to me, but it's the first time I use Rust so ..

clauswilke:
Agreed. And it's the only pattern that is reliable for other data types, e.g. doubles.

Ilia-Kosenkov:
I was about to complain how strings passed to `Err` are formatted -- i.e. not capitalized and sentences not ending with e.g. a period, but then I noticed that pretty much all `Err` calls in `from_robj.rs` have arbitrary string formatting/capitalization.

Should we unify how error messages are formed throughout the project? If yes, then `Err` calls introduced in this PR should be corrected to satisfy unified requirements.

dfalbel:
I agree! Is there any style guide that is generally used?

Ilia-Kosenkov:
@dfalbel, I am unfortunately unaware if we have any applied to extendr projects.
In my opinion though, I'd prefer to have properly capitalized sentences. Not sure about punctuation at the end. For instance, if an error is intercepted and a message is incorporated into a string displayed to a user, we end up in a situation like this:
```R
glue("Error on the Rust side: '{err_msg}'.")
# > "Error on the Rust side: 'Input must not be NA.'.
```

clauswilke:
The beauty of joining a project in its infancy is you can make up the rules. :-)

In general, I would suggest following tidyverse style, unless there's a specific reason to deviate from those guidelines in the context of R-Rust integration: https://style.tidyverse.org/error-messages.html

dfalbel:
Great! I have capitalized and added punctuation. 

I am happy do to do a follow up PR fixing the error messages for the other objects.

@Ilia-Kosenkov I think if we want to do something like that in the future we can follow the tidyverse style guide and error with a bullet list like:

```
#> Error in Rust side:
#> ✖ Input must not be NA.
```

Ilia-Kosenkov:
@dfalbel,
I suggest you open a separate issue so we can discuss this topic and inform other contributors -- `tidyverse` guideline is an obvious choice but there might be nuances.

Your updated error messages look much better now and are consistent within the function, so it's a good start.

andy-thomason (approved):
Looks good to me.

andy-thomason:
The tests are failing because the R test is pulling from master.

PR #141, opened 2021-01-23
Clean up extendrtests package issues

clauswilke:
Clean-up to make sure extendrtests package remains consistent going forward.


andy-thomason (approved):
Nice one @clauswilke 

It would be good to have a Cargo.toml that works in all branches.

clauswilke:
@andy-thomason @dfalbel Could you look at the latest version of this PR? I think it is the best compromise for local paths in extendrtests. It allows building of the package from RStudio by clicking on "Install and Restart" but requires a manual call to `rcmdcheck()` for checking. I think this is workable. Downside is that the check info is written in parallel to the main `extendr` folder. This was needed due to the folder structure that `rcmdcheck()` uses.

@yutannihilation I see code formatting failures that I can't reproduce locally. Could you check if you understand what's going on there? 

https://github.com/extendr/extendr/pull/141/checks?check_run_id=1755169261

The problem I have is I cannot find the incorrectly formatted line of code in my sources.

clauswilke:
@yutannihilation Never mind, I thought I had the latest version of master but apparently I missed a commit. I see now where the formatting problem comes from.

dfalbel:
Good idea! Looks good to me!

PR #146, opened 2021-01-24
Make extendr-engine dev-dependency

### LINKED ISSUE #144: Some tests are not conditionally compiled
Currently the integration tests trigger a note that the compiled code contains non-API entry points:
https://github.com/extendr/extendr/pull/141/checks?check_run_id=1755212957#step:14:90

This isn't reported as an error, so tests seem to pass, but it would actually prevent us from submitting a package to CRAN, so it needs to be fixed.

One case where we have tests without conditional compilation is here:
https://github.com/extendr/extendr/blob/f43d2539e8ac0d5dcd562aa1d2282369528ae1a2/extendr-api/src/robj_ndarray.rs#L59-L61

There may be others. These can be difficult to find. It would be good to have some sort of automated test for these, so they don't sneak in to the code base unnoticed.

yutannihilation:
Fix #144

The main purpose of extendr-engine was not to call non-API symbols in R (#46). But, in actual, extendr-api unconditionally depends on it to execute tests. So, we need to opt-out the tests and accordingly the dependency on extendr-engine.

I'm yet to figure out how to do this elegantly. Currently, I made `extendr-engine` as the default feature because otherwise I probably need to add `no_run` to all the example code... Any ideas?

### DIFF
--- a/extendr-api/src/functions.rs
+++ b/extendr-api/src/functions.rs
@@ -376,18 +376,3 @@ pub fn blank_scalar_string() -> Robj {
 pub fn na_str() -> &'static str {
     unsafe { std::str::from_utf8_unchecked(&[b'N', b'A']) }
 }
-
-/// Extendr test harness.
-///
-/// See also [test!]
-/// ```
-/// extendr_api::test(|| {
-///   Ok(())
-/// })
-/// ```
-#[cfg(test)]
-pub fn test<F: FnOnce() -> Result<()>>(f: F) {
-    extendr_engine::start_r();
-
-    f().unwrap();
-}
--- a/extendr-api/src/functions.rs
+++ b/extendr-api/src/functions.rs
@@ -385,14 +385,9 @@ pub fn na_str() -> &'static str {
 ///   Ok(())
 /// })
 /// ```
-#[cfg(feature = "extendr_engine")]
+#[cfg(test)]
 pub fn test<F: FnOnce() -> Result<()>>(f: F) {
     extendr_engine::start_r();
 
     f().unwrap();
-}
-
-#[cfg(not(feature = "extendr_engine"))]
-pub fn test<F: FnOnce() -> Result<()>>(_: F) {
-    ()
 }
--- a/extendr-api/src/functions.rs
+++ b/extendr-api/src/functions.rs
@@ -385,8 +385,14 @@ pub fn na_str() -> &'static str {
 ///   Ok(())
 /// })
 /// ```
+#[cfg(feature = "extendr_engine")]
 pub fn test<F: FnOnce() -> Result<()>>(f: F) {
     extendr_engine::start_r();
 
     f().unwrap();
+}
+
+#[cfg(not(feature = "extendr_engine"))]
+pub fn test<F: FnOnce() -> Result<()>>(_: F) {
+    ()
 }
--- a/extendr-api/Cargo.toml
+++ b/extendr-api/Cargo.toml
@@ -20,7 +20,7 @@ extendr-engine = { path = "../extendr-engine", version = "0.1.11", optional = tr
 ndarray = { version = "0.13.1", optional = true }
 
 [features]
-default = []
+default = ["extendr-engine"]
 
 # All features to test
 tests-all = ["ndarray", "extendr-engine", "libR-sys/use-bindgen"]
--- a/extendr-api/Cargo.toml
+++ b/extendr-api/Cargo.toml
@@ -16,14 +16,16 @@ repository = "https://github.com/extendr/extendr"
 [dependencies]
 libR-sys = "0.2.0"
 extendr-macros = { path = "../extendr-macros", version = "0.1.11" }
-extendr-engine = { path = "../extendr-engine", version = "0.1.11", optional = true }
 ndarray = { version = "0.13.1", optional = true }
 
+[dev-dependencies]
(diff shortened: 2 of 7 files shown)


clauswilke:
I think you're way overthinking this. All tests need to be marked with `#[cfg(test)]`. This was the case at some point in the past and things worked fine. But some recent commits have added tests without `#[cfg(test)]`.

The one other change I would recommend is to put `extendr-engine` into `dev-dependencies` so downstream crates don't have to depend on it if they don't care for it.

clauswilke:
And, all tests should live in tests modules: https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-module-and-cfgtest
I believe we don't systematically do this currently.

yutannihilation:
Thanks, using `dev-dependencies` works. For test structure, I have no idea what it should be, at the moment.

clauswilke:
All tests should be set up like the ones here:
https://github.com/extendr/extendr/blob/daee5c88842b0ec226c32ae9472244b6c6a8a63e/extendr-api/src/lib.rs#L345

extendr-api/src/functions.rs
```diff
@@ -385,8 +385,14 @@ pub fn na_str() -> &'static str {
 ///   Ok(())
 /// })
 /// ```
+#[cfg(feature = "extendr_engine")]
```
clauswilke, line 388: Oh, I see, this code is the problem. The annotation here should be `#[cfg(test)]`, I think.

extendr-api/src/functions.rs
```diff
@@ -385,8 +385,14 @@ pub fn na_str() -> &'static str {
...
 pub fn test<F: FnOnce() -> Result<()>>(f: F) {
     extendr_engine::start_r();
 
     f().unwrap();
 }
+
+#[cfg(not(feature = "extendr_engine"))]
+pub fn test<F: FnOnce() -> Result<()>>(_: F) {
+    ()
+}
```
clauswilke, line 398: Can you leave this out if all tests are properly enclosed into `#[cfg(test)]`? This would then be another check against accidentally including tests outside a tests module.

yutannihilation:
Oh, I didn't remember with `#[cfg(test)]`, thanks! But it triggers another error, which probably should be fixed.

```
---- src/functions.rs - functions::base_namespace (line 162) stdout ----
error[E0423]: expected function, found macro `test`
 --> src/functions.rs:164:1
  |
5 | / test! {
6 | |    assert_eq!(base_namespace().parent().ok_or("no parent")?, global_env());
7 | | }
  | |_^ not a function
  |
  = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
help: use `!` to invoke the macro
  |
28|         test!(|| {
  |             ^
```

clauswilke:
Yes, I think this is the issue: The introduction of the `test!` macro and associated `test()` function added testing code to all api builds. If we set things up so `test!()` can only be compiled inside of test code we have also added an additional check against testing code outside of test modules.

clauswilke:
Hm, there are various failures though. Maybe a function and a macro shouldn't have the same name?

clauswilke:
Maybe @andy-thomason should take a look, since he added the `test!` macro.

The latest failure is in a documentation example. Not sure if that matters.

clauswilke:
I think the answer is that the `test()` function needs to live in extendr-engine, not in extendr-api.

yutannihilation:
> Maybe a function and a macro shouldn't have the same name?

(I think you already found the answer, but let me note here.) Since function `test` is compiled conditionally, it's unavailable when not in a testing context; we cannot use test-only functions in a macro, which is unconditionally compiled.

> I think the answer is that the `test()` function needs to live in extendr-engine, not in extendr-api.

Sounds good, but, from a user's perspective, might it be convenient to let `test()` live in extendr-api? I can't decide which one is better as an API.

clauswilke:
I think `test!` can remain in extendr-api, since it's a macro and only needs to compile when it's used. `test()` needs to be in extendr-engine since it requires the engine to be present. The macro can explicitly call `extendr-engine::test()`.

If a user needs to call the `test()` function directly, rather than use it via the `test!` macro, they'll have to also use `extendr-engine::test()`. But I don't think `test()` is meant to be used directly. It is meant to be accessed via the macro.

yutannihilation:
Makes sense, thanks.

yutannihilation:
It seems this needs some more thoughts. 

* `test()` returns `Result<()>`, which is defined in extendr-api
* As the document actually uses `test!`, it's always compiled (and currently fails). This needs to be moved to optional dependency again, instead of dev-dependency?

yutannihilation:
After looking at the commit that introduced `test!` macro (https://github.com/extendr/extendr/pull/107/commits/22f6388a4afb08aae5f844f27c01d90d1dc7500b), I'm now inclined to move extendr-engine back to optional dependency because, while `exendr_engine::test()` isn't meant to be used directly by users, `exendr_engine::start_r()` is always needed when used solely by Rust, and I guess testing and documentation are the case. For testing, we can use `[dev-dependencies]`, but there's no convenient way to use "documentation-only dependency," if I understand correctly.

andy-thomason:
I'll convert all the tests to test! next. I also want to introduce a
prelude module and phase out the global pub uses in lib.rs

yutannihilation:
Thanks. So, maybe I should wait for it? I think I don't understand what things should be placed enough to complete this pull request. Let's keep discussing on #144

PR #147, opened 2021-01-24
Clean up generated wrappers

### LINKED ISSUE #145: Wrapper info generated from Rust needs to be more configurable
The metadata wrapper code currently makes a few incorrect assumptions and needs to be fixed:

1. Roxygen tags (`#' @param`, `#' @export`, etc) should not be added automatically. They need to be under the control of the user, because not every function needs to be exported. (In fact, many Rust functions will likely not be exported from a package.) See how this is done in Rcpp: https://gallery.rcpp.org/articles/documenting-rcpp-packages/
2. This point is irrelevant if we properly address point 1, but at present, `#' @export` is written before the documentation, and that is invalid roxygen output. It'll break if you try to document as is.
3. The wrapper code assumes that we're developing a package (by adding things such as `#' @useDynLib packagename`), but that is not the only use case. Wrappers for rextendr use should not contain these lines. So this needs to be configurable.

The wrapper code is written here:
https://github.com/extendr/extendr/blob/43c5c7eafd48577f824c7d61ae353527e9116cb3/extendr-api/src/metadata.rs#L110-L111

and here:
https://github.com/extendr/extendr/blob/43c5c7eafd48577f824c7d61ae353527e9116cb3/extendr-api/src/metadata.rs#L193-L224

clauswilke:
Attempt to fix #145.


tests/extendrtests/tests/testthat/test-wrappers.R
```diff
@@ -1,3 +1,31 @@
+test_that("Wrapper code is up-to-date", {
+  # if this test fails check whether the wrapper code needs 
+  # updating via make_wrappers()
```
Ilia-Kosenkov, line 3: I think there should be a 'comma' after 'fails', I tried reading this sentence several times before understanding what was meant.

tests/extendrtests/tests/testthat/test-wrappers.R
```diff
@@ -1,3 +1,31 @@
...
+    tmp <- file.path("..", "..", "00_pkg_src", "extendrtests", "R", "extendr-wrappers.R")
+    if (file.exists(tmp)) {
+      source <- tmp
+    } else {
+      skip("Cannot locate wrapper code.")
+    }
+  }
+  
+  y <- brio::read_lines(source)
+  expect_equal(x, y)
```
Ilia-Kosenkov, line 26: Perhaps add a few sentences atop of `test_that` to explain what is achieved here (again, it was not immediately evident to me what is tested), as the test body looks complicated.
Something like `"check the existence of 'extendr-wrappers.R' and compare it to the generated up-to-date version. If this test fails...."`. Use better wording, I am not good at this :)

Ilia-Kosenkov:
Also, I saw a typo in `extendrtests` description
https://github.com/extendr/extendr/blob/daee5c88842b0ec226c32ae9472244b6c6a8a63e/tests/extendrtests/DESCRIPTION#L10
Perhaps you could also fix it here?

tests/extendrtests/tests/testthat/test-wrappers.R
clauswilke: Completely rewritten this part.

tests/extendrtests/tests/testthat/test-wrappers.R
clauswilke: Done.

clauswilke:
Fixed `DESCRIPTION` also.

Ilia-Kosenkov (approved):
The rest looks good to me, so if the current tests pass, it should be merged.

PR #149, opened 2021-01-24
Sanitizes Rust parameter names in R

### LINKED ISSUE #125: Sanitize wrapper function parameter names in R

Ilia-Kosenkov:
This should fix #125. 
Code should be likely updated after #147.

So I am not entirely sure if I did this correctly, but I fixed argument generation in two places, in `extendr-api`
https://github.com/extendr/extendr/blob/daee5c88842b0ec226c32ae9472244b6c6a8a63e/extendr-api/src/metadata.rs#L113-L118
https://github.com/extendr/extendr/blob/daee5c88842b0ec226c32ae9472244b6c6a8a63e/extendr-api/src/metadata.rs#L148-L159

and in `extendr-macros`
https://github.com/extendr/extendr/blob/daee5c88842b0ec226c32ae9472244b6c6a8a63e/extendr-macros/src/output_r.rs#L40-L45
which should be verified. The last one seems to affect how names are generated in `rextendr`:
Before patching:
``` r
rextendr::rust_function("fn rust_add(_x : i32, _y : i32) -> i32 {_x + _y}")
#> build directory: C:\Users\[redacted]\AppData\Local\Temp\RtmpQNqybE\file456c82e7747
#> Error in source(r_path, local = env): C:\Users\[redacted]\AppData\Local\Temp\RtmpQNqybE\file456c82e7747/target/extendr_wrappers.R:7:22: unexpected input
#> 6:         
#> 7: rust_add <- function(_
#>                         ^
rust_add(`_y` = 42, `_x` = 100)
#> Error in rust_add(`_y` = 42, `_x` = 100): could not find function "rust_add"
```

<sup>Created on 2021-01-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

After patching:
``` r
rextendr::rust_function(
    "fn rust_add(_x : i32, _y : i32) -> i32 {_x + _y}",
    patch.crates_io =  c(
        'extendr-api = { git = "https://github.com/Ilia-Kosenkov/extendr", branch = "sanitize-parameter-names" }',
        'extendr-macros = { git = "https://github.com/Ilia-Kosenkov/extendr", branch = "sanitize-parameter-names" }'
    ))
#> build directory: C:\Users\[redacted]\AppData\Local\Temp\RtmpSQkCQO\file31409115634
rust_add(`_y` = 42, `_x` = 100)
#> [1] 142
```

<sup>Created on 2021-01-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Please verify if I did this correctly and that this PR covers all places where R wrapper functions are generated.


clauswilke (approved):
This looks good to me. I think the long-term goal is to get rid of `extendr-macros/src/output_r.rs`, as it could break at some point in the future with changes in rustc, but for now fixing it is the right way to go.

Ilia-Kosenkov:
@clauswilke,
I am unfortunately still trying to get a full picture of `extendr`, so my changes are mostly cosmetic.
Do you think I should wait for other PR to resolve conflicts first?

clauswilke:
I think you can merge this now. Eventually we will want to have a test of this in extendrtests, but for that I'd wait until #147 is merged.

If you'd rather wait until #147 is merged, and then add a test, and then merge that's also fine.

Ilia-Kosenkov:
@clauswilke,
That was actually my plan to wait for #147. 
`extendrtests` needs some rework because right now I see no way of testing wrappers because it does not use any emitted wrappers at all.
I am leaving this open for discussion and possible improvements.

clauswilke:
Could you review #147? Do you see any issues that still need addressing?

Ilia-Kosenkov:
Awaiting merge of #153 to fix `()` - returning Rust functions not being wrapped in `invisible(...)`.

Ilia-Kosenkov:
There is a strange documentation error, see e.g. here:
https://github.com/extendr/extendr/pull/149/checks?check_run_id=1760988205#step:14:135

My tests function has two documented arguments:
https://github.com/Ilia-Kosenkov/extendr/blob/2560830d34c2c4cbf96a94e1b239f65ef28160f1/tests/extendrtests/R/extendr-wrappers.R#L40-L44

but for some reason, the second argument is not documented correctly (even though it is present in the doc file).
https://github.com/Ilia-Kosenkov/extendr/blob/sanitize-parameter-names/tests/extendrtests/man/special_param_names.Rd

This should be resolved because we test argument names and their handling by wrappers, so it cannot be renamed. If left untouched, tests will fail.
Any help resolving this problem is welcomed!

andy-thomason (approved):
Looks good to me.

There may be other cases than a leading _ where the identifiers diverge.

The quoting mechanism is much better than my type name default. We can use this
in the type name sanitiser.

Ilia-Kosenkov:
@andy-thomason,
I used this for reference https://doc.rust-lang.org/reference/identifiers.html.
To summarize, valid identifiers are `[A-z][A-z0-9_]*|_[A-z0-9_]+`, and the only problematic type is the underscore prefix, which I fix. 
This works now:
``` r
rextendr::rust_function(
    "fn rust_test(__ : i32) {}",
    patch.crates_io =  c(
        'extendr-api = { git = "https://github.com/Ilia-Kosenkov/extendr", branch = "sanitize-parameter-names" }',
        'extendr-macros = { git = "https://github.com/Ilia-Kosenkov/extendr", branch = "sanitize-parameter-names" }'
    ))
#> build directory: C:\Users\[redacted]\AppData\Local\Temp\RtmpMj6kGo\file36ac761028a2
print(rust_test)
#> function(`__`) {
#>     invisible(.Call("wrap__rust_test", `__`, PACKAGE = "rextendr1"))
#> }
rust_test(`__` = 100L)
```

<sup>Created on 2021-01-25 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>
But this obviously does not:
``` r
rextendr::rust_function(
    "fn __rust_test(__ : i32) {}",
    patch.crates_io =  c(
        'extendr-api = { git = "https://github.com/Ilia-Kosenkov/extendr", branch = "sanitize-parameter-names" }',
        'extendr-macros = { git = "https://github.com/Ilia-Kosenkov/extendr", branch = "sanitize-parameter-names" }'
    ))
#> build directory: C:\Users\[redacted]\AppData\Local\Temp\Rtmp8GovX4\file2adc1113276c
#> Error in source(r_path, local = env): C:\Users\iliak\AppData\Local\Temp\Rtmp8GovX4\file2adc1113276c/target/extendr_wrappers.R:7:1: unexpected input
#> 6:         
#> 7: _
#>    ^
print(`__rust_test`)
#> Error in print(`__rust_test`): object '__rust_test' not found
`__rust_test`(`__` = 100L)
#> Error in `__rust_test`(`__` = 100L): could not find function "__rust_test"
```

<sup>Created on 2021-01-25 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I can try to update my PR to also fix at least function and method names.

andy-thomason:
We will need a more general function that takes a string and santises it,
including my type names.

This would apply to anything that generates an R identifier.

This is the definition for Rust:

https://doc.rust-lang.org/reference/identifiers.html

We can have r#match, for example in rust.

Andy.

clauswilke:
You need to put the parameters into backticks after `@param` in the rust code here:
https://github.com/extendr/extendr/blob/2560830d34c2c4cbf96a94e1b239f65ef28160f1/tests/extendrtests/src/rust/src/lib.rs#L42-L47

Like so:
```rust
 /// Test whether `_arg` parameters are treated correctly in R 
 /// @param `_x` an integer scalar, ignored 
 /// @param `_y` an integer scalar, ignored 
 /// @export 
 #[extendr] 
 fn special_param_names(_x : i32, _y : i32) {} 
```

Ilia-Kosenkov:
@clauswilke,
Did not work, it then complains that I document ``` `_x` ``` and ``` `_y` ``` while actual params are `_x` and `_y`.
Notice that it complains only about the second argument, which is very strange.

clauswilke:
It’s possible that this is a bug in roxygen2. Need to investigate more.

clauswilke:
If we can’t resolve this within the next day or so you can just make your test unexported and delete the `@param` statements.

clauswilke:
To get at the question whether it's a roxygen2 bug, you could add the backticks manually in the `.Rd` file here:
https://github.com/extendr/extendr/blob/2560830d34c2c4cbf96a94e1b239f65ef28160f1/tests/extendrtests/man/special_param_names.Rd#L10-L12
and see if then the package builds. (Don't rerun `devtools::document()`.)

You could also add the backticks back in the wrapper code and see what roxygen2 makes of this in the `.Rd` file. I suspect it interprets the parameters in backticks as markdown instead of handing them through as is.

Ilia-Kosenkov:
@clauswilke,
I think roxygen writes docs correctly (check the generated document)
https://github.com/Ilia-Kosenkov/extendr/blob/sanitize-parameter-names/tests/extendrtests/man/special_param_names.Rd

Perhaps I can solve this by using only one underscored argument or mixing regular/underscored ones. In the end, one argument is enough to test everything.

clauswilke:
I believe it should be:
```
 \item{`_x`}{an integer scalar, ignored} 
  
 \item{`_y`}{an integer scalar, ignored} 
```
instead of the same without backticks.

It's also possible though that R has a problem with parameter names in backticks and nobody has ever tested this carefully. It's not like people normally do this.

Ilia-Kosenkov:
@clauswilke,
Sorry I missed one of your comments. I did exactly that -- added backticks to documentation file only and got an error that I document ``` `_x` ``` while the argument is just `_x`.

clauswilke:
Then it's probably a problem with R itself. You could add a comment to the docstring that documenting and exporting functions with non-standard parameter names may not work as expected.

Ilia-Kosenkov:
@clauswilke,
The documentation issue was resolved by *not wrapping* first parameter with backticks and *wrapping* the second one, which is weird yet works.

I have another set of features that fixes class names -- as @andy-thomason suggested earlier.
These features are ready to be pushed here or can be added as a separate PR.

clauswilke:
I still would add a comment reminding us that documenting non-standard parameters seems to have issues. For example, even if your workaround passes checks now, the generated `.Rd` page will probably not render correctly. (Both parameters should be shown as enclosed in backticks, I think.)

Regarding the other features, it's up to you if you want to merge this now or keep adding to this PR.

Ilia-Kosenkov:
@clauswilke,
Ok, I will add a note. I will attempt to add more features to sort of finalize this name-fixing update.

Ilia-Kosenkov:
So this is it. With my updates, we can correctly wrap Rust code like this
https://github.com/Ilia-Kosenkov/extendr/blob/1910d0b1bfa4c0a9cc0a1df4b03c27db28cd7bf7/tests/extendrtests/src/rust/src/lib.rs#L49-L63

And later test it in R
https://github.com/Ilia-Kosenkov/extendr/blob/1910d0b1bfa4c0a9cc0a1df4b03c27db28cd7bf7/tests/extendrtests/tests/testthat/test-wrappers.R#L38-L49

Reviews & comments are welcomed.

tests/extendrtests/src/rust/src/lib.rs
```diff
@@ -46,6 +46,22 @@ fn char_scalar(x: String) -> String { x }
...
 fn char_vec(x: Vec<String>) -> Vec<String> {x}
 
+// Weird behavior of parameter descriptions:
+// first passes tests as is, second -- only in backqutoes.
+/// Test whether `_arg` parameters are treated correctly in R
+/// @param _x an integer scalar, ignored
+/// @param `_y` an integer scalar, ignored
+/// @export
+#[extendr]
+fn special_param_names(_x : i32, _y : i32) {}
```
clauswilke, line 56: I suggest you do some non-commutative computation on `_x` and `_y` (e.g., `_x - _y`) and return, so you can test that the parameters are actually applied correctly in your R code.

tests/extendrtests/tests/testthat/test-wrappers.R
```diff
@@ -35,6 +35,19 @@ test_that("Wrapper code is up-to-date", {
...
+  expect_invisible(`__00__special_function_name`())
+})
+
+test_that("Rust classes and methods prefixed with `_` can be invoked", {
+  x <- `__MyClass`$new()
+  expect_invisible(x$`__name_test`())
+})
+
+test_that("Rust `_arg` is correctly wrapped in R", {
+  expect_invisible(special_param_names(`_y` = 42L, `_x` = 100L))
```
clauswilke, line 48: Since you flipped `_x` and `_y` here, it would be good to test that they are transferred correctly to Rust. See my other comment.

Ilia-Kosenkov:
Let me know if there are any other issues with this PR.
cc: @clauswilke, @andy-thomason
If not, I would like to merge it to start working on properties & accessors.

extendr-api/src/metadata.rs
```diff
@@ -96,6 +96,14 @@ fn write_doc(w: &mut Vec<u8>, doc: &str) -> std::io::Result<()> {
     Ok(())
 }
 
+/// Wraps invalid R identifers, like `_function_name`, into backticks.
+fn sanitize_identifier(name: &str) -> String {
+    match name.chars().next() {
+        Some('_') => format!("`{}`", name),
+        _ => name.to_string(),
+    }
```
yutannihilation, line 104: Sorry for nitpicking comment..., using `starts_with()` looks easier to read to me (but this might be just my preference).

``` rust
    if name.starts_with('_') {
        format!("`{}`", name)
    } else {
        name.to_string()
    }
```

Ilia-Kosenkov:
@yutannihilation,
Done! I was looking for a method of this signature, but for some reason missed it, thank you for noticing.

yutannihilation (approved):
Thanks, looks good!

clauswilke:
Yes, go ahead and merge. Thanks!

PR #150, opened 2021-01-24
Remove extendr-engine call from non-testing code.

### LINKED ISSUE #144: Some tests are not conditionally compiled

clauswilke:
Fixes #144.


extendr-api/src/robj_ndarray.rs
```diff
@@ -56,6 +56,7 @@ make_array_view_2!(f64, "Not a floating point matrix.");
 //     }
 // }
 
+#[cfg(test)]
```
yutannihilation, line 59: I don't think this line is needed.

yutannihilation (approved):
Wow, so simple! Thanks for taking this.

extendr-api/src/robj_ndarray.rs
clauswilke: Wouldn't an R package that turns on the ndarray feature run into the same problem? We don't currently do that in extendrtests, but we probably should.

yutannihilation: I meant, as there's a bare function here and it's marked `#[test]`, the compilation happens only on testing. When we start using test module, we'll need `#[cfg(test)]`.

Let's confirm it here: https://github.com/yutannihilation/extendr/commit/189edba82acdd4906704112981523ef45dc7ec6e

clauswilke: Ah, Ok, yes that seems to be correct.

PR #156, opened 2021-01-27
Add prelude

andy-thomason:
This PR adds a prelude to extendr_api.

From now on, use:

```
use extendr_api::prelude::*;
```
and not
```
use extendr_api::*;
```
The latter will be turned off in the next few weeks.

The tests have been updated to use the test! macro and the prelude.


clauswilke (approved):

clauswilke:
There are still some problems with the integration tests, but otherwise looks good to me.

Ilia-Kosenkov:
There is some issue with R-devel for Windows, I saw this earlier in my fork as well:
https://github.com/extendr/extendr/pull/156/checks?check_run_id=1780350499#step:4:15
Perhaps something happened during an update to R actions, I suggest to temporarily roll back to release.

yutannihilation:
It seems it's not only us seeing this failure: https://github.com/r-lib/actions/issues/234

extendr-api/src/functions.rs
```diff
@@ -4,7 +4,7 @@ use crate::*;
 ///
 /// Use the Env{} wrapper for more detail.
 /// ```
-/// use extendr_api::*;
+/// use extendr_api::prelude::*;;
```
yutannihilation, line 7: This file contains some double `;`. Maybe you mistyped a regex replacement?

andy-thomason: Very likely!

yutannihilation (approved):

PR #157, opened 2021-01-28
Correct windows testing

Ilia-Kosenkov:
I suggest adding an additional set of bindgen tests (or replace the existing one) to allow correct testing of  Windows 32-bit builds.
This script is not trivial, as we need more than one Rust toolchain.
It also has linking issues with `stable-i686-gnu` toolchain (see https://github.com/rust-lang/cargo/issues/8990), but there is a bypass.
Again, the purpose of this config is to show that there are issues in Win-32 build which our tests previously did not reveal (but now do).

NB: I left MacOS and Ubuntu configurations enabled to verify it also works for other systems as is. I can disable/remove unused targets to preserve computational resources.


Ilia-Kosenkov:
@andy-thomason, here is the Win-32 issue I mentioned earlier.
https://github.com/extendr/extendr/pull/157/checks?check_run_id=1784355866#step:8:580

Ilia-Kosenkov:
The reason is no support for cross-compilation of doc tests, 
https://github.com/rust-lang/rust/issues/44404

.github/workflows/test.yml
```diff
@@ -331,3 +331,234 @@ jobs:
             
             echo "::endgroup::"
           }
+
+  
+  
+  correct_bindgen:
```
clauswilke, line 337: Maybe pick a name that is a bit more informative? If the main point is no cross compile call it `bindgen_no_cross_compile` maybe?

Also, could you add a few comments explaining what this test is meant to do compared to the other tests?

Ilia-Kosenkov: @clauswilke,
I added some comments. Currently, we have 'old' bindgen tests and new 'no cross-compilation' bindgen tests, which do pretty much the same thing. Should we remove 'old' bindgen job because it does not work for Windows (at all)?

clauswilke: > Should we remove 'old' bindgen job because it does not work for Windows (at all)?

Yes, I think that's fine.

Ilia-Kosenkov: Ok, removed old tests.

clauswilke (approved):

Issue #140, opened 2021-01-19
Consider making the init functions public


dfalbel:
Can we consider making `init__` functions public? I mean here:

https://github.com/extendr/extendr/blob/10d0ab8a64ab4fafb125d69cf555f08b5b090593/extendr-macros/src/lib.rs#L274

and here:

https://github.com/extendr/extendr/blob/10d0ab8a64ab4fafb125d69cf555f08b5b090593/extendr-macros/src/lib.rs#L412

The use case is to be able split the implementation in multiple files. With this one could have a file named `utils.rs` with:

```
use extendr_api::*;

#[extendr]
pub fn hello () -> String {
    "hello world".to_string()
}
```

and then export the hello function in `lib.rs` with:

```
use extendr_api::*;
mod utils;
use utils::*;

extendr_module! {
    mod helloextendr;
    fn hello;
}
```

If this is undesirable, Is there any approach for splitting the implementation in multiple files?

clauswilke:
Maybe use `pub(crate)` or `pub(super)`?

clauswilke:
However, PR #128 is doing away with init functions, so maybe you could review it and see whether it requires modifications to allow splitting the implementation into multiple files? I think generally it's a good idea to be able to do this.

dfalbel:
Ok, I reviewed #128 . A similar approach would work - just making the below functions public allows exporting functions implemented in other modules.

https://github.com/extendr/extendr/blob/dc7c647e6c135789a556013f171ee373a8bbae33/extendr-macros/src/lib.rs#L329

and 

https://github.com/extendr/extendr/blob/dc7c647e6c135789a556013f171ee373a8bbae33/extendr-macros/src/lib.rs#L484

andy-thomason:
Whilst the meta__ functiions are not public, the get_xxx_metadata functions, which call them, are.

Thus you should be able to pub use them.

However, the `extendr_module!` macro currently assumes that the DLL name is the same as the module name
as the `R_init_xxx` function name is DLL-specific.

We could add a syntax to chain metadata:

```
extendr_module! {
  mod dll_name;
  use my_submodule;
}
```

This could add my_submodule's metadata to the main dll metadata.
Omitting the "mod" would prevent the R_init_ function from being generated.

Now the registration is done entirely from metadata, many more possibilities arise.

clauswilke:
Is this addressed now or does something remain to be done?

dfalbel:
No, #160 addressed this. Closing

[...]

Issue #151, opened 2021-01-24
Allow for multiple extendr modules in one Rust library


clauswilke:
I'd like to be able to have multiple extendr modules in one Rust library. Something like this:

```rust
#[extendr]
fn add(x: f64, y: f64) -> f64 {
    x + y
}

#[extendr]
fn sub(x: f64, y: f64) -> f64 {
    x - y
}

extendr_module! {
    mod add;
    fn add;
}

extendr_module! {
    mod sub;
    fn sub;
}
```
This is particularly useful for extendrtests, but may have applications more broadly.

This works in principle, except for one issue: If we try to call the routine registration functions for the two modules, the most recent registration always overwrites any previous registration for the same library. So, for example, this does not work:
```C
void R_init_extendrtests(void *dll) {
  R_init_add_extendr(dll);
  R_init_sub_extendr(dll);
}
```
Only the `sub` module gets registered.

However, we could implement something like the following which would work:
```C
const char* mods[] = { "add", "sub" };

void R_init_extendrtests(void *dll) {
  R_init_extendr(dll, mods, 2);
}
```
I.e., we provide a generic init function that registers all modules it is given.

This feature could be implemented in addition to the current initialization scheme, so people who have only a single module can call that module's function directly but people who have multiple modules can register all of them at once.

clauswilke:
The api I outlined here may not work, as it may not be that easy (or even possible) to call the metadata functions for the different modules by module name provided as a string. However, an alternative approach will definitely be possible: Call Rust functions from C to retrieve pointers to the Rust data structures with the metadata, and then call a Rust function to register all functions in that metadata.

clauswilke:
This seems to be addressed with #160.

PR #160, opened 2021-01-30
Allow submodules

dfalbel:
This PR is an attempt to allow submodules as suggested by @andy-thomason here https://github.com/extendr/extendr/issues/140#issuecomment-764071735

The missing part would be to allow the submodule to be exported with something like this, ie. without the `mod` declaration.

```
extendr_module! {
  fn my_function;
  impl my_impl;
}
```
I couldn't find a way to do this because the `modulename` is used to generate the `get_{modname}_metadata` function that is then used to fuse the main module metadata with submodules.

Thus the `init__*` functions are still generated, but not used, because the metadata is included in the main module when using:

```
extendr_module! {
  mod dll_name;
  use my_submodule;
}
```

I'd appreciate some guidance on how to solve this issue.


clauswilke:
See also #151, which may be related.

Would it be possible to give submodules their own module name and then somehow add them to the main module with something like `use my_submodule;`? 

In any case, all functions to be registered with R need to be collected and registered at once, we cannot separately call the current registration functions for submodules because they will overwrite previous registrations.

dfalbel:
Yeah, this is what this PR is doing. In `submodule.rs` we need to declare:

```
extendr_module! {
    mod submodule;
    fn hello_submodule;
    impl MySubmoduleClass;
}
```

And in `lib.rs` we do:

```
use extendr_api::prelude::*;
mod submodule;
use submodule::*;

extendr_module! {
    mod extendrtests;
    use submodule;
}
```

With this the `R_init_submodule_extendr` is still created, but it's never used because all registration happens in `R_init_extendrtests_extendr`.

clauswilke:
I guess I'm not sure I fully understand the issue then. Is it just aesthetics, that there is a second init routine that is generated but not used, or is there a problem with functionality (functions that should be registered but aren't)?

I think all modules and submodules will always need a name. So this will never work I think:
```
extendr_module! {
  fn my_function;
  impl my_impl;
}
```

dfalbel:
AFAICT it's just aesthetic. All functionality seems to work fine, it's just that we register something that we don't use...

In theory, it would be possible to have 'unnamed' modules and generate `submodule::get_metadata` instead of `submodule::get_submodule_metadata`, ie. let the namespace prefix to disambiguate, but it might be better to keep the names anyway.

clauswilke:
Yeah, looks good to me. I'm not so worried about extra registration functions. Who knows, maybe somebody can do some nifty things with it. (You could make one Rust library behave in two different ways in R, depending on which registration function you call. Not sure if that's in any way useful, but maybe it is.)

Are the wrapper generator functions getting registered, i.e., `wrap__make_<submodule>_wrappers()`? Can I generate wrappers just for a submodule? (This would be great for testing, where we could have a separate small submodule where we test that the wrappers come out just as expected, and that would never change even as we add more integration tests otherwise.)

extendr-macros/src/lib.rs
```diff
@@ -710,6 +719,9 @@ pub fn extendr_module(item: TokenStream) -> TokenStream {
                 hidden: true,
             });
 
+            #( functions.extend(#usemetanames().functions); )*
+            #( impls.extend(#usemetanames().impls); )*
```
clauswilke, line 723: Maybe move these two lines up to where the module functions and implementations are added? (I.e., above the metadata access function.)

clauswilke (approved):
While you're at it, could you add a comment to these two lines:

```
#( #fnmetanames(&mut functions); )*
#( #implmetanames(&mut impls); )*
```
Something like `// add function metadata` or similar? It took me a while to figure out what these two functions do, so adding a comment may make this code more maintainable in the future.

dfalbel:
Thanks @clauswilke  for the review!

> Yeah, looks good to me. I'm not so worried about extra registration functions. Who knows, maybe somebody can do some nifty things with it. (You could make one Rust library behave in two different ways in R, depending on which registration function you call. Not sure if that's in any way useful, but maybe it is.)

Makes sense to me! I think having the same code path is more valuable here and simplifies the implementation.

> Are the wrapper generator functions getting registered, i.e., wrap__make_<submodule>_wrappers()? Can I generate wrappers just for a submodule? (This would be great for testing, where we could have a separate small submodule where we test that the wrappers come out just as expected, and that would never change even as we add more integration tests otherwise.)

Yeah, it seems that `wrap__make_<submodule>_wrapers` is registered:

``` r
library(extendrtests)
.Call(extendrtests:::wrap__make_submodule_wrappers, TRUE, "extendrtests")
#> [1] "# Generated by extendr: Do not edit by hand\n#\n# This file was created with the following call:\n#   .Call(\"wrap__make_submodule_wrappers\", use_symbols = TRUE, package_name = \"extendrtests\")\n\n#' @docType package\n#' @usage NULL\n#' @useDynLib extendrtests, .registration = TRUE\nNULL\n\n#' Return string `\"Hello world!\"` to R.\n#' @export\nhello_submodule <- function() .Call(wrap__hello_submodule)\n\n#' Class for testing (exported)\n#' @examples\n#' x <- MySubmoduleClass$new()\n#' x$a()\n#' x$set_a(10)\n#' x$a()\n#' @export\nMySubmoduleClass <- new.env(parent = emptyenv())\n\nMySubmoduleClass$new <- function() .Call(wrap__MySubmoduleClass__new)\n\nMySubmoduleClass$set_a <- function(x) invisible(.Call(wrap__MySubmoduleClass__set_a, self, x))\n\nMySubmoduleClass$a <- function() .Call(wrap__MySubmoduleClass__a, self)\n\nMySubmoduleClass$me <- function() .Call(wrap__MySubmoduleClass__me, self)\n\n#' @rdname MySubmoduleClass\n#' @usage NULL\n#' @export\n`$.MySubmoduleClass` <- function (self, name) { func <- MySubmoduleClass[[name]]; environment(func) <- environment(); func }\n\n"
```

> Maybe move these two lines up to where the module functions and implementations are added? (I.e., above the metadata access function.)

Makes sense! I'll move the code up and add comments! 

Thanks again!

clauswilke:
There's a small code formatting problem now. Do you think it's ready to be merged once that is fixed?

dfalbel:
Fixed the code formatting issue! I think this is ready to merge.

clauswilke:
Thanks!

Issue #138, opened 2021-01-19
Allow R `NULL` with Option


dfalbel:
Currently when passing a NULL to an `Option<T>`, for example: https://github.com/extendr/extendr/blob/master/extendr-api/src/robj/from_robj.rs#L172-L182

I get an error:

```
expected an integer scalar
```

Should we allow NULL's here and return `None` in this case as we do with `NA's`? IMHO, it's more common to use `NULL` than `NA` when you want your argument to be optional.

More generally, I wonder if it would make sense to have an `ROption` enum, that would distinct NULL's from NA's, eg:

```
enum ROption<T> {
    None,
    Some(T),
    Na
}
```

clauswilke:
In my opinion `NULL` and `NA` live in different worlds and thus don't belong into one option. `NA` indicates absence of a value in a vector. `NULL` indicates absence of the entire object.

Not sure how to best map this onto Rust data types, though. For scalar inputs, mapping `NULL` onto `None` seems fine. For vector inputs, I'm not sure. `NULL` would correspond to a vector of length 0, whereas `NA` would correspond to a vector of length 1 containing no value?

andy-thomason:
We could have a Nullable type that can wrap anything. We could then do a
blanket wrap.

For optional parameters, I'm looking at a way of putting default values
into r wrappers. You could choose NA or null. There is also a special
missing arg value we could use.

dfalbel:
@clauswilke ! Agree that they are different concepts and shouldn't live in the same place.

Considering `Option` for `NULL` we could have something similar to what we have for logical values [here](https://github.com/extendr/extendr/blob/master/extendr-api/src/logical.rs) for all the other types. 

Thus what we would have is something like:

```
Option<bool> # accepts `NULL` or logical values (not NA)
Option<RBool> # accepts `NULL` or logical values including NA_LOGICAL

Vec<bool> # vector of logical (don't accept NA's)
Vec<RBool> # vector of logical (accepting NA's)
Option<Vec<bool>> accepts NULL or a vector of with no NA's
Option<Vec<RBool>> aceepts NULL or a vector of logical (possibly including NA_LOGICAL)
Option<Vec<Option<RBool>>> # errors, or probably not implemented in extendr-api.
```

Would that make sense?

clauswilke:
[...]

PR #163, opened 2021-01-31
Add a Nullable wrapper type to handle NULLs.

andy-thomason:
Solution to https://github.com/extendr/extendr/issues/138

Wrapper for handling potentially NULL values.

```
    let s1 = r!(1);
    let n1 = <Nullable<i32>>::from_robj(&s1)?;
    assert_eq!(n1, Nullable::NotNull(1));
    let snull = r!(NULL);
    let nnull = <Nullable<i32>>::from_robj(&snull)?;
    assert_eq!(nnull, Nullable::Null);

    assert_eq!(r!(Nullable::<i32>::Null), r!(NULL));
    assert_eq!(r!(Nullable::<i32>::NotNull(1)), r!(1));
```

### DIFF
--- a/extendr-api/src/wrapper.rs
+++ b/extendr-api/src/wrapper.rs
@@ -11,9 +11,9 @@ use libR_sys::*;
 /// ```
 /// use extendr_api::prelude::*;
 /// test! {
-/// let symbol = r!(Symbol("xyz"));
-/// assert_eq!(symbol.as_symbol(), Some(Symbol("xyz")));
-/// assert!(symbol.is_symbol());
+///     let symbol = r!(Symbol("xyz"));
+///     assert_eq!(symbol.as_symbol(), Some(Symbol("xyz")));
+///     assert!(symbol.is_symbol());
 /// }
 /// ```
 /// Note that creating a symbol from a string is expensive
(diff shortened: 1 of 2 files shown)


clauswilke:
To be clear how this interacts with `NA`, could you add an example where a nullable value is set to `NA`?

Would it be something like this?

```rust
let s1 = r!(NA);
let n1 = <Nullable<Option<i32>>>::from_robj(&s1)?;
assert_eq!(n1, Nullable::NotNull(None));
```

extendr-api/src/wrapper.rs
```diff
@@ -542,10 +563,11 @@ impl Robj {
...
     // /// use extendr_api::prelude::*;
     // /// test! {
-    // /// let builtin = r!(Primitive("+"));
-    // /// let special = r!(Primitive("if"));
-    // /// assert_eq!(builtin.sexptype(), libR_sys::BUILTINSXP);
-    // /// assert_eq!(special.sexptype(), libR_sys::SPECIALSXP);
+    // ///  let builtin = r!(Primitive("+"));
+    // ///  let special = r!(Primitive("if"));
+    // ///  assert_eq!(builtin.sexptype(), libR_sys::BUILTINSXP);
+    // ///  assert_eq!(special.sexptype(), libR_sys::SPECIALSXP);
```
clauswilke, line 569: Should this be indented four spaces?

clauswilke (approved):

Ilia-Kosenkov (approved):
No objections from me, but I agree with @clauswilke that at least one test showing how it works (and verifying conversions) should be added.

andy-thomason:
I'll try to get an NA example in.as in r!(NA_INTEGER).

PR #168, opened 2021-02-02
Wrapping cargo calls on CI

### LINKED ISSUE #158: [CI] Define a wrapper of cargo to throw error nicely
After googling around, I think it might be an option to define something like

``` ps1
# We cannot specify `cargo` in function `cargo`, otherwise it will go to infinite recursion
$cargo_path = (Get-Command -Name cargo -CommandType Application).Path

if (!$cargo_path) {
  Throw "cargo not found"
}

# The basic idea is from https://stackoverflow.com/a/48877892
function cargo {
  # to prevent 2> redirections from triggering a terminating error.
  $ErrorActionPreference = 'Continue'

  & $cargo_path $args[-0]
  if ($LASTEXITCODE) {
    # Instead of Throw, we can use Write-Error to throw nicer error message
    # c.f. https://docs.microsoft.com/powershell/scripting/learn/deep-dives/everything-about-exceptions?view=powershell-7.1
    Write-Error -Message "cargo failed with status code $LASTEXITCODE (args: $args)" -ErrorAction Stop
  }
}
```

so that we can just write `cargo build/test` instead of always placing the `if` statement at the tail like this

``` ps1
            cargo build ...
            if (!$?) { 
              echo "::error::Build extendr-api for $target target" ;
              throw "Last exit code $LASTEXITCODE"
            }
```

But, of course this probably decrease the readability of the CI results as it's difficult to include a friendly error message like `::error::Build` or `$target` above. So, I'm not sure if this is good or bad at the moment.

Ilia-Kosenkov:
This fixes #158.
`cargo` is replaced by `ci-cargo` (from `ci-cargo.ps1`), which handles exceptional situations better.
It also provides rudimentary logging through `-ActionName` parameter, utilizing CI's `"::group::"` syntax.
`ci-cargo` itself is documented and can be used outside of CI (e.g., when testing locally).

<details>
<summary>ci-cargo docs</summary>

```

NAME
    ci-cargo
    
SYNOPSIS
    Runs cargo with specified args, adapting error handling and output to CI.
    
    
SYNTAX
    ci-cargo [[-CargoArgs] <String[]>] [-ActionName <String>] [<CommonParameters>]
    
    
DESCRIPTION
    Runs cargo in a `try` block, catches exceptions and non-zero exit codes.
    Explicitly logs the beginning and the end of cargo execution, as well as the error message.
    

PARAMETERS
    -CargoArgs <String[]>
        Arguments passed to cargo, as-is.
        Note that `--` separator is handled by powershell itself,
        so it should be wrapped in quotes `'--'` and passed as string.
        
        Required?                    false
        Position?                    1
        Default value                
        Accept pipeline input?       false
        Accept wildcard characters?  false
        
    -ActionName <String>
        Optional string that is used to format logs and error messages.
        
        Required?                    false
        Position?                    named
        Default value                
        Accept pipeline input?       false
        Accept wildcard characters?  false
        
    <CommonParameters>
        This cmdlet supports the common parameters: Verbose, Debug,
        ErrorAction, ErrorVariable, WarningAction, WarningVariable,
        OutBuffer, PipelineVariable, and OutVariable. For more information, see
        about_CommonParameters (https://go.microsoft.com/fwlink/?LinkID=113216). 
    
INPUTS
    None. You cannot pipe objects.
    
    
OUTPUTS
    No explicit output.
    
    
    -------------------------- EXAMPLE 1 --------------------------
    
    PS>ci-cargo --version
        ::group::
        Running cargo --version
        cargo 1.49.0 (d00d64df9 2020-12-05)
        ::endgroup::
    
    
    
    
    
    
    -------------------------- EXAMPLE 2 --------------------------
    
    PS>ci-cargo -ActioName "Build"  build
    
    
    
    
    
    
    -------------------------- EXAMPLE 3 --------------------------
    
    PS>ci-cargo +stable-x86_64-pc-windows-gnu test --features tests-all --target i686-pc-windows-gnu '--' --nocapture -ActionName "Called from documentation"
    
    
    
    
    
    
    
RELATED LINKS
    Used by: https://github.com/extendr/extendr 




```

</details>


Ilia-Kosenkov:
Windows bindgen job (which fails on i686 tests anyway) failed due to some transient network error. It can be rerun before the merge or if fixes are requested.

.github/workflows/test.yml
```diff
@@ -9,7 +9,14 @@ on:
     branches:
       - main
       - master
+  workflow_dispatch:
```
yutannihilation, line 12: How is this useful?

yutannihilation (approved):
Looks good! Just curious about the option I've never used.

.github/workflows/test.yml
Ilia-Kosenkov: Wow, this is an artifact of my ci debug process. As you probably know. debugging CI is a nightmare, so `workflow_dispatch` allows you to run CI 'on-click', from Actions menu. You can push any number of commits to the branch and run CI when you think it is needed, not on plain 'push'.
I found that simply adding my branch to 'push' list is enough to debug CI.

yutannihilation: Oh, thanks, this can be helpful in some cases. Glad to know the feature!

>  I found that simply adding my branch to 'push' list is enough to debug CI.
Yeah, this is the way I always do.

Ilia-Kosenkov:
So if there are no objections, I will merge this PR.

clauswilke (approved):

PR #169, opened 2021-02-02
Convert Vec<Robj> to Robj

dfalbel:
Not sure if this is useful at all or if we want to have this kind of wrappers. 
It took me a while to realize it was this simple :)

### DIFF
--- a/extendr-api/src/robj/into_robj.rs
+++ b/extendr-api/src/robj/into_robj.rs
@@ -527,3 +527,10 @@ impl<'a> From<HashMap<&'a str, Robj>> for Robj {
         res
     }
 }
+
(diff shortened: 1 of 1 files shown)


andy-thomason:
This is a side effect of the List() wrapper which will convert many types, including Vec<Robj> to lists.

I would like to finish the NamedList wrapper if anyone wants to give it a go. The same deal, but generate
two vectors, one for the names and the other for the data from one (&str, Into<Robj>) iterator.

andy-thomason:
I'm trying to get time to document the wrappers in the book. Forgive me if I'm a bit slow.

andy-thomason:
I'll try to put a List example into the main module documentation.

PR #170, opened 2021-02-02
Consider always exporting the `$` method

dfalbel:
I think we should consider always exporting the `$` method for implementations.

In my use case I have a class that is only used as output for a function, so I don't want to expose it's constructors. However I want the `$` methods to be accessible to users can do `$method()` on the output objects.

It's also considered a good practice in the R packages book:

> A method for an internal generic: technically, you don’t need to export these methods. However, I recommend exporting every S3 method you write because it’s simpler and makes it less likely that you’ll introduce hard to find bugs. Use devtools::missing_s3() to list all S3 methods that you’ve forgotten to export.

[Link](https://r-pkgs.org/namespace.html?q=internal#export-s3)


andy-thomason (approved):
Looks good to me.

extendr-api/src/metadata.rs
```diff
@@ -266,6 +266,11 @@ fn write_impl_wrapper(
         writeln!(w, "#' @rdname {}", imp.name)?;
         writeln!(w, "#' @usage NULL")?;
         writeln!(w, "#' @export")?;
+    } else {
+        // we always export the `$` method so the method is correctly
+        // added to the NAMESPACE even if we don't export the class
+        // itself and it's initializers.
```
clauswilke, line 272: minor typo: "its", not "it's"

dfalbel: Thanks! just fixed!

andy-thomason (approved):

Issue #127, opened 2021-01-14
Add rextendr use case to integration tests


clauswilke:
The current integration tests test whether we can use extendr code in an R package, but they don't test whether we can call `rust_source()` or `rust_function()` from rextendr. I think we should add such a test, to make sure we don't accidentally break rextendr down the road.

cc @Ilia-Kosenkov

Ilia-Kosenkov:
PR #173 introduced a new CI step that duplicates `R CMD check` procedure also for the latest version of {rextendr}.
{rextendr}'s dynamically-generated rust libraries are linked to the dev version of `extendr-api` that is tested by the CI.

This solution ensures that only one {rextendr} test codebase exists, and that any new features added to and tested by {rextendr}  are also automatically tested by `extendr` CI.

PR #173, opened 2021-02-03
Dumb rextendr integration tests

### LINKED ISSUE #171: Failed R CMD check will not be uploaded
In one of the previous PRs `rcmdcheck` was updated and `check_dir` was set to `"../../../"`, which is actually *outside* of repo's root.
This was done to reference `extendr-*` crates using relative paths.
https://github.com/extendr/extendr/blob/f235216793607e59d7d87877fcf927e08a4bd654/.github/workflows/test.yml#L171

The problem is here
https://github.com/extendr/extendr/blob/f235216793607e59d7d87877fcf927e08a4bd654/.github/workflows/test.yml#L174-L179

If the check fails, artifacts are uploaded from `./check/`, which is no longer produced.
In fact, because check dir is now outside of repo's root, artifacts contaminate the folder outside of extendr's workspace, which is not good. Also, because it is `../../../` and not `../../../check_dir`, artifacts are on the same level as `extendr` repo folder. If I run rcmdcheck on my local machine, then the `../` folder will look like:
```
extendr
extendrtests.Rcmdcheck
extendrtests_0.1.0.tar.gz
```
And I have to remember to delete these artifacts every time.

Ilia-Kosenkov:
`tests/extendrtests` test integration of Rust code into an R package. `{rextendr}` generates rust libraries on the fly and loads them dynamically, using a slightly different method. 
So in #127 it was suggested to test `{rextendr}` as well.

I can suggest two obvious solutions:
1. Write more tests in this repo, testing `{rextendr}` behavior. However, we will have to watch `{rextendr}` changes and update tests to keep up, effectively maintaining two test codebases: here and in `extendr/rextendr`.
2. Use `{rextendr}` itself.

In this PR I suggest using `{rextendr}` as a test framework. We can run `rcmdcheck` on `{rextendr}` and verify its tests pass.
However, there is (yet again) an issue similar to #89 -- `{rextendr}` auto-generated `Cargo.toml` should point to the last commit in the currently tested/merged branch.

I patched `{rextendr}` https://github.com/extendr/rextendr/pull/31, allowing usage of environemnt variables to configure how rust libraries are build, in particular, `REXTENDR_TOOLCHAIN` configures an overriding rust toolchain, and `REXTENDR_PATCH_CRATES_IO` overrides crate references.
So, using `patch.crates_io` it is possible to point `{rextendr}` to the root of the tested extendr repo, effectively fixing #127.

I also found an issue with `extendrtests` and artifact uploading, so this PR also fixes #171 (this is a minor tweak).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -205,6 +205,20 @@ jobs:
             cat("::warning:: R version is not compatible with 'rextendr'\n")
           } else {
             cat("::group::Checking 'rextendr'\n")
+            
+            patch.crates_io <-
+              paste(
+                paste0(
+                  "extendr-api = { path = \"",
+                  normalizePath(file.path(getwd(), "extendr-api"), winslash = "/"),
+                  "\" }"),
+                paste0(
+                  "extendr-macros = { path = \"",
+                  normalizePath(file.path(getwd(), "extendr-macros"), winslash = "/"),
+                  "\" }"),
+                sep = ";")
+
+            Sys.setenv(REXTENDR_PATCH_CRATES_IO = patch.crates_io)
             rcmdcheck::rcmdcheck(
               path = "tests/rextendr", 
               args = c("--no-manual", "--as-cran", "--force-multiarch"), 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -155,6 +155,7 @@ jobs:
         run : |
           git clone https://github.com/extendr/rextendr ./tests/rextendr
 
+      # Dependencies from both test packages are combined
       - name: Query dependencies for integration testing
         run: |
           install.packages('remotes')
@@ -176,6 +177,7 @@ jobs:
           key: ${{ runner.os }}-${{ hashFiles('.github/R-version') }}-1-${{ hashFiles('.github/depends.Rds') }}
           restore-keys: ${{ runner.os }}-${{ hashFiles('.github/R-version') }}-1-
 
+      # Currently rextendr requires R > 4, so if R version is older (e.g. 3.6.3), rextendr tests are omitted
       - name: Install R dependencies for integration testing
         run: |
           remotes::install_deps(pkgdir = "tests/extendrtests", dependencies = TRUE)
@@ -185,6 +187,7 @@ jobs:
         
       # Windows: modified PATH will launch 32-bit Rscript if the build targets only i686/i386, so no conflict will arise
       # Rust pseudo multi-targeting: R CMD CHECK receives an extra argument '--no-multiarch' if BUILD_TARGETS is not 'default' and does not contain any commas
+      # If R is older than 4.0, rextendr is not tested and a warning is emitted. 
       - name: Run R integration tests
         id: r_integration_tests
         env:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,7 +5,6 @@ on:
     branches:
       - main
       - master
-      - 'rextendr-tests'
   pull_request:
     branches:
       - main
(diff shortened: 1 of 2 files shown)


tests/extendrtests/src/rust/Cargo.toml
```diff
@@ -19,7 +19,10 @@ extendr-api = "*"
 ## Build against local extendr sources. Check locally by 
 ## running rcmdcheck::rcmdcheck(check_dir = "../../../") from
 ## the root of extendrtests.
-extendr-api = { path = "../../../../../extendr/extendr-api/"}
+## The code is compiled in 
+## 'extendr/extendrtests_check/extendrtests.Rcheck/00_pkg_src/extendrtests/src/rust'
+## Cargo should go (relatively) 7 directories up (../) and then into './extendr/extendr-api'.
+extendr-api = { path = "../../../../../../../extendr/extendr-api/"}
```
clauswilke, line 25: With this change, can you still build extendrtests locally from within RStudio by clicking on "Install and Restart"? That was the reason for the strange setup, that both the local build and the CI runs would work.

Ilia-Kosenkov: Oh, sorry, I completely forgot about RStudio, I don't use it that much.
Let me try and see what happens.

clauswilke: Yeah, it probably won't work.

As I said elsewhere, you can move everything one folder down and then you can put the check files into the root of extendr rather than parallel to the root. I don't have a strong opinion one way or another.

clauswilke: E.g., `extendr/tests/R-integration/extendrtests/`

Ilia-Kosenkov: @clauswilke,
I actually encountered another problem with paths -- something related to how `tar` stores paths and a limit of 100 symbols or so, so the smaller the (absolute) paths are the better. 
I'd like to first look for some non-intrusive solutions before moving everything around.

Ilia-Kosenkov: @clauswilke,
With our relative path structure, neither "Test Package" nor "Check package" work from the RStudio menu, only "Install and Restart".

clauswilke: Test package should work. It doesn't actually compile anything, it just runs R code. I just tried with current master and all tests passed.

Check package does not work, and I don't think there's a way to fix this. But it's also not that important for development work, what is important is that we can easily install the package and run the testing code.

.github/workflows/test.yml
```diff
@@ -154,29 +167,97 @@ jobs:
...
+
+          api_path <- normalizePath(file.path(getwd(), "extendr-api"), winslash = "/")
+          toml_path <- file.path(getwd(), "tests", "extendrtests", "src", "rust", "Cargo.toml")
+          lines <- readLines(toml_path)
+          lines <- gsub(
+            "(^\\s*extendr-api\\s*=\\s*\\{\\s*path\\s*=\\s*\")(.*?)(\"\\s*\\})",
+            paste0("\\1", api_path, "\\3"),
+            lines
+          )
+          writeLines(lines, toml_path)
```
Ilia-Kosenkov, line 188: This is no great solution, but I find this the easiest way to solve the path problem on CI: simply inject the correct absolute path. 
Cargo.toml is configured to work with RStudio (as requested).
This can be revised/improved if `patch.crates-io` is no longer used or significantly modified.

clauswilke: Ah, Ok, this seems fine.

Ilia-Kosenkov:
@clauswilke,
This PR lost attention, is there anything else that should be done before merging?

clauswilke:
Can you merge in master so we can see how the tests do now?

clauswilke:
I actually meant merge master into this branch and rerun the tests, not merge this PR into master, but it's fine either way.

Ilia-Kosenkov:
@clauswilke,
Oh, sorry, I misunderstood you.
I hope it is OK. I watched CI output and it seems OK, including Windows-x86 with disabled `catch_r_error` test and new {rextendr} integration.

PR #175, opened 2021-02-10
Avoid r preserve object

andy-thomason:
This PR introduces an alternative to `R_PreserveObject`.

This implementation tracks `SEXP` pointers and reference counts them so we can implement
safer owned `Robj` instances.

This should allow us to implement `Clone` on Robj by incrementing the reference.

It will not be cost free, but probably better than the linked-list version of `R_PreserveObject`
which is the default.

We will have to benchmark it in due course and may be able to improve the performance with
a custom hasher.

### DIFF
--- a/extendr-api/src/ownership.rs
+++ b/extendr-api/src/ownership.rs
@@ -319,7 +319,7 @@ mod test {
 
                 // Make some test objects.
                 let sexp_pres = Rf_allocVector(VECSXP, test_size as R_xlen_t);
-                R_PreserveObject(sexp_pres);
+                Rf_protect(sexp_pres);
 
                 let sexps = (0..test_size).map(|i| {
                     let sexp = Rf_ScalarInteger(1);
@@ -341,7 +341,7 @@ mod test {
                     own.check_objects();
                 }
 
-                R_ReleaseObject(sexp_pres);
+                Rf_unprotect(1);
             });
         }
     }
--- a/extendr-api/src/ownership.rs
+++ b/extendr-api/src/ownership.rs
@@ -1,0 +1,328 @@
+//! Maintain ownership of R objects.
+//!
+//! This provides the functions protect() and unprotect()
+//! A single preserved vector holds ownership of all protected objects.
+//!
+//! Objects are refernce counted, so multiple calls are possible,
+//! unlike R_PreserveObject.
+//!
+//! This module exports two functions, protect(sexp) and unprotect(sexp).
+
+use lazy_static::lazy_static;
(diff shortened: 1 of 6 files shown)


andy-thomason:
I suspect that the large number of allocations in the test are exposing weaknesses elsewhere hence
the occasional CI failures.

extendr-api/src/ownership.rs
```diff
@@ -0,0 +1,328 @@
...
+                    own.check_objects();
+                }
+
+                // Force a garbage collect.
+                let test_size = INITIAL_PRESERVATION_SIZE + EXTRA_PRESERVATION_SIZE * 5;
+
+                // Make some test objects.
+                let sexps = (0..test_size).map(|_| {
+                    let sexp = Rf_ScalarInteger(1);
+                    R_PreserveObject(sexp);
```
clauswilke, line 308: I bet here you're triggering the issue described here: https://github.com/RcppCore/Rcpp/issues/1081
The crash was fixed in recent versions of R, that's why only the oldrel CI runs fail.

Any reason you're using `R_PreserveObject()` instead of `Rf_protect()`? The latter shouldn't cause the problem.

andy-thomason:
I did originally use protect, but overflowed the stack limit.

I'll check it out in the morning.

Andy.

clauswilke:
Yeah, I guess `Rf_protect()` is meant for rapid protecting and unprotecting. Maybe just allocate an R vector and add these pointers to that vector, just like you do in your actual protection code also.

andy-thomason:
Good idea.

extendr-api/src/ownership.rs
```diff
@@ -303,9 +318,12 @@ mod test {
                 let test_size = INITIAL_PRESERVATION_SIZE + EXTRA_PRESERVATION_SIZE * 5;
 
                 // Make some test objects.
-                let sexps = (0..test_size).map(|_| {
+                let sexp_pres = Rf_allocVector(VECSXP, test_size as R_xlen_t);
+                R_PreserveObject(sexp_pres);
```
clauswilke, line 322: While this will work, it would be more idiomatic at this point to use `Rf_protect()`. You’re just protecting the `SEXP`, doing a quick calculation with it, and then release again.

andy-thomason:
:tada: Build is green.

andy-thomason:
I'll update it, then do a merge.

This PR is quite a big change as it affects the object model in fundamental
ways.

We may be able to deal with mutability in a more sensible way after this.

PR #176, opened 2021-02-13
Named lists

andy-thomason:
This was an attempt to generate a named list wrapper, but evolved into a re-think of
attributes and the "borrowed" Robj type.

This PR adds specialised functions to set attributes, especially "set_names" which
enables construction of named lists and other vector types with correct type checking.

There is also a "set_class" for classes.

We acknowledge that setting attributes is a dangerous business and so there
are new error types to handle this.

"set_attrib" now returns the object, not the value set, so they can be chained.

This means that the NamedList type is unnecessary.


extendr-api/src/matrix.rs
```diff
@@ -161,9 +161,10 @@ where
     fn from(array: RArray<T, [usize; 2]>) -> Self {
         let res = array.data.into();
         let dim = [array.dim[0] as i32, array.dim[1] as i32];
-        res.set_attrib(Symbol("dim"), dim);
-        res.set_attrib(Symbol("class"), "matrix");
-        res
+        res.set_attrib(dim_symbol(), dim)
+            .unwrap()
+            .set_attrib(class_symbol(), "matrix")
+            .unwrap()
```
clauswilke, line 167: Here and throughout: is it a good idea to have all these `unwrap()` calls in the API? Can we somehow catch potential errors properly, or at least use `expect()`?

clauswilke (approved):
It generally looks good to me, except all these `unwrap()` calls that were added. Is there anything that can be done about that?

extendr-api/src/matrix.rs
andy-thomason: Sounds good. These errors should never occur because the preconditions are met, but expect is better for locating regressions.

PR #177, opened 2021-02-14
Create RType and rtype(). Move parse and eval_string to functions.

andy-thomason:
This is a response to a request from @Ilia-Kosenkov to mirror the SEXPTYPE enum in Rust.

We introduce the RType enum and the method Robj::rtype.

As a side-effect, move parse and eval_string to functions and lose the Robj:: prefix.


clauswilke:
Since you"re calling it `RType` and not `SEXPType`, would it make sense to also use human readable values? Something like `RType::Character`, `RType::Integer`, `RType::Real`, `RType::Closure`, `RType::Pairlist`, etc.

andy-thomason:
That would certainly be more readable.

Ilia-Kosenkov:
@clauswilke, @andy-thomason,
I am all for readable (`{cpp11}` - style) names.

andy-thomason:
The names mostly match the corresponding wrappers.

I may also make an RAny enum which converts to/from any of the wrappers.

extendr-api/src/lib.rs
```diff
@@ -370,6 +370,36 @@ impl IsNA for &str {
...
+    Dot,
+    Any,
+    List,
+    Expression,
+    Bytecode,
+    ExternalPtr,
+    WeakRef,
+    Raw,
+    S4,
+    Unknown,
```
clauswilke, line 400: Could you add comments here for each stating which SEXPTYPE it corresponds to? Like so?

```
Logical,  // LGLSXP
Integer,  // INTSXP
Real,       // REALSXP
```
etc?

andy-thomason:
Ok. I'll do this.

PR #181, opened 2021-02-21
Generic FromRobj for Vec<T>

dfalbel:
Often I found that I had to write `FromRobj` for a type `T` and for `Vec<T>`.

I wonder if it's possible to have a general `impl FromRobj<'a, T> for Vec<T>` given that `impl FromRobj` for that type is already implemented.

I started implementing something but didn't go far because I get this, of course, because we already have many specific implementations.

> conflicting implementation for `std::vec::Vec<i32>`

Is it possible to express the 'use the generic unless an specific implementation exists` in the `where` clause?


clauswilke:
Are the specific implementations needed if you write the generic one?

dfalbel:
I think that for performance reasons we probably want to keep them.

andy-thomason:
In the example you give, the List wrapper does precisely this.
We cannot impl FromRobj for a generic as Rust demands that
all conversions be disjoint forever. C++ is much more permissive
with often disastrous results.

Once you have constructed a List, you should be able to move out
of it with var.0

As a Rust user it seems unintuitive to convert iterators to vectors
unless the lifetime rules demand it, for example if you are
returning results from a function or storing results in a structure
with a long lifetime.

The iterators and wrappers exist to make these conversions easy
with map able to convert between any types.

While we could add a broad spectrum of FromRobj, we only really
need the ones that actually represent data as everything else can
be done with a map.

The FromRobj are only intended to map input parameters.

INTSXP -> &[i32] or Int.
REALSXP -> &[f64] or Real
LOGICALSXP -> &[Bool] or Logical
VECSXP -> ListIter
STRSXP -> StrIter

If you want some other type, just use:

fn x(a: Int) {
   let other_int : Vec<MyType> = a.map(|v| v.into()).collect();
   ...
}

At present, we can't map VECSXP to &[Robj] because
Robj is larger than the SEXP pointer, but I hope to reduce
Robj to a map of a SEXP pointer with the next iteration.

so untimately.

VECSXP -> &[Robj] or ListIter

The iterators and slices are zero-copy which is a critical concept
if you want performance.

In RCpp, beginner users always convert R objects to vectors
before use, and there are good reasons for this as indexing
vectors is inefficient in Rcpp. But if you want to index a vector
in extendr, just use a slice.

Saying this, if it suits new users to do this, we may have to support
it but encourage through documentation to use iterators
which will be better supported in future versions.

Can you give an example use case where this would be useful?

Sorry to go on...

andy-thomason:
Saying this, it seems there is no generic `FromRobj for List<T> where T: Into<Vec<E>>` which is a significant 
issue.

andy-thomason:
I'm looking at a way to do this `List<T> where T: FromIterator<E>, E: FromRobj`.

I'll get back with results if it is possible.

dfalbel:
I am still understanding how Rust works so probably forcing very unintuitive patterns.
I am mostly writing glue code, so I am passing Rust objects to R and back all the time.

Here is an example of something that I spent a quite some time to figure out:

https://github.com/mlverse/hftokenizers/blob/842f7a4808fe69cc46e8c4efbf9121a3818ef408/src/rust/src/tokenizer.rs#L86-L138

This function takes an R list ( with 1 or 2 character vectors) then converts this into a Vec<tokenizer::EncodeInputs> and make the call. We then massage the `Vec<REncoding>` output into an `Robj` to be able to return.

My initial intuition is that if I could write a `impl FromRobj for tokenizer::EncodeInputs` and since `REncoding` is known to extendr, because `REncoding` is exported. I cold reduce this function to something like:

```
fn encode_batch (&self, inputs: Vec<EncodeInputs>, add_special_tokens: bool) -> Vec<REncoding> {
        self
            .tokenizer
            .encode_batch(inputs, add_special_tokens)
            .unwrap()
            .iter()
            .collect()
}

**Comment by andy-thomason** [comment]
I can see why this is a challenge. I though that I had this under control,
but it
is bigger than I realised.

I would like it to be easy to write converters to/from lists for structs
and vectors of structs.

I've been playing with a pattern for struct deserialisation.

I'll try to publish an example if I get a chance tomorrow.

Meanwhile, apologies for the big posts, I very much appreciate
you trying to work with what we have and giving us some excellent
feedback.

**Comment by dfalbel** [comment]
No worries! The long posts are great for learning :)
`extendr` is a really nice project!

**Comment by andy-thomason** [comment]
Likewise with R and ggplot. I'm learning quite a bit about the language and
visualsation.

I'm giving a talk on R internals at work next Monday for some of our R
users including Gil McVean and Chris Spencer
who are stats masters - this will test my understanding...

**Comment by andy-thomason** [comment]
After a lot of experimentation, the closest I can get is this:

FromRobj was always a bit of a compromise and we should have

a) Used TryFrom
b) Used a  standard Error type.

If you implement `TryFrom<Robj>` for your type, this will use the robj
directly, consuming the robj 

```
use std::convert::{Into, TryFrom};

impl TryFrom<Robj> for i32 {
    type Error = &'static str;

    fn try_from(robj: Robj) -> std::result::Result<Self, Self::Error> {
        if let Some(val) = robj.as_integer() {
            Ok(val)
        } else {
            Err("Input must be an integer.")
        }
    }
}

impl<'a, T: TryFrom<Robj>> FromRobj<'a> for List<Vec<T>>
where
    <T as TryFrom<Robj>>::Error: Into<&'static str>,
    Vec<T>: std::iter::FromIterator<T>,
{
    fn from_robj(robj: &'a Robj) -> std::result::Result<Self, &'static str> {
        if let Some(iter) = robj.as_list_iter() {
            let inner = iter
                .map(|r| <T>::try_from(r).map_err(|e| e.into()))
                .collect::<std::result::Result<Vec<T>, &'static str>>();

            inner.map(|vec| List(vec))
        } else {
            Err("Input must be a list.")
        }
    }
}
#[test]
fn test() {
    test! {
        let l = r!(List(vec![1, 2, 3]));
        let v = <List<Vec<i32>>>::from_robj(&l);
        println!("{:?} {:?}", l, v);
        assert_eq!(v, Ok(List(vec![1, 2, 3])));
    }
}

```

andy-thomason:
I hope this works for your use case.

Note that using FromRobj tries to convert the temp object created by ListIter
to a reference which does not work.

andy-thomason:
@dfalbel I'm happy to take the time to explain this. It reached the limits of my Rust ability.

Do you want me to make a PR for this?

I'd be very pleased if you wanted to own this PR yourself.

dfalbel:
Thanks very much @andy-thomason ! 
I will try it out first thing tomorrow and I'll probably have questions.
Yes, I am happy to try finishing the PR.

andy-thomason:
Give me a call if you want some help on this. I'm happy to pair.

dfalbel:
Hi @andy-thomason, 

Sorry for the late reply! I got caught by a hairy bug in another project. 
I have updated the PR with the code chunk you provided but instead of making the generic for `List<Vec<T>>` I am using `Vec<T>`. 

This indeed makes much easier to take vectors of objects.

I wonder if you think it's worth investigating if we could have a macro that generates the `TryFrom<Robj> for Vec<T>` trait from a specific `FromRobj<'a> for T` trait. 

I am also happy to chat about this PR!

Best,

andy-thomason:
I'm going to close this as it has gone somewhat stale unless anyone objects.

You have inspired my to completely refactor the wrappers and restructure the project.

I'm happy to pair at any time if anyone wants to try new things.

PR #185, opened 2021-02-23
Use failure() to trigger steps on failures

### LINKED ISSUE #183: R CMD check results not uploaded
I noticed in this failed run
https://github.com/extendr/extendr/pull/182/checks?check_run_id=1952491229
that the step that uploads results is not triggered correctly:
https://github.com/extendr/extendr/blob/ddef053a00ca751921f36f7c98805e85c77000fc/.github/workflows/test.yml#L248-L260
I believe I messed up with the conditions when wrote this, can anyone take a look and help me diagnose this?

yutannihilation:
Fix #183

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -240,7 +240,6 @@ jobs:
               args = c("--no-manual", "--as-cran", "--force-multiarch"), 
               error_on = "warning", 
               check_dir = "rextendr_check")
-            stop()
             cat("::endgroup::\n")
           }
         shell: Rscript {0}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -246,14 +246,14 @@ jobs:
 
 
       - name: Upload extendrtests check results from R integration tests
-        if: steps.extendrtests_check.outcome != 'success'
+        if: failure() && steps.extendrtests_check.outcome != 'success'
         uses: actions/upload-artifact@main
         with:
           name:  extendrtests-${{ matrix.config.os }}-R-${{ matrix.config.r }}-rust-${{ matrix.config.rust-version }}
           path: extendrtests_check
 
       - name: Upload rextendr check results from R integration tests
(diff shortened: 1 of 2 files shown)


Ilia-Kosenkov:
@yutannihilation, could you simulate a failure in `rextendr` test step, not in `extendrtests`? 
As you can see, right now artifacts are uploaded for both tests even though only the first one failed.
I believe it is because if `extendrtests` fail, `rextendr` also does not succeed and hence `steps.rextendr_check.outcome != 'success'` is true.
We either need to learn all values of `outcome` and write suitable condition, or perhaps use something like:
`if: failure() && steps.extendrtests_check.outcome == 'success' && steps.rextend_check.outcome != 'success'` to specifically catch it when `extendrtests` worked but `rextendr` failed.

yutannihilation:
> We either need to learn all values of `outcome` and write suitable condition,

It's listed here:

> Possible values are success, failure, cancelled, or skipped. 
https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#steps-context

I'm still thinking what's the best condition here, but I'm getting inclined to use just `failure()`. Probably there won't be problems except for the artifact size to upload both results even when either of one didn't fail. `if: failure() && steps.extendrtests_check.outcome == 'success' && steps.rextend_check.outcome != 'success'` might be technically a correct choice, but I prefer simplicity.

clauswilke:
Agreed. Just uploading all artifacts upon failure should be fine. Github will delete them after a while anyways.

yutannihilation:
Confirmed now artifacts are uploaded. I think this is good to merge if we agree on using `if: failure()`.

https://github.com/extendr/extendr/actions/runs/596207390#artifacts

image

clauswilke (approved):

yutannihilation:
I'm merging this. I guess we'll want to improve this anyway (e.g. hundreds of megabytes is a bit too huge...). Let's revisit here when we find nicer way.

PR #188, opened 2021-03-08
Revise Function wrapper and support calls.

andy-thomason:
This PR converts the `Func<F, B, E>` wrapper to a Function wrapper supporting
calls.

This wrapper is more like an RCpp wrapper, having only a single SEXP.

This provides a simpler interface to users.

We can now create a function from scratch and call functions without the worry of the function
not being callable.

Both Function and Robj support call(pairlist!(...)).


clauswilke (approved):

yutannihilation:
While I think it's fine to merge this for now, I'm not sure if calling `pairlist()` to create a pairlist is a clean solution. We might need to generate pairlist directly using R's C API (just like cpp11 does [here](https://github.com/r-lib/cpp11/blob/3c65bc56250a882613ed919d271b789aad557402/inst/include/cpp11/function.hpp#L35-L53)) to avoid errors in some unusual situations, for example, when another `pairlist` is defined in the global environment.

``` r
rextendr::rust_function(r"(
fn call_with_one2(f: Robj) -> Robj {
  f.call(pairlist!(1)).unwrap()
}
)", patch.crates_io = list(`extendr-api` = list(git = "https://github.com/extendr/extendr", branch = "call-robj"))
)
#> build directory: /tmp/RtmpQVbHmP/fileaa29822beb636
#> Writting wrappers to:
#> /tmp/RtmpQVbHmP/fileaa29822beb636/target/extendr_wrappers.R

call_with_one2(function(x) x + 1)
#> [1] 2

pairlist <- stop
call_with_one2(function(x) x + 1)
#> Error in call_with_one2(function(x) x + 1): call_with_one2 paniced.
```

<sup>Created on 2021-03-09 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

yutannihilation:
To be clear, in the case of `data_frame!`, we have no choice other than actually calling `data.frame()` as it's a complex job to create a data.frame correctly. But, as pairlist is a very low-level one, I feel it might make sense to handle it at the level of R's C API.

(If we just want to prevent the corner case above from failing, maybe we can just merge this and revise it to `base::pairlist` after implementing support for `::`.)

clauswilke:
I second the concerns regarding pairlists. Also, creating a pairlist by calling `pairlist()` (or even worse, `base::pairlist()`) is quite slow. As function calls happen frequently, this matters a lot, and I'm always surprised how inefficient the calls in the R code are. Even functions such as `lang3()`, `lang4()`, `lang5()` etc are inefficient as they always start recursing from the front for each new argument they add.

A while back I wrote this little note as (in my opinion) the simplest and most efficient way to create a call from C:
https://github.com/hadley/r-internals/blob/master/pairlists.md#creating-a-call
We should mimic something like it in extendr.

Btw, where in the code is the error generated that is shown in the reprex? There's a typo, it should say "panicked" not "paniced".

yutannihilation:
"paniced" probably comes from here:

https://github.com/extendr/extendr/blob/48cb40c6271450204a736b2d466c0cd39fef4f03/extendr-macros/src/lib.rs#L291

extendr-api/src/wrapper/function.rs
```diff
@@ -0,0 +1,136 @@
...
+            Ok(f)
+        } else {
+            Err("Not a function")
+        }
+    }
+}
+
+impl From<Function> for Robj {
+    /// Make an robj from a function wrapper.
+    /// The function wrapper is guaranteed to cntain a function object.
```
yutannihilation, line 48:
```suggestion
    /// The function wrapper is guaranteed to contain a function object.
```

andy-thomason:
The variadic macros need a bit of work, ideally to use List() and Pairlist() wrappers.
I would like to make them into procedural macros as we can do more from there.
In particular, the R! macro could have a lot of intelligence and be parameterisable.

While I'm here, tidying up the design of the wrappers is a good goal, making them conform
to expectations.

I've started to split the wrappers into modules, the first being function.rs

I apologise if we are creating breaking changes but it it best to get them done early.

yutannihilation:
Thanks for clarification. Considering we are going to change the crate structure, I think we should just merge this and do some tidying later. The interface looks good to me at least!

extendr-api/src/wrapper/function.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    /// Do the equivalent of x(a, b, c)
+    /// ```
+    /// use extendr_api::prelude::*;
+    /// test! {
+    ///     let function = R!(function(a, b) a + b).unwrap().as_func().unwrap();
+    ///     assert_eq!(function.call(pairlist!(a=1, b=2)).unwrap(), r!(3));
+    /// }
+    /// ```
+    pub fn call(self, args: Robj) -> Result<Robj> {
```
yutannihilation, line 100: Sorry, this might be a Rust newbie question, but can this be `&self`? In my understanding, if I want to call `Function` more than twice, I always need to `clone()`. But if I treat it as `Robj`, there's no problem (as its `call()` takes `&self`).

``` r
rextendr::rust_function(r"(
fn call_twice(f: Function) -> Robj {
  println!("{:?}", f.call(pairlist!(1)));
  f.call(pairlist!(1)).unwrap()
}
)", patch.crates_io = list(`extendr-api` = list(git = "https://github.com/extendr/extendr", branch = "call-robj"))
)
#> build directory: /tmp/RtmpIw6Qen/filebb53e4047ac42
#> Error: Rust code could not be compiled successfully. Aborting.

rextendr::rust_function(r"(
fn call_twice(f: Robj) -> Robj {
  println!("{:?}", f.call(pairlist!(1)));
  f.call(pairlist!(1)).unwrap()
}
)", patch.crates_io = list(`extendr-api` = list(git = "https://github.com/extendr/extendr", branch = "call-robj"))
)
#> build directory: /tmp/RtmpIw6Qen/filebb53e4047ac42
#> Writting wrappers to:
#> /tmp/RtmpIw6Qen/filebb53e4047ac42/target/extendr_wrappers.R

call_twice(function(x) x + 1)
#> [1] 2
```

<sup>Created on 2021-03-10 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

clauswilke:
I think it's inevitable that we'll make a few more breaking changes, so I agree let's just get them out of the way.

andy-thomason:
With the separation into modules, it is easier to see what is implemented for each type.

extendr-api/src/wrapper/function.rs
andy-thomason: Thanks for spotting this. It was a side effect of the earlier design...

andy-thomason:
どうもありがとう!

yutannihilation:
こちらこそ！

PR #190, opened 2021-03-09
More wrapper tidying

andy-thomason:
This PR is an overhaul of the wrapper and conversion process.

Whilst very similar on the top, we will now wrap an Robj in a struct and implement
functions more in the RCpp/C++11 model.

I concede that this a philosophical U-turn, but it makes sense.

Each wrapper will now contain an Robj of the correct type.

Each wrapper implements:

A "from_xxx" impl which constructs the type from components.

A FromRobj impl for legacy parameters.
A TryFrom<Wrapper> impl for new style of parameters.
A From<Robj> impl for returning values.

We hope to migrate from FromRobj to TryFrom in due course.
The TryFrom uses the official extendr Error type to return failure
if the object type is unexpected.

All the as_* impls in Robj can be replaced by try_from versions
we can write a macro to do this.

The names of the is_* impls should match the type.


extendr-api/src/wrapper/pairlist.rs
```diff
@@ -0,0 +1,90 @@
+use super::*;
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct Pairlist {
+    robj: Robj,
+}
+
+impl<'a> FromRobj<'a> for Pairlist {
+    /// Convert an object that may be null to a rust type.
```
clauswilke, line 9: Copy-paste error? This is about creating pairlists, not nullables

extendr-api/src/wrapper/primitive.rs
```diff
@@ -0,0 +1,36 @@
...
+#[derive(Debug, PartialEq, Clone)]
+pub struct Primitive<'a>(pub &'a str);
+
+impl<'a> From<Primitive<'a>> for Robj {
+    /// Make a primitive object, or NULL if not available.
+    /// ```
+    /// use extendr_api::prelude::*;
+    /// test! {
+    ///     let builtin = r!(Primitive("+"));
+    ///     let special = r!(Primitive("if"));
```
clauswilke, line 22: Should there be a test here? E.g. that `builtin` and `special` are not `NULL`? And maybe add a case where the result is `NULL`?

extendr-api/src/wrapper/raw.rs
```diff
@@ -0,0 +1,49 @@
...
+            for (i, &v) in val.iter().enumerate() {
+                *ptr.offset(i as isize) = v;
+            }
+            Robj::Owned(sexp)
+        })
+    }
+}
+
+impl<'a> FromRobj<'a> for Raw<'a> {
+    /// Convert an input value to a Raw wrapper around bytes.
```
clauswilke, line 34: Should "Raw" be in backticks here?

clauswilke:
I have read through the entire PR and marked a few minor issues in the documentation. It all broadly looks good to me, but I'm not sure I fully understand the implications, so somebody else should review also.

andy-thomason:
Thanks for taking a look, @clauswilke It is still a WIP, so things may resolve in the end.

I may remove FromRobj completely in favour of TryFrom with proper error handling
(the big new Error enum).

extendr-api/src/wrapper/mod.rs
```diff
@@ -102,42 +102,59 @@ macro_rules! make_conversions {
...
     "Not an Environment"
 );
 
+make_conversions!(List, ExpectedList, is_list, "Not a List");
+
+make_conversions!(
+    Expression,
+    ExpectedExpression,
+    is_expression,
+    "Not an Expression"
```
clauswilke, line 126: We should be consistent with capitalization here. Compare "Not an Expression" to "Not a function".  I would probably use lower case throughout, but upper case is fine if it's consistently used.

andy-thomason: It has been a bit of a whirl...

Thanks for spotting these typos.

I would love to drop `FromRobj` entirely in the future and convert everything to `TryFrom`
using the formal Error. The error would implement `Display` to generate the error message,

The essence is that From and `TryFrom` handle all the input and output cases.

The `Deref` means that you can use the wrapper as a `Robj`.

Rust doesn't have inheritance, per-se, but `Deref` allows you to access the `Base class`
which in this case is the wrapped Robj.

You should now be able to use just `Function`, `Environment`, `Symbol` etc. as parameter
types. These types provide access to the underlying object.

I plan to move many methods from Robj to their relevant wrappers.

The real test will be how easy it is for library writers to get to grips with this.

extendr-api/src/wrapper/environment.rs
```diff
@@ -6,6 +6,45 @@ pub struct Environment {
...
+    /// use extendr_api::prelude::*;
+    /// test! {
+    ///     let env = Environment::new();
+    ///     assert_eq!(env.len(), 0);
+    /// }
+    /// ```
+    pub fn new() -> Self {
+        // 14 is a reasonable default.
+        Environment::new_with_capacity(14)
+    }
```
clauswilke, line 20: Is there any reason we don't have a `new()` function that allows us to specify a parent environment? That's a very common use case.

extendr-api/src/wrapper/list.rs
```diff
@@ -60,7 +73,7 @@ impl IntoIterator for List {
     type IntoIter = NamedListIter;
     type Item = (&'static str, Robj);
 
-    /// Convert a List into an interator, consuming th epairlist.
+    /// Convert a List into an interator, consuming the pairlist.
```
clauswilke, line 76: Did you mean "consuming the list"? A pairlist is different from a regular list. I think here we're dealing with regular lists.

extendr-api/src/wrapper/environment.rs
andy-thomason: I'll put one in...

yutannihilation:
It seems the test failures are caused by the recent changes on rextendr's side, not by the changes of this pull request. Will investigate.

https://github.com/extendr/extendr/pull/190/checks?check_run_id=2108144242#step:16:394

```
  Error: Error: No root directory found in D:/a/extendr/extendr/rextendr_check/rextendr.Rcheck/tests_i386/testthat or its parent directories. Root criterion: contains a file "DESCRIPTION" with contents matching "^Package: "
  Backtrace:
       x
    1. +-testthat::expect_equal(rust_eval("2 + 2"), 4) test-eval.R:2:2
    2. | \-testthat::quasi_label(enquo(object), label, arg = "object")
    3. |   \-rlang::eval_bare(expr, quo_get_env(quo))
    4. \-rextendr::rust_eval("2 + 2")
    5.   \-rextendr::rust_function(code = code_wrapped, ...)
    6.     \-rextendr::rust_source(code = code, env = env, ...)
    7.       \-rextendr:::make_wrappers(...)
    8.         \-rextendr:::pretty_rel_path(outfile, search_from = path)
    9.           +-fs::path_rel(path, start = rprojroot::find_package_root_file(path = search_from))
   10.           | +-fs::path_abs(path_expand(start))
   11.           | | \-fs::path_norm(path(getwd(), start))
   12.           | \-fs::path_expand(start)
   13.           \-rprojroot::find_package_root_file(path = search_from)
   14.             \-rprojroot::find_root_file(..., criterion = criterion, path = path)
   15.               \-rprojroot::find_root(criterion = criterion, path = path)
```

andy-thomason:
Cheers for looking at the build.

This PR should give a lot more options for conversions and tighter type control.

For example.

```
global_env() is now an Evironment

Functions can be built from Pairlist, Language and Environment.

Almost everything is now an Robj wrapper and can return from
an R function.

There are TryFrom conversions for nearly everything.

The Matrix classes have been re-written and also contain an Robj.
```
If you all can give me feedback on what methods you would like, I can add some
more.

Now I'm keen to start writing some real applications to show it off.

extendr-api/src/wrapper/environment.rs
andy-thomason: Note that the parent must be an Environment. This should tighten things up and reduce checks.

clauswilke:
@andy-thomason I'd still say a high-priority use case is calling functions from a package and capturing the result. Let's say the following:
```r
data <- tibble::tibble(x = 1:5, y = 6:10)
```
Anything you can do to make this type of use as simple as possible would be great.

clauswilke:
Maybe this would argue for a `Function::find_in_namespace()` constructor? So I can write something like:
```rust
let fun = Function::find_in_namespace("tibble", "tibble");
let out = fun.call(pairlist!(x = r!(1..=5), y = r!(6..=10)))
```

yutannihilation:
Not sure if this is a scope of this pull request, but I really want to build a `Pairlist` (or `List`) with some elements unnamed programmatically: https://github.com/extendr/extendr/issues/191

IIUC, current implementation is not correct in that it assumes `NA` as the unnamed one, but it's actually `""` (empty string).

``` r
names(pairlist(1, a = 2))
#> [1] ""  "a"
```

<sup>Created on 2021-03-15 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

Look at here. The problem is the line calling `sym_pair()`.

https://github.com/extendr/extendr/blob/660ad8c17952566de5764aaaff18554b1742323e/extendr-api/src/wrapper/pairlist.rs#L29-L37

It attempts to convert `name` into a symbol, but an empty string cannot be a symbol (got an error "attempt to use zero-length variable name"). Should this really be converted to symbols? As `set_names()` works for an empty string, I feel something might be wrong here.

https://github.com/extendr/extendr/blob/dda901ad20cc5d233a9dadd0f6c604d56910f001/extendr-api/src/wrapper/mod.rs#L597-L599

Ilia-Kosenkov:
Just to add a little bit of context: names can be set using e.g. `names<-` function and can be (almost) arbitrary strings. For example, several identical names are allowed:
<details>

``` r
l <- list(1, 2, 3, 4, 5, 6)
names(l) <- c("", "a", "", "a", "", "a")
l
#> [[1]]
#> [1] 1
#> 
#> $a
#> [1] 2
#> 
#> [[3]]
#> [1] 3
#> 
#> $a
#> [1] 4
#> 
#> [[5]]
#> [1] 5
#> 
#> $a
#> [1] 6

l <- list(1, 2)
names(l) <- c("+", "...")
l[["+"]]
#> [1] 1
l[["..."]]
#> [1] 2
```

<sup>Created on 2021-03-15 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>


</details>

clauswilke:
Tests mostly work now. Not clear to me whether the fail for R devel is a fluke or a systematic problem.

andy-thomason:
With from_pairs(...) I've arranged so that na_str() should not set the tag. I should also arrange so that an empty string does
the same. List uses empty strings to signal "no tag".

```
> l <- list(a=1, 2)
> .Internal(inspect(l))
@55f60dd6ace8 19 VECSXP g0c2 [REF(1),ATT] (len=2, tl=0)
  @55f60f15f6e0 14 REALSXP g0c1 [REF(3)] (len=1, tl=0) 1
  @55f60f15f6a8 14 REALSXP g0c1 [REF(3)] (len=1, tl=0) 2
ATTRIB:
  @55f60f1740d0 02 LISTSXP g0c0 [REF(1)] 
    TAG: @55f60d226990 01 SYMSXP g0c0 [MARK,REF(5645),LCK,gp=0x4000] "names" (has value)
    @55f60dd6ad28 16 STRSXP g0c2 [REF(1)] (len=2, tl=0)
      @55f60d507140 09 CHARSXP g0c1 [MARK,REF(6),gp=0x61] [ASCII] [cached] "a"
      @55f60d229058 09 CHARSXP g0c1 [MARK,REF(8),gp=0x60] [ASCII] [cached] ""
```

```
> p <- pairlist(a=1, 2)
> .Internal(inspect(p))
@55f60f25a9e8 02 LISTSXP g0c0 [REF(1)] 
  TAG: @55f60d5e2b60 01 SYMSXP g0c0 [MARK,REF(23)] "a"
  @55f60f15f600 14 REALSXP g0c1 [REF(5)] (len=1, tl=0) 1
  @55f60f15f5c8 14 REALSXP g0c1 [REF(5)] (len=1, tl=0) 2
```

I'm trying to make List and Pairlist look much the same. Both take (string-ish, object-ish) tuples and I'll fix the list! and pairlist! macros to generate these more efficiently when we are happy.

extendr-api/src/error.rs
```diff
@@ -56,17 +55,87 @@ pub enum Error {
...
 
-pub type Result<T> = std::result::Result<T, Error>;
-
 impl std::fmt::Display for Error {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{:?}", self)
+        match self {
+            Error::Panic => write!(f, "Panic detected."),
+            Error::NotFound => write!(f, "Not found."),
+            Error::EvalError(robj) => write!(f, "Evalualtion error in {:?}.", robj),
```
clauswilke, line 78: Typo: "Evaluation"

extendr-api/src/error.rs
```diff
@@ -56,17 +55,87 @@ pub enum Error {
...
     Other(String),
 }
 
-pub type Result<T> = std::result::Result<T, Error>;
-
 impl std::fmt::Display for Error {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{:?}", self)
+        match self {
+            Error::Panic => write!(f, "Panic detected."),
```
clauswilke, line 76: The error messages in this long list of errors are inconsistent in whether they do or don't end in a period. I think they should all be the same.

extendr-api/src/error.rs
```diff
@@ -56,17 +55,87 @@ pub enum Error {
...
+        }
     }
 }
+pub type Result<T> = std::result::Result<T, Error>;
+
+// impl std::fmt::Display for Error {
+//     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+//         write!(f, "{:?}", self)
+//     }
+// }
```
clauswilke, line 138: Is there a reason this code is commented out?

yutannihilation:
@andy-thomason 
> I've arranged so that na_str() should not set the tag.

If this works, I'm fine. But I suspect this condition is never met

``` rust
     if !name.is_na() { 
```

because a symbol cannot be missing.

``` r
rextendr::rust_function(r"(
fn f() -> bool {
  let x = Symbol::from_str(na_str());
  x.is_na()
}
)", patch.crates_io = list(`extendr-api` = list(git = "https://github.com/extendr/extendr", branch = "more-wrapper-tidying"))
)
#> build directory: /tmp/RtmpC8nsIa/fileb52b7bccd66e
#> ✓ Writting wrappers to /tmp/RtmpC8nsIa/fileb52b7bccd66e/target/extendr_wrappers.R.

f()
#> [1] FALSE
```

<sup>Created on 2021-03-16 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

I guess that's why we are seeing the element named "NA", instead of an unnamed one.

<details>

See #191 for more details

``` r
rextendr::rust_function(r"(
fn call_pairlist(f: Robj) -> Robj {
  let names_and_values = vec![("a", r!(1)), ("b", r!(2)), (na_str(), r!(3))];
  let pairlist = Pairlist{ names_and_values };
  f.call(pairlist.into_robj()).unwrap()
}
)", patch.crates_io = list(`extendr-api` = list(git = "https://github.com/extendr/extendr"))
)
#> build directory: /tmp/Rtmp3Tjm0Q/file1022ad72ffb210
#> Writting wrappers to:
#> /tmp/Rtmp3Tjm0Q/file1022ad72ffb210/target/extendr_wrappers.R

call_pairlist(function(...) list(...))
#> $a
#> [1] 1
#> 
#> $b
#> [1] 2
#> 
#> $`NA`
#> [1] 3
```

</details>

andy-thomason:
We do have unbound_value() for symbols, but the pairlist usually nulls the
Tag.

The logic is handled by the SymPair trait which is designed to pass through
symbols (they are
expensive to create).

We could make "", na_str() and unbound_value() all leave the tag unset.

I'll get working on a replacement for pairlist! and list! at the earliest.

I have only the atomic vectors Int, Real and Logical to convert to the
standardised wrappers.

andy-thomason:
Under the new system call should take a Pairlist argument as well to make
it clear.

yutannihilation:
Thanks, confirmed it works now! I'm still struggling to see what's the supposed way of handling `""` and `na_str()` (as I don't know much about the internals around here, sorry), but I'm fine with the current implementation for my usage.

``` r
rextendr::rust_function(r"(
fn call_pairlist(f: Robj) -> Robj {
  let pairs = vec![("a", r!(1)), ("b", r!(2)), (na_str(), r!(3))];
  let pairlist = Pairlist::from_pairs(pairs);
  f.call(pairlist.into_robj()).unwrap()
}
)", patch.crates_io = list(`extendr-api` = list(git = "https://github.com/extendr/extendr", branch = "more-wrapper-tidying"))
)
#> build directory: /tmp/RtmpiG5eGz/file1a1041cacbf6b
#> ✓ Writing wrappers to /tmp/RtmpiG5eGz/file1a1041cacbf6b/target/extendr_wrappers.R.

call_pairlist(function(...) list(...))
#> $a
#> [1] 1
#> 
#> $b
#> [1] 2
#> 
#> [[3]]
#> [1] 3
```

<sup>Created on 2021-03-17 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

andy-thomason:
Thanks for trying this.

The conversion of strings to symbols now makes "" and na_str() into
unbound_value().

I've moved all the symbol related functions into the symbol module and I'm
tightening up the type requirements.

We could lose all the "as" methods as try_into() should do the same job
and we are getting closer to replacing FromRobj with try_into() in the
function
wrappers. This should make the error handling a little more consistent.

I still have Int, Real and Logical (Complex?) to do in the same style, but
after
that it should be more uniform.

yutannihilation:
(I guess this isn't the main point of this pull request, so please ignore)

>  The conversion of strings to symbols now makes "" and na_str() into unbound_value().

Yes, I understand the design. But, again, I'm still struggling to figure out if it does the things right. It differs from the behaviour of usual R functions, and exposing `unbound_value()` to users might be a bit too low-level.

``` r
as.name("")
#> Error in as.name(""): attempt to use zero-length variable name
as.name(NA_character_)
#> `NA`
```

<sup>Created on 2021-03-19 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

clauswilke:
> But, again, I'm still struggling to figure out if it does the things right.

How does Rcpp or cpp11 do this? In R itself, you make unnamed arguments by simply not providing a name, but languages such as C++ or Rust may not provide this flexibility, so it's a little more tricky.

For Rust, I assume we could write a macro that figures out whether a name is given or not and behaves accordingly.

andy-thomason:
I'm hoping to update the call! macro so that it will take:
```
call!("func", ...)
call!(func_object, ...)
call!("namespace::func", ...)
```
Ideally, this should become a procedural macro, where more options are possible.

We could also support default arguments for inputs.

```
#[extendr]
fn fred(x: &str, y: i32=1) {
}
```
The syntax is not legal Rust (they are fighting default arguments because it has safety
implications), but we can mutate the token stream.

andy-thomason:
Probably time to wrap up this PR if everyone is happy.

yutannihilation:
@clauswilke 
I'll investigate it and file a new issue.

@andy-thomason 
I'm happy with this PR.

clauswilke:
Same here. Go ahead.

clauswilke:
If you merge in the current master branch then the CI failures should resolve.

PR #200, opened 2021-03-21
Remove unused lazy_static from extendr-macro

yutannihilation:
Currently, lazy_static is used only in extendr-api, so it seems we can remove this line.


clauswilke (approved):
Yes, it was used at some point but not anymore.

PR #201, opened 2021-03-31
Refactor extendr_macros

andy-thomason:
Tidy up the procedural macros.

This splits the `#[extendr]` and `extendr_module!` macros into separate modules for clarity.


clauswilke:
I'm a bit confused. Where is the new macro for the fast pairlist?

And also, is this mostly just moving code around or is it also introducing changes? It's difficult to review. Maybe break it into two PRs where the first only rearranges and the second adds functional changes?

andy-thomason:
I probably should have submitted the refactoring and the new macro as separate PRs.

With this infrastructure in place, we should be able to do list!, lang! and data_frame!.

I'm also hoping to do R! with better token handling like R!(data.frame...) and parameters.

A big question is which characters to pick for a parameter delimiter.

We could do:

R!(print( ${rust_expr()} ))

For example

extendr-macros/src/pairlist.rs
```diff
@@ -0,0 +1,30 @@
+use proc_macro::TokenStream;
```
andy-thomason, line 1: This is the actual pairlist code.

clauswilke:
> A big question is which characters to pick for a parameter delimiter.

Can you do curly curly (`{{`)? That would mirror the rlang approach: 
https://www.tidyverse.org/blog/2019/06/rlang-0-4-0/#a-simpler-interpolation-pattern-with-

Hopefully people won't want to do R pseudoquotation inside Rust pseudoquotation.

extendr-macros/src/pairlist.rs
clauswilke: Ah, Ok. For some reason it didn't show up when I did a search for "pairlist".

andy-thomason:
I've split this PR into two. This is now just the refactor.

clauswilke (approved):

PR #203, opened 2021-04-01
R! as a procedural macro.

andy-thomason:
This PR upgrades the R! macro to take parameters and quoted strings.

The quoted strings are to make the code parseable by R, even with data.frame
and backticks.

The parameters substitute Rust expressions into an environment which is then used
to execute the expression.

Examples:

R!(1) -> 1.0 as before

R!("1") -> 1.0 instead of "1.0"

R!("1 + {{ 1.0 }}") -> 2.0 where the 1.0 is any Rust expression convertable to and R object.

R!(r"
   x <- {{ (0..4).collect_robj() }}
   x
") -> [0, 1, 2, 3]

Later we should be able to bake in the parsing and maybe even compile R expressions into Rust.

At the moment, there is some overhead for creating the parameters, the environment and so on.

### DIFF
--- a/extendr-api/src/functions.rs
+++ b/extendr-api/src/functions.rs
@@ -203,6 +203,7 @@ pub fn parse(code: &str) -> Result<Robj> {
 }
 
 /// Parse a string into an R executable object and run it.
+/// Used by the R! macro.
 /// ```
 /// use extendr_api::prelude::*;
 /// test! {
@@ -223,6 +224,37 @@ pub fn eval_string(code: &str) -> Result<Robj> {
     })
 }
 
+/// Parse a string into an R executable object and run it using
+///   parameters param.0, param.1, ...
+///
+/// Used by the R! macro.
+/// ```
+/// use extendr_api::prelude::*;
+/// test! {
+///    let res = eval_string_with_params("param.0", &[&r!(3.)]).unwrap();
+///    assert_eq!(res, r!(3.));
+/// }
+/// ```
+pub fn eval_string_with_params(code: &str, values: &[&Robj]) -> Result<Robj> {
+    single_threaded(|| {
+        let env = Environment::new(global_env());
+        for (i, &v) in values.into_iter().enumerate() {
+            let key = Symbol::from_string(format!("param.{}", i));
+            env.set_local(key, v);
+        }
+
+        let expr = parse(code)?;
+        let mut res = Robj::from(());
+        if let Some(expr) = expr.as_expression() {
+            for lang in expr.values() {
+                res = lang.eval_with_env(&env)?
+            }
+        }
+
+        Ok(res)
+    })
+}
+
 /// Find a function or primitive that may be in a namespace.
 /// ```
 /// use extendr_api::prelude::*;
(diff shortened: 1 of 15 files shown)


clauswilke:
I'm not sure I like the idea that the macro takes either raw expressions or quoted strings. How do I then create an R string? I think it'd be better to have two separate macros, one that takes raw expressions and one that takes strings, e.g. `R!` and `R_str!`.

andy-thomason:
> How do I then create an R string?

We can use raw strings 

```
R!(r#"
    print("hello")
"#)
```

I want to discourage people from using the unquoted variety as they are a support nightmare.
Also, the unquoted form does not support parameters.

`R!(x <- {{expr()}})` does not work.
`R!("x <- {{expr()}}")` does work.

I would like to drop the unquoted form, but will implement both for now if that is desirable.

I'll keep the quoted version short, say Rp!("") for "parameterised".

clauswilke:
I'm Ok with dropping the unquoted form altogether (at least in the long run). Is there a way to trigger a compiler warning when it is encountered, so people could be alerted to the fact that what they're doing is deprecated and no longer recommended?

clauswilke:
> I'm Ok with dropping the unquoted form altogether (at least in the long run).

However, then I think all the examples should be changed accordingly.

andy-thomason:
It seems that we are waiting on a new feature to stabilise for warnings.

https://github.com/rust-lang/rust/issues/54140

I'd be happy if we just change all the R! documentation to R!("") or R!(r#""#)

andy-thomason:
The problem is that the token stream we are given is opaque and does not contain the original text.

This is why we can't tell data.frame from data . frame

clauswilke:
> I'd be happy if we just change all the `R!` documentation to `R!("")` or `R!(r#""#)`

Works for me. Probably there should also be a statement in the changelog that unquoted `R!()` is discouraged and may be disabled in the future.

> The problem is that the token stream we are given is opaque and does not contain the original text. This is why we can't tell `data.frame` from `data . frame`

Yes, I had realized that as well, and I think this will probably never change. So better to just work with strings always. On the flip side, in rextendr, we also take Rust code in a string.

andy-thomason:
> On the flip side, in rextendr, we also take Rust code in a string.

Important to have respect for each language :)

andy-thomason:
Way past our bed time here. Have a happy remainder of your Easter Sunday!

I'll merge this tomorrow if there are no objections before then.

Andy.

yutannihilation (approved):
Sorry that I couldn't review this (and #202) in time... I'm still at the stage of learning how to write procedural macro with [a tutorial for beginners](https://github.com/dtolnay/proc-macro-workshop) so I'm not sure if I can fully review this, but, as far as playing around with this a bit, this looks good to me.

Just curious, can we use the actual `{{` in the R expression?

``` rust
// c.f. https://dplyr.tidyverse.org/articles/programming.html
R!(r#"
var_summary <- function(data, var) {
  data %>%
    summarise(n = n(), min = min({{ var }}), max = max({{ var }}))
}
"#)
```

I don't want to complicate the things so I think it's fine not to support it, at least at the moment.

clauswilke:
> Just curious, can we use the actual `{{` in the R expression?

I think this is why we want a parameterized and a non-parameterized version of the macro. We could use the R `{{` in the non-parameterized version.

yutannihilation:
Ah, makes some sense, thanks.

andy-thomason:
Thanks for the feedback.

I've added a Rraw! variant to cope with Yuta's example.

PR #205, opened 2021-04-10
Fix more clippy warnings.

andy-thomason:
Fix remaining clippy warnings.

I've now added `cargo clippy` to my pre-commit hook and I suggest that others do the same.

I didn't agree with one of the clippy checks and put an #[allow()] block in.


Ilia-Kosenkov (approved):
@andy-thomason, looks good to me, but I was not following extendr recently, so my knowledge of this topic is outdated.

andy-thomason:
I thought that I should include you in the conversation as you have been working hard.

Issue #211, opened 2021-04-21
CI failure on master


yutannihilation:
https://github.com/extendr/extendr/runs/2360651290#step:10:122

```
  stderr:
  Error: Error: cannot add binding of 'myvar' to the base environment
  Execution halted
  
  test src/rmacros.rs - var (line 40) ... FAILED
```

malcolmbarrett:
This may have to do with recent changes in R devel to lock the base env https://github.com/wch/r-source/commit/26165fd368d3f55f4535e5b6d0628e9dfa9caa17

yutannihilation:
Ah, thanks! I remember I saw it on R-devel's NEWS. Then we can just fix this test to use a newly created empty env. 

https://github.com/extendr/extendr/blob/7b8112ed0aed8d5e59bf1e2c04aeb82483db51b0/extendr-api/src/rmacros.rs#L44-L45

But, I also feel base env is not what we want for `current_env()` here. I'm not immediately sure how this should be fixed...

malcolmbarrett:
rlang's definition of `current_env()` is just `parent.frame()`, so maybe that's a better option

yutannihilation:
I think it makes sense when it's used with R code, so it's probably different in the case of Rust (or C/C++).

yutannihilation:
What `current_env()` does is basically `R_GlobalContext->sysparent`.

https://github.com/extendr/extendr/blob/7b8112ed0aed8d5e59bf1e2c04aeb82483db51b0/extendr-api/src/functions.rs#L84-L86

> ```
> /* Return the current environment. */
> SEXP R_GetCurrentEnv() {
>     return R_GlobalContext->sysparent;
> }
> ```
https://github.com/wch/r-source/blob/79298c499218846d14500255efd622b5021c10ec/src/main/context.c#L795-L798

According to the doc,

> ```
>  *	sysparent	the environment the closure was called from
> ```
https://github.com/wch/r-source/blob/79298c499218846d14500255efd622b5021c10ec/src/main/context.c#L46

so this is probably intended to be called from another frame, but probably it's not the case when it's used in Rust code?

andy-thomason:
`current_env()` will be set by R when calling a function.

This means that `var!` will access variables in your calling function first.

Unfortunately, it is not easy to test this outside of R as there is no way I know to create a calling
context frame.

andy-thomason:
I have been thinking of how we can wrap a closure in an R object without registering a function...

This may or may not be possible.

[...]

PR #213, opened 2021-05-18
Fix for builds on arm32, arm64, ppc, etc

nickbp:
Fixes this build error on platforms where `c_char` is defined as u8:
```
   --> /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/extendr-api-0.2.0/src/wrapper.rs:368:25
    |
368 |     unsafe { Rf_install(bytes.as_ptr() as *const i8) }
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `u8`, found `i8`
    |
    = note: expected raw pointer `*const u8`
               found raw pointer `*const i8`
```

The underlying `Rf_install` binding defines this as a C char, see e.g.: https://github.com/extendr/libR-sys/search?q=rf_install
That C char type can either be an `i8` or a `u8`, depending on platform: https://doc.rust-lang.org/src/std/os/raw/mod.rs.html#48-90

### DIFF
--- a/extendr-api/src/wrapper/mod.rs
+++ b/extendr-api/src/wrapper/mod.rs
@@ -36,7 +36,7 @@ pub(crate) fn make_symbol(name: &str) -> SEXP {
     let mut bytes = Vec::with_capacity(name.len() + 1);
     bytes.extend(name.bytes());
     bytes.push(0);
-    unsafe { Rf_install(bytes.as_ptr() as *const i8) }
+    unsafe { Rf_install(bytes.as_ptr() as *const ::std::os::raw::c_char) }
 }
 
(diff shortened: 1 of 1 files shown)


andy-thomason:
Thanks for this @nickbp 

I should have written this more formally in the first place.

andy-thomason:
Side note:

Do we still have that "myvar" test in there? It should have gone long ago.
We can't test the base_env() using just libR.

clauswilke (approved):

clauswilke:
Yes, the `myvar` test needs to go, but that has nothing to do with this PR. See #211.

nickbp:
Good to merge? I'm not able to merge myself

PR #218, opened 2021-06-10
Remove MSYS2 path on GitHub Actions CI

yutannihilation:
As https://github.com/extendr/rextendr/issues/91 did commited on rextendr, probably we can do the same thing on extendr as well?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -91,20 +91,17 @@ jobs:
         if: startsWith(runner.os, 'Windows')
         # 1. Add rust target
         # 2. Add target name to the $targets variable
-        # 3. Add mingw32/mingw64 bin folders to PATH
-        # 4. Add R x64/i386 folders to PATH
+        # 3. Add R x64/i386 folders to PATH
         run: |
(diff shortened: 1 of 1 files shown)


Ilia-Kosenkov:
@yutannihilation, you cannot throw away `mingw` dependency completely. You may try to replace `msys2` with Rtools on the `PATH`, but I am not 100% sure all extendr tests can be executed without `msys2`.

Right now the CI fails because there are no `mingw` compilers on the path.

yutannihilation:
I see, thanks. Apparently, we cannot go away from msys2 on w/ bindgen cases...

Ilia-Kosenkov:
I checked locally, it seems that almost all tests rely on bindgen (except for `test-minimal` subset and perhaps some other), and bindgen requires clang, which is not provided by Rtools. The only solution is to use `msys2` for this pruposes.

yutannihilation:
Hmm, okay, I optimistically thought `tests_no_bindgen` cases are without bindgen, as the name suggests... Sorry for taking your time. I give up here.

PR #221, opened 2021-07-12
Update README

yutannihilation:
We now can check these items, right?

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,15 +112,15 @@ accidental segfaults and failures.
 - [x] Begin documentation.
 - [ ] Begin book-form documentation.
 - [ ] Paper for Bioinformatics.
-- [ ] Build and publish CRAN R package.
+- [x] Build and publish CRAN R package.
 - [ ] Publish Use R! series book.
 
 ### Automation
 - [x] Auto-generate binding wrappers.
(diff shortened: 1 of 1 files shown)


andy-thomason:
Yes! Absolutely.

yutannihilation:
Thanks for confirming!

clauswilke:
In the long run, though, as the project matures, I would probably just delete this roadmap from the README altogether. It's always better to underpromise and overdeliver and roadmaps tend to create the opposite effect.

andy-thomason:
It has probably served its purpose.

yutannihilation:
I agree with your concern on showing the roadmap. But I'd add that the roadmap is probably important if you consider getting funded or sponsored (i.e. GitHub Sponsors), at the risk of overpromising.

I think the README needs an overhaul anyway, considering that in-R usages are described in rextendr's document and each crate now has its own README. I'm wondering what information should be here.

andy-thomason:
Yes. Definitely in need of an overhaul.

I've been looking at sponsorship sources as I intend to leave my current job to focus on teaching and open source.
I may return, at least partially, to academia.

We should set up Github sponsors for the project.

yutannihilation:
Wish you great success! :)

Issue #219, opened 2021-06-14
Different conversion rule between Option<T> vs T


yutannihilation:
Was this intentional? Sorry if I forget some discussion on this...

``` r
rextendr::rust_function(r"(
fn f1(x: i32) -> i32 {
    x
}
)", patch.crates_io = list(`extendr-api` = list(git = "https://github.com/extendr/extendr"))
)
#> ℹ build directory: '/tmp/RtmpVy1jIt/file4bcab43332e1e'
#> ✓ Writing '/tmp/RtmpVy1jIt/file4bcab43332e1e/target/extendr_wrappers.R'.

rextendr::rust_function(r"(
fn f2(x: Option<i32>) -> i32 {
    if let Some(x) = x {
        x
    } else {
        1
    }
}
)", patch.crates_io = list(`extendr-api` = list(git = "https://github.com/extendr/extendr"))
)
#> ℹ build directory: '/tmp/RtmpVy1jIt/file4bcab43332e1e'
#> ✓ Writing '/tmp/RtmpVy1jIt/file4bcab43332e1e/target/extendr_wrappers.R'.

# without option, tolerant to numeric
f1(1)
#> [1] 1

# with option, intolerant
f2(1)
#> Error in f2(1): expected an integer scalar
```

<sup>Created on 2021-06-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

clauswilke:
Seems inconsistent to me. I've added this to the 0.3 milestone.

andy-thomason:
I'll take a look when I'm back online.

Ilia-Kosenkov:
There is a certain degree of inconsistency in type conversion and error reporting. Here is an example
```R
rextendr::rust_function("fn test_na(x : i32) -> i32 {x}")
#> i build directory: 'C:\Users\...\AppData\Local\Temp\RtmpaGV9dI\file1c845d623969'
#> v Writing 'C:/Users/.../AppData/Local/Temp/RtmpaGV9dI/file1c845d623969/target/extendr_wrappers.R'.
test_na(NA)
#> Error in test_na(NA): unable to convert R object to primitive
test_na(NA_integer_)
#> Error in test_na(NA_integer_): Input must not be NA.
test_na(NA_real_)
#> Error in test_na(NA_real_): Input must not be NA.
```
The rules for regular types are broad (I personally not a fan of). Yet for `Option<T>` it is much more strict.
[...]

PR #222, opened 2021-07-14
Add option to use try from in extendr macro

andy-thomason:
This PR adds an option:

```
fn test_i32(val: i32) -> i32 {
    val
}
```

This is a prelude to supporting the try_from error handling formally.

It should have no effect on the main branch and so, we may merge this to enable others to contribute
to writing conversions.

We may need the failing conversion to give a useful error message by using to_string() on the returned
error.

We should then write integration tests to check these messages in R.

Note that Option<T> gives None if T is a NA value, not NULL.

There is another wrapper `Nullable` which handles NULLs.

### DIFF
--- a/extendr-api/src/robj/mod.rs
+++ b/extendr-api/src/robj/mod.rs
@@ -1,6 +1,6 @@
 //! R object handling.
 //!
-//! See. https://cran.r-project.org/doc/manuals/R-exts.html
+//! See. <https://cran.r-project.org/doc/manuals/R-exts.html>
 //!
 //! Fundamental principals:
 //!
--- a/extendr-api/tests/extendr_macro.rs
+++ b/extendr-api/tests/extendr_macro.rs
@@ -1,0 +1,109 @@
+use extendr_macros::*;
+use extendr_api::prelude::*;
+
+#[extendr(use_try_from=true)]
+fn test_i32(val: i32) -> i32 {
+    val
+}
+
+#[extendr(use_try_from=true)]
+fn test_i16(val: i16) -> i16 {
+    val
+}
+
+#[extendr(use_try_from=true)]
+fn test_option_i32(val: Option<i32>) -> i32 {
+    if let Some(i) = val {
+        i
+    } else {
+        -1
+    }
+}
+
+#[extendr(use_try_from=true)]
+fn test_option_f64(val: Option<f64>) -> f64 {
+    if let Some(i) = val {
+        i
+    } else {
+        -1.0
+    }
+}
+
+#[extendr(use_try_from=true)]
+fn test_option_i16(val: Option<i16>) -> i16 {
+    if let Some(i) = val {
+        i
+    } else {
+        -1
+    }
+}
+
+#[test]
+fn happy_path() {
+    unsafe {
+        test! {
+            // Matching integer.
+            assert_eq!(new_owned(wrap__test_i32(r!(1).get())), r!(1));
+
+            // i32 takes any numeric.
+            assert_eq!(new_owned(wrap__test_i32(r!(1.0).get())), r!(1));
+
+
+            // Matching integer.
+            assert_eq!(new_owned(wrap__test_option_i32(r!(1).get())), r!(1));
+
+            // Option<i32> takes any numeric.
+            assert_eq!(new_owned(wrap__test_option_i32(r!(1.0).get())), r!(1));
+
+            // NA input.
+            assert_eq!(new_owned(wrap__test_option_i32(r!(NA_REAL).get())), r!(-1));
+
+            // NA input.
+            assert_eq!(new_owned(wrap__test_option_i32(r!(NA_INTEGER).get())), r!(-1));
+
+
+            // Matching integer.
+            assert_eq!(new_owned(wrap__test_option_i16(r!(1).get())), r!(1));
+
+            // Option<i16> takes any numeric.
+            assert_eq!(new_owned(wrap__test_option_i16(r!(1.0).get())), r!(1));
+
+            // NA input.
+            assert_eq!(new_owned(wrap__test_option_i16(r!(NA_REAL).get())), r!(-1));
+
+            // NA input.
+            assert_eq!(new_owned(wrap__test_option_i16(r!(NA_INTEGER).get())), r!(-1));
+
+
+            // Matching integer.
+            assert_eq!(new_owned(wrap__test_option_f64(r!(1).get())), r!(1.0));
+
+            // Option<f64> takes any numeric.
+            assert_eq!(new_owned(wrap__test_option_f64(r!(1.0).get())), r!(1.0));
+
+            // NA input.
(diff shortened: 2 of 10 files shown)


extendr-api/tests/extendr_macro.rs
```diff
@@ -0,0 +1,109 @@
...
+fn test_option_i16(val: Option<i16>) -> i16 {
+    if let Some(i) = val {
+        i
+    } else {
+        -1
+    }
+}
+
+#[test]
+fn happy_path() {
```
clauswilke, line 42: Could you use a more descriptive function name? E.g. `tests_without_error()`?

extendr-api/src/robj/mod.rs
```diff
@@ -1,6 +1,6 @@
 //! R object handling.
 //!
-//! See. https://cran.r-project.org/doc/manuals/R-exts.html
+//! See. <https://cran.r-project.org/doc/manuals/R-exts.html>
 //!
 //! Fundamental principals:
```
clauswilke, line 5: Do you mean "principles"?

extendr-api/tests/extendr_macro.rs
```diff
@@ -0,0 +1,109 @@
...
+            assert_eq!(new_owned(wrap__test_option_f64(r!(NA_REAL).get())), r!(-1.0));
+
+            // NA input.
+            assert_eq!(new_owned(wrap__test_option_f64(r!(NA_INTEGER).get())), r!(-1.0));
+        }
+    }
+}
+
+// Win32 does not support catch_unwind.
+#[cfg(not(target_arch = "x86"))]
```
clauswilke, line 94: Would it make sense to make this test Windows specific? Or is `x86` specific to Windows?

extendr-api/tests/extendr_macro.rs
```diff
@@ -0,0 +1,109 @@
...
+#[test]
+fn sad_path() {
```
clauswilke, line 96: See above. E.g., `tests_with_errors()`.

clauswilke (approved):
Generally looks fine to me. I had a few minor comments.

extendr-api/tests/extendr_macro.rs
```diff
@@ -0,0 +1,109 @@
...
+
+#[extendr(use_try_from = true)]
+fn test_option_i16(val: Option<i16>) -> i16 {
+    if let Some(i) = val {
+        i
+    } else {
+        -1
+    }
+}
+
```
brendanrbrown, line 40: a `test_option_bool` as well?

extendr-api/tests/extendr_macro.rs
```diff
@@ -0,0 +1,109 @@
...
+            // Option<f64> takes any numeric.
+            assert_eq!(new_owned(wrap__test_option_f64(r!(1.0).get())), r!(1.0));
+
+            // NA input.
+            assert_eq!(new_owned(wrap__test_option_f64(r!(NA_REAL).get())), r!(-1.0));
+
+            // NA input.
+            assert_eq!(new_owned(wrap__test_option_f64(r!(NA_INTEGER).get())), r!(-1.0));
+        }
+    }
```
brendanrbrown, line 90: if adding `test_option_bool` above: a series for logical

brendanrbrown:
looks nice -- i've commented on one test that snuck away this time, but adding it could be work for others as you mentioned in #219.

extendr-api/tests/extendr_macro.rs
andy-thomason: I'll merge this into main. Feel free to add more tests.

extendr-api/tests/extendr_macro.rs
andy-thomason: I think that the problem (catch_unwind not supported) is x86 specific, rather than windows,
but Windows (XP) is currently our only x86 target.

Getting the tests to run green is a priority here, we could refine later.

extendr-api/tests/extendr_macro.rs
```diff
@@ -0,0 +1,109 @@
...
+
+#[test]
+fn happy_path() {
+    unsafe {
+        test! {
+            // Matching integer.
+            assert_eq!(new_owned(wrap__test_i32(r!(1).get())), r!(1));
+
+            // i32 takes any numeric.
+            assert_eq!(new_owned(wrap__test_i32(r!(1.0).get())), r!(1));
```
Ilia-Kosenkov, line 49: I am not sure what 'takes any numeric' means here. What will happen if we pass `1.5`?
What would be the behavior?
If the value is somehow truncated/converted to integer space, this may be an issue, and extremely hard to debug.

Same applies to `Option<i32>`

If this is OK, may I then suggest to add another option, like `#[extendr(use_strict_types = true)]` or `#[extendr(vctrs_compat = true)]` to enable strict type checks?

andy-thomason: If you feel like doing so, we should replace the "as" in `impl_try_from_scalar!` with TryFrom.
This would also range-check small numbers.

I'm going to close this ASAP so that more PRs can follow.

extendr-api/tests/extendr_macro.rs
```diff
@@ -0,0 +1,109 @@
...
+
+            // NA input.
+            assert_eq!(new_owned(wrap__test_option_i32(r!(NA_REAL).get())), r!(-1));
+
+            // NA input.
+            assert_eq!(new_owned(wrap__test_option_i32(r!(NA_INTEGER).get())), r!(-1));
+
+
+            // Matching integer.
+            assert_eq!(new_owned(wrap__test_option_i16(r!(1).get())), r!(1));
```
Ilia-Kosenkov, line 66: Without having a deep understanding of what `#[extendr(use_try_from = true)]` produces, I would suggest adding tests for int values outside of `i16` value space.

extendr-api/tests/extendr_macro.rs
andy-thomason: We now have a template for adding other options, so suggestions like this are welcome.

extendr-api/tests/extendr_macro.rs
andy-thomason: See the comment above about replacing the "as".

Ilia-Kosenkov (approved):
I like the new approach using options in `#[extendr]`, this should speed up feature development.

extendr-api/tests/extendr_macro.rs
clauswilke: Just wanted to point out that I think strict type checks should be the default, not optional. If somebody writes `i32` as their input type we shouldn't silently round from 1.5 to 1, we should throw a runtime error.

Ilia-Kosenkov: I fully support this. My preferred solution would be something similar to {vctrs} behavior, which should at least be available as an option.

PR #225, opened 2021-07-17
Get `SymPair` to return `Option<Symbol>`

### LINKED ISSUE #195: Does unbound_value need to be exposed just to support pairlist creation from string tuples?
(second round of #187)

### Background

When we call a function, there's two types of cases:

1. use a fixed number of arguments that is determined on compile
2. use an arbitrary number of arguments that is determined on runtime

For case 1, we'll able to write as the following after #190.

``` r
f.call(pairlist!(a = r!(1), r(2)))
```

We can provide more sophisticated interface by using procedural macro. However, case 2 is a different problem. As the structure of the argument is determined on runtime, the macro cannot solve the problem, iiuc. If you are curious about the real use case, please refer to my experimental package: https://github.com/yutannihilation/rr4r#rr4r_replace

The problem is how to handle named and unnamed arguments transparently. The implementation in #190 solves this by treating `""` and `na_str()` as `R_UnboundValue` so that we can write:

``` rust
let pairs = vec![("a", r!(1)), (na_str(), r!(2))];
let pairlist = Pairlist::from_pairs(pairs);
f.call(pairlist.into_robj()).unwrap()
```

### Problem

While the current approach works, I'm afraid it makes the handling of symbols different from what it should be; It differs from the behaviour of usual R functions,

``` r
as.name("")
#> Error in as.name(""): attempt to use zero-length variable name
as.name(NA_character_)
#> `NA`
```

<sup>Created on 2021-03-19 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

and exposing `unbound_value()` to users might be a bit too low-level.

```
> rextendr::rust_function(r"(
+ fn as_sym(x: String) -> Robj {
+   Symbol::from_str(x.as_str()).into_robj()
+ }
+ )", patch.crates_io = list(`extendr-api` = list(git = "https://github.com/extendr/extendr", branch = "more-wrapper-tidying"))
+ )
build directory: /tmp/RtmpafHjhM/file13fe8440d442b

    Updating crates.io index
   Compiling rextendr9 v0.0.1 (/tmp/RtmpafHjhM/file13fe8440d442b)
    Finished dev [unoptimized + debuginfo] target(s) in 1.69s
✓ Writing wrappers to /tmp/RtmpafHjhM/file13fe8440d442b/target/extendr_wrappers.R.
> x <- as_sym("")
Error: attempt to bind a variable to R_UnboundValue
> as_sym("")
Error: invalid multibyte string at '<90><c0><a8>y<f4>U'
```

Can we let the construction of a pairlist just work without tweaking the symbols' interface?

### How do cpp11/Rcpp handle pairlists?

For case 1, cpp11 uses variadic template and polymorphism to treat named and unnamed argument differently. I guess the same thing can be done by procedural macro. 

* operator `()`: https://github.com/r-lib/cpp11/blob/3c65bc56250a882613ed919d271b789aad557402/inst/include/cpp11/function.hpp#L20-L54
* `named_arg`: https://github.com/r-lib/cpp11/blob/c41056b557cabd30bb2e5b901ca92a26262e7718/inst/include/cpp11/named_arg.hpp#L11-L38

For case 2, I couldn't find the interface (maybe due to my lack of C++ skill). Maybe users need to create an unnamed list and then `set_names()`? This also looks fine to me.

yutannihilation:
Fix #195 

This pull request is to remove the necessity to use unbound value to represent the absence of the name. I decided to follow [Andy's comment](https://github.com/extendr/extendr/issues/195#issuecomment-817832510). Instead, users can use `""` (empty string).

This is consistent with the behavior of R (c.f. https://github.com/extendr/extendr/issues/195#issuecomment-803497556, https://github.com/extendr/extendr/issues/195#issuecomment-803498672)

While this is a breaking change in that `SymPair::sym_pair()` now returns `(Option<Robj>, Robj)` instead of `(Robj, Robj)`, I expect this doesn't affect the existing users much as this is rather an expert-only feature at the moment.

### DIFF
--- a/extendr-api/src/wrapper/environment.rs
+++ b/extendr-api/src/wrapper/environment.rs
@@ -65,7 +65,9 @@ impl Environment {
             let robj = call!("new.env", TRUE, parent, dict_len).unwrap();
             for nv in names_and_values {
                 let (n, v) = nv.sym_pair();
-                unsafe { Rf_defineVar(n.get(), v.get(), robj.get()) }
+                if let Some(n) = n {
+                    unsafe { Rf_defineVar(n.get(), v.get(), robj.get()) }
+                }
             }
             Environment { robj }
         })
--- a/extendr-api/src/wrapper/symbol.rs
+++ b/extendr-api/src/wrapper/symbol.rs
@@ -18,25 +18,17 @@ pub struct Symbol {
 impl Symbol {
     /// Make a symbol object from a string.
     ///
-    /// The values na_str() and "" give unbound_value().
-    ///
     /// ```
     /// use extendr_api::prelude::*;
     /// test! {
     ///     let chr = r!(Symbol::from_string("xyz"));
     ///     assert_eq!(chr, sym!(xyz));
-    ///     assert_eq!(Symbol::from_string(na_str()).is_unbound_value(), true);
-    ///     assert_eq!(Symbol::from_string("").is_unbound_value(), true);
     /// }
     /// ```
     pub fn from_string<S: AsRef<str>>(val: S) -> Self {
         let val = val.as_ref();
-        if val.is_empty() || val.is_na() {
-            unbound_value()
-        } else {
-            Symbol {
-                robj: unsafe { new_owned(make_symbol(val)) },
-            }
+        Symbol {
+            robj: unsafe { new_owned(make_symbol(val)) },
         }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 
 - Use Use failure() to trigger steps on failures
 
+- `SymPair::sym_pair()` now returns `(Option<Robj>, Robj)`
 
 ## extendr 0.2.0
 
--- a/extendr-api/src/wrapper/mod.rs
+++ b/extendr-api/src/wrapper/mod.rs
@@ -200,7 +200,7 @@ impl Robj {
     /// ```
     /// use extendr_api::prelude::*;
     /// test! {
-    ///     let names_and_values = vec![("a", r!(1)), ("b", r!(2)), (na_str(), r!(3))];
+    ///     let names_and_values = vec![("a", r!(1)), ("b", r!(2)), ("", r!(3))];
     ///     let pairlist = Pairlist::from_pairs(names_and_values);
     ///     let robj = r!(pairlist.clone());
     ///     assert_eq!(robj.as_pairlist().unwrap(), pairlist);
@@ -272,7 +272,7 @@ impl Robj {
 }
 
 pub trait SymPair {
-    fn sym_pair(self) -> (Robj, Robj);
+    fn sym_pair(self) -> (Option<Robj>, Robj);
 }
 
 impl<S, R> SymPair for (S, R)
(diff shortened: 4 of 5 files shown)


extendr-api/src/wrapper/environment.rs
```diff
@@ -65,7 +65,9 @@ impl Environment {
             let robj = call!("new.env", TRUE, parent, dict_len).unwrap();
             for nv in names_and_values {
                 let (n, v) = nv.sym_pair();
-                unsafe { Rf_defineVar(n.get(), v.get(), robj.get()) }
+                if let Some(n) = n {
+                    unsafe { Rf_defineVar(n.get(), v.get(), robj.get()) }
+                }
```
yutannihilation, line 70: One thing not very obvious is what we should do when the user try to create an environment from a pairlist containing unnamed element. Currently it's just ignored. Should we raise an error?

clauswilke: What is ignored? The unnamed element only, or the entire call? And if you triggered an error, would it be compile time or runtime?

extendr-api/src/wrapper/symbol.rs
```diff
@@ -18,25 +18,17 @@ pub struct Symbol {
...
     /// ```
     pub fn from_string<S: AsRef<str>>(val: S) -> Self {
         let val = val.as_ref();
-        if val.is_empty() || val.is_na() {
-            unbound_value()
-        } else {
-            Symbol {
-                robj: unsafe { new_owned(make_symbol(val)) },
-            }
+        Symbol {
```
clauswilke, line 30: Just so I understand: What happens now if `val` is empty or is `NA`? Why don't we need to check for this anymore?

clauswilke:
I have reviewed this but don't feel qualified to assess whether the change from `(Robj, Robj)` to `(Option<Robj>, Robj)` will cause any issues. @andy-thomason Could you review also?

extendr-api/src/wrapper/environment.rs
yutannihilation: > The unnamed element only, or the entire call?

Currently this ignores the unnamed elements only.

> And if you triggered an error, would it be compile time or runtime?

Runtime.

extendr-api/src/wrapper/symbol.rs
yutannihilation: They are just passed to R's C API. So, these two should behave in the same way as how R does.

```
as.name("")
#> Error in as.name(""): attempt to use zero-length variable name
as.name(NA_character_)
#> `NA`
```

extendr-api/src/wrapper/environment.rs
clauswilke: It's unclear to me. Usually, my philosophy is that code should use meaningful fallback behavior on empty input, so that would argue against raising an error. I have encountered so many cases where code would complain about `NA` or `NULL` input instead of silently ignoring it, and then I have to special-case these situations, and the code I'm calling is also special-casing, and it just doesn't make sense.

I assume the counter-argument is that silently ignoring invalid input can create difficult-to-find bugs.

yutannihilation: I'm getting inclined to raise an error, rather than handling the invalid inputs gracefully. I think this operation is for special cases when there's a lot of inputs generated programmatically, as the example shows. For more dynamic inputs, users can process the elements one by one with `Environment::set_local()` by themselves.

``` rust
    ///     let names_and_values = (0..100).map(|i| (format!("n{}", i), i));
    ///     let mut env = Environment::from_pairs(global_env(), names_and_values);
    ///     assert_eq!(env.len(), 100);
```

clauswilke: Ok, sounds good to me.

andy-thomason (approved):
It should be a bit faster if anything.

It is also a bit clearer when names are not present.

PR #226, opened 2021-07-17
[WIP] Check length of input in <&str>::try_from()

### LINKED ISSUE #102: Conversion of string vector to string scalar fails with "not a string object"
If a Rust function expecting a single string is handed a string vector from R, then it raises an error stating "not a string object". See here:
https://github.com/extendr/extendr/blob/9752c7f73c109bef321e17d2cdba3f332153d7b7/tests/extendrtests/tests/testthat/test-type-conversion.R#L29

The error is appropriate, but the error message is confusing. A better error would be "Input must be of length 1", as is used for the other types, see e.g. here:
https://github.com/extendr/extendr/blob/9752c7f73c109bef321e17d2cdba3f332153d7b7/tests/extendrtests/tests/testthat/test-type-conversion.R#L7

yutannihilation:
Fix #102

While other `try_from()` implementations use the following logic, `&str` is a bit tricky in that it fails with `cannot return value referencing local variable` error, 

``` rust
        if let Some(v) = robj.as_<type>_slice() {
            match v.len() {
                0 => Err(Error::ExpectedNonZeroLength(robj)),
                1 => Ok(v[0]),
                _ => Err(Error::ExpectedScalar(robj)),
            }
        } else {
            Err(Error::ExpectedLogical(robj))
        }
```

so I chose to check the length before trying to convert the input to the actual variable.

### DIFF
--- a/extendr-api/src/robj/try_from_robj.rs
+++ b/extendr-api/src/robj/try_from_robj.rs
@@ -84,11 +84,18 @@ impl TryFrom<Robj> for &str {
 
     fn try_from(robj: Robj) -> Result<Self> {
         if robj.is_na() {
-            Err(Error::MustNotBeNA(robj))
-        } else if let Some(s) = robj.as_str() {
-            Ok(s)
-        } else {
-            Err(Error::ExpectedString(robj))
+            return Err(Error::MustNotBeNA(robj));
+        }
+        match robj.len() {
+            0 => Err(Error::ExpectedNonZeroLength(robj)),
+            1 => {
+                if let Some(s) = robj.as_str() {
+                    Ok(s)
+                } else {
+                    Err(Error::ExpectedString(robj))
+                }
+            }
+            _ => Err(Error::ExpectedScalar(robj)),
         }
     }
 }


clauswilke (approved):
Looks good to me. 

However, shouldn't this require changes in the integration tests, here:
https://github.com/extendr/extendr/blob/2103979ab085376194b1f51a5a20b28e2f165691/tests/extendrtests/tests/testthat/test-type-conversion.R#L29

yutannihilation:
In my understanding, `try_from()` isn't used yet by `#[extendr]` macro. It still uses `from_robj()` (c.f. https://github.com/extendr/extendr/issues/219#issuecomment-864336345). I'm wondering if I should modify `from_robj()` as well at the moment, considering it is still used but will be replaced...

clauswilke:
Ah, Ok. We probably need a complete implementation of `try_from()` for all types before we can switch over. It's not quite clear to me whether we're there yet or not.

yutannihilation:
> In my understanding, `try_from()` isn't used yet by `#[extendr]` macro.

Sorry, I was a bit confused. We can confirm the behavior with `#[extendr(use_try_from = true)]`.

``` r
gen_fun <- function(fun_name, use_try_from) {
  use_try_from <- if(isTRUE(use_try_from)) "true" else "false"
  code <- r"(
#[extendr(use_try_from = {use_try_from})]
fn {fun_name}(x: &str) {{
    println!("{{}}", x)
}}
)"

  rextendr::rust_source(
    code = glue::glue(code),
    patch.crates_io = list(`extendr-api` = list(git = "https://github.com/yutannihilation/extendr", branch = "fix/issue-102-improve-string-scalar-error")),
    env = globalenv()
  )
}

# Use try_from impl
gen_fun("f1", TRUE)
#> ℹ build directory: '/tmp/Rtmp9mx0FQ/file2b2fb60e771c1'
#> ✓ Writing '/tmp/Rtmp9mx0FQ/file2b2fb60e771c1/target/extendr_wrappers.R'.
# Use from_robj impl
gen_fun("f2", FALSE)
#> ℹ build directory: '/tmp/Rtmp9mx0FQ/file2b2fb60e771c1'
#> ✓ Writing '/tmp/Rtmp9mx0FQ/file2b2fb60e771c1/target/extendr_wrappers.R'.

f1(letters)
#> Error in f1(letters): Expected Scalar, got String
f2(letters)
#> Error in f2(letters): Not a string object.
```

<sup>Created on 2021-07-18 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

andy-thomason (approved):

Issue #224, opened 2021-07-15
Integration tests for TryFrom conversions.


andy-thomason:
We should update the integration tests to cover the full spectrum of types.

If necessary add more tests than we have conversions for, to act as incentive to add more conversions.

We may also want to check that ALTREP types are converted correctly.

See https://github.com/extendr/extendr/tree/master/tests/extendrtests

clauswilke:
To provide a better pointer to where in the code this lives.

The Rust test functions are defined here:
https://github.com/extendr/extendr/blob/2103979ab085376194b1f51a5a20b28e2f165691/tests/extendrtests/src/rust/src/lib.rs#L19-L49

And tests from R are then called here:
https://github.com/extendr/extendr/blob/2103979ab085376194b1f51a5a20b28e2f165691/tests/extendrtests/tests/testthat/test-type-conversion.R#L1-L37

I wrote this set of tests mostly to demonstrate how writing tests works. They are clearly not exhaustive.

@brendanrbrown This might be a good issue for a first PR.

brendanrbrown:
just to confirm that i'll make a PR for this --- shouldn't take too long.

andy-thomason:
I'm going to migrate more tests to the tests directory in time.

Feel free to make more source files.

Just a small PR would be great for a first one, we can expand on this.

PR #227, opened 2021-07-17
Draft: Testing try from

brendanrbrown:
A start on the additional integration tests, for from and tryfrom via the new extendr macro option. Addresses #224. 

Posting here for making / soliciting comments on a few things.

1. To highlight certain errors that lead to compilation failure, there are placeholder rust functions with the forced panic `todo!` which then fail in R `testthat` runs. Where available I include a brief indicator of what the compiler error message was.

2. Some failures are very simple: e.g. the error message for double vector conversions says 'Integer' instead of 'Real'. In other cases, a 'String' conversion is implemented but '&str' is not, or vice versa. These are noted in various places in [`tests/extendrtests/tests/testthat/test-type-conversion.R`](https://github.com/brendanrbrown/extendr/blob/testing-try-from/tests/extendrtests/tests/testthat/test-type-conversion.R). These are not addressed here, but I could go ahead and create a separate issue / PR for these as they are mainly typographical.

3. The extendr macro used with the new try_from option, as `#[extendr(use_try_from = true)]`, seems still to require a `From<Robj>` implementation, as evidenced by the compiler error snippets given in the [`try_bool_vec`](https://github.com/brendanrbrown/extendr/blob/e763c360c9eede19b9e7e69520c2ccef2fcea629/tests/extendrtests/src/rust/src/lib.rs#L169) and [`try_list_string_hash`](https://github.com/brendanrbrown/extendr/blob/e763c360c9eede19b9e7e69520c2ccef2fcea629/tests/extendrtests/src/rust/src/lib.rs#L186) testing functions.

4. I did not yet attempt to integrate the new try_from tests for Option types in [`extendr-api/tests/extendr_macro.rs](https://github.com/extendr/extendr/blob/master/extendr-api/tests/extendr_macro.rs) into this PR, but I would be happy to do so if appropriate.


clauswilke:
A couple of comments:

Point 1: To merge into the main code line, I'd want all tests to pass. If some functionality is still missing I'd comment out the corresponding test.

Point 2: Why don't you make a separate PR for these minor issues and we merge that first?

Point 4: I'd put that into a separate PR. Smaller PRs are more manageable.

And also, I don't think we need all these `.Rd` files. Could you add the line:
```
  /// @usage NULL
```
to all Rust functions exported to R? I believe this will suppress the generation of documentation files.

brendanrbrown:
Thanks @clauswilke for these clarifications. Given these preferences, I will close this PR, stash/delete this branch and incorporate the changes into smaller PRs.

PR #228, opened 2021-07-18
Use R_NewEnv

yutannihilation:
`R_NewEnv` is available as of R 4.1, and it's already exported in the dev version of libR-sys. So, the implementation using this is very straight forward. But, I have no idea how to detect the user's R version, either on compilation time or on runtime, in order to decide whether to use `R_NewEnv` or to call `new.env()` with `call!` macro.


clauswilke:
Info about the R version is available in the bindings:
https://github.com/extendr/libR-sys/blob/7418fade4bca9723ac1c1382c5d570a0fdb083c8/bindings/bindings-linux-x86_64-R4.0.rs#L285-L293

You could do a simple test of whether `R_VERSION` exceeds some numeric value. We just have to figure out what the appropriate value would be.

yutannihilation:
Thanks, but I think we need to switch on compilation time, where we cannot refer to these variables, iiuc. Fortunately, it seems we can use `DEP_R_R_VERSION_*` variables.

c.f. https://doc.rust-lang.org/cargo/reference/build-script-examples.html#conditional-compilation

yutannihilation:
While I think we should wait until the next version of libR-sys gets published, this pull request should be ready to review now. Feedback is welcome.

clauswilke:
Do we need to publish a new version of libR-sys? Should we publish one whenever a major version of R gets released?

yutannihilation:
> Do we need to publish a new version of libR-sys?

I think so.

> Should we publish one whenever a major version of R gets released?

I thought we should wait until the new API is really needed, but now I come to feel one publish per major release is a good idea.

andy-thomason (approved):
Very nice.

I've never been happy with calling new.env

We might have been able to roll it from scratch, but R_NewEnv is the correct way to go.

andy-thomason (approved):

andy-thomason:
This should run green with the PRSEEN fix in place.

PR #236, opened 2021-07-22
Remove SET_PRSEEN

yutannihilation:


andy-thomason:
I'm guessing that PRSEEN is a version specific field.

andy-thomason (approved):

PR #238, opened 2021-07-22
Calling r functions in extension packages

andy-thomason:
Allow more flexible expressions in call! macros and update the Function
wrapper to accept primitives.

### DIFF
--- a/extendr-api/tests/call_tests.rs
+++ b/extendr-api/tests/call_tests.rs
@@ -5,13 +5,13 @@ fn formula_test() {
     test! {
         // As one R! macro call
         let confint1 = R!(confint(lm(weight ~ group - 1, PlantGrowth)))?;
-       
+
         // As many parameterized calls.
         let formula = lang!("~", sym!(weight), lang!("-", sym!(group), 1.0)).set_class(["formula"])?;
         let plant_growth = global!(PlantGrowth)?;
         let model = call!("lm", formula, plant_growth)?;
         let confint2 = call!("confint", model)?;
-       
+
         assert_eq!(confint1.as_real_vector(), confint2.as_real_vector());
     }
 }
--- a/extendr-api/tests/call_tests.rs
+++ b/extendr-api/tests/call_tests.rs
@@ -1,0 +1,27 @@
+use extendr_api::prelude::*;
+
+#[test]
+fn formula_test() {
+    test! {
+        // As one R! macro call
+        let confint1 = R!(confint(lm(weight ~ group - 1, PlantGrowth)))?;
+       
+        // As many parameterized calls.
+        let formula = lang!("~", sym!(weight), lang!("-", sym!(group), 1.0)).set_class(["formula"])?;
+        let plant_growth = global!(PlantGrowth)?;
+        let model = call!("lm", formula, plant_growth)?;
+        let confint2 = call!("confint", model)?;
+       
+        assert_eq!(confint1.as_real_vector(), confint2.as_real_vector());
+    }
+}
+
+#[test]
+fn special_test() {
+    test! {
+        assert_eq!(call!("`+`", 1, 2)?, r!(3));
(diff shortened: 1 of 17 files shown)


extendr-api/tests/call_tests.rs
```diff
@@ -0,0 +1,27 @@
+use extendr_api::prelude::*;
+
+#[test]
+fn formula_test() {
+    test! {
+        // As one R! macro call
+        let confint1 = R!(confint(lm(weight ~ group - 1, PlantGrowth)))?;
```
clauswilke, line 7: I think it would be best to put the argument to `R!()` into a string. Let's get into the habit of always doing that.

extendr-macros/src/lib.rs
```diff
@@ -112,6 +113,15 @@ pub fn pairlist(item: TokenStream) -> TokenStream {
     pairlist::pairlist(item)
 }
 
+/// Call a
+/// ```ignore
+///     assert_eq!(pairlist!(a=1, 2, 3), Pairlist::from_pairs(&[("a", 1), ("", 2), ("", 3)]));
```
clauswilke, line 118: Is this the correct comment here? This is the `call` macro, not `Pairlist`.

extendr-macros/src/call.rs
```diff
@@ -0,0 +1,59 @@
...
+
+        while !input.is_empty() {
+            input.parse::<Token![,]>()?;
+            res.pairs.push(input.parse::<Expr>()?);
+        }
+        Ok(res)
+    }
+}
+
+pub fn call(item: TokenStream) -> TokenStream {
```
clauswilke, line 28: Could you add some comments to the body of this macro explaining what the different pieces of the code do?

extendr-macros/src/lib.rs
```diff
@@ -112,6 +113,15 @@ pub fn pairlist(item: TokenStream) -> TokenStream {
     pairlist::pairlist(item)
 }
 
+/// Call a
```
clauswilke, line 116: Looks like something's missing here.

clauswilke (commented):
Generally looks good to me. I had a few minor comments and questions, see specific comments.

extendr-api/tests/call_tests.rs
andy-thomason: I've changed all the R! macros without quotes I could find to have quotes.

extendr-macros/src/call.rs
andy-thomason: Done.

extendr-macros/src/lib.rs
andy-thomason: Done.

andy-thomason:
There are many ways we can optimise here when the time comes.

We could even generate stubs for R libraries at compile time with something like:

```
mod base {
   include_R_namespace!("base");
}

mod stats {
   include_R_namespace!("stats");
}


...

base::sum(...);
rstats::dbeta(...);
```

extendr-macros/src/call.rs
```diff
@@ -0,0 +1,66 @@
...
+                if let Expr::Path(ExprPath { path, .. }) = &**left {
+                    let s = path.get_ident().unwrap().to_string();
+                    return parse_quote!( (#s, extendr_api::Robj::from(#right)) );
+                }
+            }
+            parse_quote!( ("", extendr_api::Robj::from(#e)) )
+        })
+        .collect::<Vec<Expr>>();
+
+    // us eval_string to convert the literal string into a callable object.
```
clauswilke, line 49: "use"

clauswilke (approved):
Great, thanks! Just saw one more minor typo.

andy-thomason:
If there are no objections, I'll merge.

clauswilke:
Yes, go ahead.

PR #242, opened 2021-07-24
Return REALSXP when the type can exceed R's integer range

### LINKED ISSUE #240: Return type i64, u64, and u32 should be converted to REALSXP
While we now can use `use_try_from = true` for type-checking, `TryFrom` is implemented one-way (R to Rust), if I understand correctly

I'm wondering what should happen in this case.

``` r
rextendr::rust_function(r"(
fn f() -> i64 {
  i64::MAX
}
)",
  patch.crates_io = list(`extendr-api` = list(git = "https://github.com/extendr/extendr")),
)
#> ℹ build directory: '/tmp/RtmpR1VmHe/file161681b33c70a'
#> ✓ Writing '/tmp/RtmpR1VmHe/file161681b33c70a/target/extendr_wrappers.R'.

f()
#> [1] -1
```

<sup>Created on 2021-07-23 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

yutannihilation:
Fix #240

### DIFF
--- a/extendr-api/src/robj/tests.rs
+++ b/extendr-api/src/robj/tests.rs
@@ -235,6 +235,15 @@ fn test_to_robj() {
         assert_eq!(Robj::from(1.0_f32), Robj::from(1.));
         assert_eq!(Robj::from(1.0_f64), Robj::from(1.));
 
+        // check large values
+        assert_eq!(Robj::from(i64::MAX), Robj::from(i64::MAX as f64));
+        assert_eq!(Robj::from(i64::MIN), Robj::from(i64::MIN as f64));
+
+        // check NaN and Inf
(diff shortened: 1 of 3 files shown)


extendr-api/src/robj/tests.rs
```diff
@@ -225,12 +225,13 @@ fn test_to_robj() {
         assert_eq!(Robj::from(true), Robj::from([Bool::from(true)]));
         //assert_eq!(Robj::from(1_u8), Robj::from(1));
```
yutannihilation, line 226: By the way, why is this test commented out?

clauswilke (approved):

PR #243, opened 2021-07-24
Check the limits and the rounding of to-int conversions

### LINKED ISSUE #223: Update impl_try_from_scalar! to check limits and rounding.
Currently, `impl_try_from_scalar!` uses "as" to convert between scalar types.

This doesn't do type checking.

We could maybe use TryFrom on scalar types to do the conversion.

We could add an error type of CouldNotConvert(Robj) in case this fails.

So if we do:

```
fn test_i32(val: i32) -> i32 {
    val
}


// Then this should fail with Error::CouldNotConvert
new_owned(wrap__test_i32(r!(1.5).get()))

```

Likewise for i16 and r!(32768)

yutannihilation:
Fix #223

### DIFF
--- a/extendr-api/src/error.rs
+++ b/extendr-api/src/error.rs
@@ -67,6 +67,8 @@ pub enum Error {
     OutOfRange(Robj),
     MustNotBeNA(Robj),
     ExpectedNonZeroLength(Robj),
+    ExpectedIntegerish(Robj),
+    OutOfLimits(Robj),
     TypeMismatch(Robj),
     NamespaceNotFound(Robj),
 
@@ -128,6 +130,10 @@ impl std::fmt::Display for Error {
             Error::OutOfRange(_robj) => write!(f, "Out of range."),
             Error::MustNotBeNA(_robj) => write!(f, "Must not be NA."),
             Error::ExpectedNonZeroLength(_robj) => write!(f, "Expected non zero length"),
+            Error::ExpectedIntegerish(robj) => {
+                write!(f, "Expected an integer-ish value, got {:?}", robj)
+            }
+            Error::OutOfLimits(robj) => write!(f, "The value is too big: {:?}", robj),
             Error::TypeMismatch(_robj) => write!(f, "Type mismatch"),
 
             Error::NamespaceNotFound(robj) => write!(f, "Namespace {:?} not found", robj),
--- a/extendr-api/src/error.rs
+++ b/extendr-api/src/error.rs
@@ -67,7 +67,7 @@ pub enum Error {
     OutOfRange(Robj),
     MustNotBeNA(Robj),
     ExpectedNonZeroLength(Robj),
-    ExpectedIntegerish(Robj),
+    ExpectedWholeNumber(Robj),
     OutOfLimits(Robj),
     TypeMismatch(Robj),
     NamespaceNotFound(Robj),
@@ -130,8 +130,12 @@ impl std::fmt::Display for Error {
             Error::OutOfRange(_robj) => write!(f, "Out of range."),
             Error::MustNotBeNA(_robj) => write!(f, "Must not be NA."),
             Error::ExpectedNonZeroLength(_robj) => write!(f, "Expected non zero length"),
-            Error::ExpectedIntegerish(robj) => {
-                write!(f, "Expected an integer-ish value, got {:?}", robj)
+            Error::ExpectedWholeNumber(robj) => {
+                write!(
+                    f,
+                    "Expected an integer or a float representing a whole number, got {:?}",
+                    robj
+                )
             }
             Error::OutOfLimits(robj) => write!(f, "The value is too big: {:?}", robj),
             Error::TypeMismatch(_robj) => write!(f, "Type mismatch"),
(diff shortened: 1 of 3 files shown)


extendr-api/src/error.rs
```diff
@@ -128,6 +130,10 @@ impl std::fmt::Display for Error {
             Error::OutOfRange(_robj) => write!(f, "Out of range."),
             Error::MustNotBeNA(_robj) => write!(f, "Must not be NA."),
             Error::ExpectedNonZeroLength(_robj) => write!(f, "Expected non zero length"),
+            Error::ExpectedIntegerish(robj) => {
+                write!(f, "Expected an integer-ish value, got {:?}", robj)
```
clauswilke, line 134: Not sure if "integer-ish value" is understandable. Maybe "integer or a double representing a whole number"?

extendr-api/src/robj/try_from_robj.rs
```diff
@@ -2,54 +2,95 @@
...
+                    _ => return Err(Error::ExpectedScalar(robj)),
+                };
+
+                // Check if the value is not a missing value
+                if robj.is_na() {
+                    return Err(Error::MustNotBeNA(robj));
+                }
+
+                // If the conversion is int-to-int, check the limits. This
+                // needs to be down by `TryFrom` because the conversion by `as`
```
clauswilke, line 24: "done" instead of "down"

extendr-api/src/robj/try_from_robj.rs
```diff
@@ -2,54 +2,95 @@
...
-                        _ => Err(Error::ExpectedScalar(robj)),
+                }
+
+                // If the conversion is float-to-int, check if the value is
+                // integer-ish. This needs to be down with `as`, as no `TryFrom`
+                // is implemented for float types. `FloatToInt` trait might
+                // eventually become available in future, though.
+                if let Some(v) = robj.as_real() {
+                    let result = v as Self;
+                    if ((result as f64 - v).abs() < f64::EPSILON) {
```
clauswilke, line 40: Didn't @andy-thomason suggest converting from float to integer back to float and testing for equality with the original value? Not sure which way is better.

extendr-api/src/robj/try_from_robj.rs
```diff
@@ -2,54 +2,95 @@
...
+                if robj.is_na() {
+                    return Err(Error::MustNotBeNA(robj));
+                }
+
+                if let Some(v) = robj.as_integer() {
+                    return Ok(v as Self);
+                }
+                if let Some(v) = robj.as_real() {
+                    return Ok(v as Self);
+                }
```
clauswilke, line 76: Not sure I understand the logic here (first trying to convert to integer, then to real). Could you add a few comments explaining what is happening?

clauswilke (commented):
Overall looks good. I had a few comments and questions.

extendr-api/src/error.rs
clauswilke: Maybe write "float" or "numeric" instead of "double"?

extendr-api/src/robj/try_from_robj.rs
yutannihilation: Yes, I first implemented so, but clippy warned.

https://rust-lang.github.io/rust-clippy/master/#float_cmp

clauswilke: Funny, because in this case we do actually want to test for exact equality. Anyways, your current implementation looks fine to me, so let's just leave it as is.

extendr-api/src/robj/try_from_robj.rs
yutannihilation: `as_xxx()` can convert only when the underlying SEXP is the same type, so we cannot use `as_real()` directly on INTSXP. Does this answer your question? I might not get your point, sorry.

extendr-api/src/robj/try_from_robj.rs
yutannihilation: I think what the current code does is just the verbose version of checking "exact equality." When the difference is unrepresentable even by the machine epsilon, two values should be the same. So, I think we can choose to silence clippy here, but this code is probably a good reminder about the trickiness of comparing two floating point numbers.

extendr-api/src/error.rs
yutannihilation: Thanks, sounds good. I picked the word "integer-ish" from [an rlang's function](https://rlang.r-lib.org/reference/is_integerish.html), but probably it's better to say "a whole number."

extendr-api/src/robj/try_from_robj.rs
clauswilke: Yes. If you could put something like that into a comment above this section of the code that would be great.

yutannihilation: I see. I wondered if we should implement such a method with coercion (`Rf_coerceVector`) like `to_real()` on `Robj`'s side, but let's document the current situation for now.

clauswilke (approved):

yutannihilation:
@andy-thomason 
Could you take a look on this if you have time? If you are busy, please ignore, I'm just merging this as this change is probably trivial.

yutannihilation:
Thanks!

PR #244, opened 2021-07-25
[WIP] document the conventions on conversions between R and Rust

yutannihilation:
Now `TryFrom` infrastructure is maturing, but we still have `FromRobj`, in addition to other methods. So, I think it's good time to have some document about the conventions (or semantics?) on various types of conversions to avoid confusion. I feel it's not only users but also us who don't get the sense of when to use what. Though I think I'm not the right person to address this, but here's my attempt to document the current status, mainly to start a discussion, rather than to actually get merged. Feedback is welcome, and if it's better to discuss on an issue instead of this pull request, I'll file a new one.

Here's an appendix, in case this is useful:
<https://gist.github.com/yutannihilation/15e1f90099985f35c331ea78b8edb4cd#file-99_appendix-md>

### DIFF
--- a/extendr-api/src/robj/mod.rs
+++ b/extendr-api/src/robj/mod.rs
@@ -8,6 +8,57 @@
 //! * Users should be able to do almost everything without using libR_sys.
 //! * The interface should be friendly to R users without Rust experience.
 //!
+//! ## Conversion between R and Rust
+//!
+//! The following types of traits and methods are provided for conversions from
+//! R objects to Rust objects:
+//!
+//! - **`try_from()`** ([TryFrom] trait): This is the main way of doing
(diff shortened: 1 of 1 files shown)


andy-thomason (approved):

andy-thomason:
Very nice...

yutannihilation:
I think this needs to be updated to reflect the discussion on #261. Sorry for the delay...

andy-thomason:
Shall I close this? We can always re-cycle the text.

We need to have a good chat about documentation ready for the next release.

yutannihilation:
Sure, let's close this. While I still think it's nice if I can finish this, it's probably not possible in the near future...

PR #245, opened 2021-07-25
Fix broken crates.io badge

dasmoth:
The badge-generation service seems to have disappeared.  Here's an updated one, to match the new website.


yutannihilation (approved):
I was wondering why the badge won't be rendered... Thanks!

PR #249, opened 2021-08-02
More try_from conversions

andy-thomason:
This PR covers i32, f64, Bool and u8 types in vectors and slices.

```
        assert_eq!(<Vec::<i32>>::try_from(Robj::from(1)), Ok(vec![1]));
        assert_eq!(<Vec::<f64>>::try_from(Robj::from(1.)), Ok(vec![1.]));
        assert_eq!(<Vec::<Bool>>::try_from(Robj::from(TRUE)), Ok(vec![TRUE]));
        assert_eq!(<Vec::<u8>>::try_from(Robj::from(0_u8)), Ok(vec![0_u8]));

        assert_eq!(<&[i32]>::try_from(Robj::from(1)), Ok(&[1][..]));
        assert_eq!(<&[f64]>::try_from(Robj::from(1.)), Ok(&[1.][..]));
        assert_eq!(<&[Bool]>::try_from(Robj::from(TRUE)), Ok(&[TRUE][..]));
        assert_eq!(<&[u8]>::try_from(Robj::from(0_u8)), Ok(&[0_u8][..]));
```

### DIFF
--- a/extendr-api/src/robj/tests.rs
+++ b/extendr-api/src/robj/tests.rs
@@ -161,6 +161,12 @@ fn test_try_from_robj() {
         assert_eq!(<&[Bool]>::try_from(Robj::from(TRUE)), Ok(&[TRUE][..]));
         assert_eq!(<&[u8]>::try_from(Robj::from(0_u8)), Ok(&[0_u8][..]));
 
+        // Note the Vec<> cases use the same logic as the slices.
+        assert_eq!(<&[i32]>::try_from(Robj::from(1.0)), Err(Error::ExpectedInteger(r!(1.0))));
+        assert_eq!(<&[f64]>::try_from(Robj::from(1)), Err(Error::ExpectedReal(r!(1))));
+        assert_eq!(<&[Bool]>::try_from(Robj::from(())), Err(Error::ExpectedLogical(r!(()))));
+        assert_eq!(<&[u8]>::try_from(Robj::from(())), Err(Error::ExpectedRaw(r!(()))));
+
         let hello = Robj::from("hello");
         assert_eq!(<&str>::try_from(hello), Ok("hello"));
 
--- a/extendr-api/src/robj/tests.rs
+++ b/extendr-api/src/robj/tests.rs
@@ -153,9 +153,19 @@ fn test_try_from_robj() {
 
         assert_eq!(<Vec::<i32>>::try_from(Robj::from(1)), Ok(vec![1]));
         assert_eq!(<Vec::<f64>>::try_from(Robj::from(1.)), Ok(vec![1.]));
+        assert_eq!(<Vec::<Bool>>::try_from(Robj::from(TRUE)), Ok(vec![TRUE]));
+        assert_eq!(<Vec::<u8>>::try_from(Robj::from(0_u8)), Ok(vec![0_u8]));
+
+        assert_eq!(<&[i32]>::try_from(Robj::from(1)), Ok(&[1][..]));
+        assert_eq!(<&[f64]>::try_from(Robj::from(1.)), Ok(&[1.][..]));
+        assert_eq!(<&[Bool]>::try_from(Robj::from(TRUE)), Ok(&[TRUE][..]));
+        assert_eq!(<&[u8]>::try_from(Robj::from(0_u8)), Ok(&[0_u8][..]));
 
         let hello = Robj::from("hello");
         assert_eq!(<&str>::try_from(hello), Ok("hello"));
+
+        let hello = Robj::from("hello");
+        assert_eq!(<String>::try_from(hello), Ok("hello".into()));
 
         // conversion from a vector to a scalar value
         let robj = Robj::from(vec![].as_slice() as &[i32]);
--- a/extendr-api/src/robj/try_from_robj.rs
+++ b/extendr-api/src/robj/try_from_robj.rs
@@ -7,6 +7,7 @@ macro_rules! impl_try_from_scalar_integer {
         impl TryFrom<Robj> for $t {
             type Error = Error;
 
+            /// Convert a numeric object to an integer value.
             fn try_from(robj: Robj) -> Result<Self> {
                 // Check if the value is a scalar
                 match robj.len() {
@@ -57,6 +58,7 @@ macro_rules! impl_try_from_scalar_real {
         impl TryFrom<Robj> for $t {
             type Error = Error;
 
+            /// Convert a numeric object to a real value.
             fn try_from(robj: Robj) -> Result<Self> {
                 // Check if the value is a scalar
                 match robj.len() {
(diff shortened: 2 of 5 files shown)


extendr-api/src/robj/try_from_robj.rs
```diff
@@ -178,6 +178,30 @@ impl TryFrom<Robj> for Vec<f64> {
...
+}
+
+impl TryFrom<Robj> for Vec<u8> {
+    type Error = Error;
+
+    fn try_from(robj: Robj) -> Result<Self> {
+        if let Some(v) = robj.as_raw_slice() {
+            Ok(Vec::from(v))
+        } else {
+            Err(Error::ExpectedInteger(robj))
```
clauswilke, line 200: Is this the right error? It looks like you're interpreting the input in raw format, shouldn't the error then be something about not being able to use the data in raw format?

extendr-api/src/robj/try_from_robj.rs
```diff
@@ -178,6 +178,30 @@ impl TryFrom<Robj> for Vec<f64> {
...
+    fn try_from(robj: Robj) -> Result<Self> {
+        if let Some(v) = robj.as_logical_slice() {
+            Ok(Vec::from(v))
+        } else {
+            Err(Error::ExpectedInteger(robj))
+        }
+    }
+}
+
+impl TryFrom<Robj> for Vec<u8> {
```
clauswilke, line 193: Is your assumption here that `Vec<u8>` should only be used when the R object is a raw vector, and never if it is an R numeric or integer? If yes, then I think this should be documented.

extendr-api/src/robj/tests.rs
```diff
@@ -153,10 +153,20 @@ fn test_try_from_robj() {
 
         assert_eq!(<Vec::<i32>>::try_from(Robj::from(1)), Ok(vec![1]));
         assert_eq!(<Vec::<f64>>::try_from(Robj::from(1.)), Ok(vec![1.]));
+        assert_eq!(<Vec::<Bool>>::try_from(Robj::from(TRUE)), Ok(vec![TRUE]));
+        assert_eq!(<Vec::<u8>>::try_from(Robj::from(0_u8)), Ok(vec![0_u8]));
+
+        assert_eq!(<&[i32]>::try_from(Robj::from(1)), Ok(&[1][..]));
```
clauswilke, line 159: What happens when there is a type mismatch, e.g.:
```rust
<&[i32]>::try_from(Robj::from(1.5))
```
Should there be assertion statements for these kinds of cases?

clauswilke (commented):
Looks good overall! I have a few comments and questions, though.

extendr-api/src/robj/try_from_robj.rs
andy-thomason: Done

extendr-api/src/robj/try_from_robj.rs
andy-thomason: I've documented all the conversions.

extendr-api/src/robj/tests.rs
andy-thomason: Tests added,

extendr-api/src/robj/tests.rs
```diff
@@ -153,10 +153,26 @@ fn test_try_from_robj() {
...
+        assert_eq!(<&[f64]>::try_from(Robj::from(1.)), Ok(&[1.][..]));
+        assert_eq!(<&[Bool]>::try_from(Robj::from(TRUE)), Ok(&[TRUE][..]));
+        assert_eq!(<&[u8]>::try_from(Robj::from(0_u8)), Ok(&[0_u8][..]));
+
+        // Note the Vec<> cases use the same logic as the slices.
+        assert_eq!(<&[i32]>::try_from(Robj::from(1.0)), Err(Error::ExpectedInteger(r!(1.0))));
```
clauswilke, line 165: Is this correct? Shouldn't it convert 1.0 into 1, and only cause an error if the float doesn't correspond to a whole number?

extendr-api/src/robj/tests.rs
```diff
@@ -153,10 +153,26 @@ fn test_try_from_robj() {
...
+        assert_eq!(<&[f64]>::try_from(Robj::from(1)), Err(Error::ExpectedReal(r!(1))));
```
clauswilke, line 166: Same as above. Shouldn't this succeed?

clauswilke (commented):
The thing that I don't fully understand yet: Shouldn't the contents of the vectors/slices also be converted with `try_from()`? It looks to me like we're applying different rules for scalars and vectors/slices right now, e.g. auto-conversion of floats to integers and back.

extendr-api/src/robj/tests.rs
andy-thomason: The error reports the object that generated it.

I guess we could have more useful error messages.

extendr-api/src/robj/tests.rs
andy-thomason: Ditto.

andy-thomason: The slices are a bit of an exception as they give immutable reference access to the interiors of the R objects.

You can still check for NA using `value.is_na()` but checking all the values of a gigabyte-long vector may be undesirable.

andy-thomason: We should consider having mutable slices also `&mut [i32]` etc.

This would allow the modification of R objects in-place.

clauswilke: Ok, I get the argument for the slices, but not for vectors. There we're making a copy of the data, correct? So then we should use the same conversion and rules that we're using for scalars (convert 1.0 to 1 etc.)

clauswilke (approved):
I've decided to approve, as this PR moves the project forward.

However, the issue of numerical vectors is not fully resolved, as R doesn't do a good job of distinguishing between integer and floating point values. We need some sort of option that allows a Rust function to accept either integer or float vectors as input and use them as either floats or integers, with auto-conversion as appropriate.

extendr-api/src/robj/tests.rs
yutannihilation: > The slices are a bit of an exception as they give immutable reference access to the interiors of the R objects.

So, this is also the reason why no other int/float variants (e.g. i64 and f32) are supported in slice's case, right? I'll document the rule on #244.

yutannihilation:
> We need some sort of option that allows a Rust function to accept either integer or float vectors as input and use them as either floats or integers, with auto-conversion as appropriate.

I agree. I think we might need to introduce some wrapper like `Nullable`. Or, if it's not necessarily on Rust's side, call `as.xxx()` function in the wrapper R function?

extendr-api/src/robj/try_from_robj.rs
```diff
@@ -195,25 +243,13 @@ impl TryFrom<Robj> for Vec<String> {
...
-//         }
-//     }
-// }
-
 macro_rules! impl_option {
     ($type : ty) => {
         impl TryFrom<Robj> for Option<$type> {
             type Error = Error;
 
+            /// Convert a scalar object that may be NA type to a n `Option` of a corresponding type.
```
yutannihilation, line 251:
```suggestion
            /// Convert a scalar object that may be NA type to an `Option` of a corresponding type.
```

PR #250, opened 2021-08-06
Full Altrep Support

andy-thomason:
This PR adds an "Altrep" wrapper and the beginnings of the means to implement new Altrep classes.

### DIFF
--- a/extendr-api/src/lib.rs
+++ b/extendr-api/src/lib.rs
@@ -457,6 +457,38 @@ pub fn rtype_to_sxp(rtype: RType) -> i32 {
     }) as i32
 }
 
+/// Convert R's SEXPTYPE to extendr's RType.
+pub fn sxp_to_rtype(sxptype: i32) -> RType {
+    use RType::*;
+    match sxptype as u32 {
+        NILSXP => Null,
+        SYMSXP => Symbol,
+        LISTSXP => Pairlist,
+        CLOSXP => Function,
+        ENVSXP => Environment,
+        PROMSXP => Promise,
+        LANGSXP => Language,
+        SPECIALSXP => Special,
+        BUILTINSXP => Builtin,
+        CHARSXP => Character,
+        LGLSXP => Logical,
+        INTSXP => Integer,
+        REALSXP => Real,
+        CPLXSXP => Complex,
+        STRSXP => String,
+        DOTSXP => Dot,
+        ANYSXP => Any,
+        VECSXP => List,
+        EXPRSXP => Expression,
+        BCODESXP => Bytecode,
+        EXTPTRSXP => ExternalPtr,
+        WEAKREFSXP => WeakRef,
+        RAWSXP => Raw,
+        S4SXP => S4,
+        _ => Unknown
+    }
+}
+
 #[doc(hidden)]
 pub fn print_r_output<T: Into<Vec<u8>>>(s: T) {
     let cs = CString::new(s).expect("NulError");
--- a/extendr-api/src/lib.rs
+++ b/extendr-api/src/lib.rs
@@ -485,7 +485,7 @@ pub fn sxp_to_rtype(sxptype: i32) -> RType {
         WEAKREFSXP => WeakRef,
         RAWSXP => Raw,
         S4SXP => S4,
-        _ => Unknown
+        _ => Unknown,
     }
 }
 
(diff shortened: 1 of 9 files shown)


clauswilke:
Super, thanks! Could you add some comments that explain what the various API functions do?

extendr-api/src/wrapper/altrep.rs
```diff
@@ -6,31 +6,49 @@ pub struct Altrep {
...
         ().into()
     }
 
+    /// Fetch the state of this object when writing to a file.
     fn serialized_state(&self) -> Robj {
         ().into()
     }
 
+    /// Duplicate this object, possibly duplicating attributes.
     fn duplicate_ex(&self, _deep: bool) -> Robj {
```
clauswilke, line 37: What does "ex" stand for? "extended"? Can you think of a different choice that is a little more intuitive?

clauswilke (approved):

andy-thomason:
I'm not sure yet how we will define new Altrep classes.

Given a trait like the ones described, we should be able to make a "universal adapter" from the Altrep
callbacks to our trait methods.

To test, I'll implement `From<Range<i32>>` as an altrep class.

extendr-api/src/lib.rs
```diff
@@ -422,6 +424,37 @@ pub enum RType {
     Unknown,
 }
 
+pub fn rtype_to_sxp(rtype: RType) -> i32 {
```
clauswilke, line 427: I think a brief, 1-2 sentence documentation would be helpful here also.

andy-thomason:
I'm probably going to narrow the definition of Altrep to make it easier to use.

Altrep objects have

* A class object (a raw object wrapping the vtable)
* A state object
* An optional materialized vector 

If we always use a struct for the state object (ie. Extptr) we can implement
the altrep functions as a methods of the state.

In an ideal world, we would just need to implement the "elt" method for an
altrep object and leave all the rest to Rust.

The state object should be a regular Rust object rather than an R object.

```rust
    struct MyCompactIntRange {
        start: i32,
        end: i32,
        step: i32,
    }

    impl AltIntegerImpl for MyCompactIntRange {
        fn elt(&self, index: usize) -> i32 {
            self.start + self.step * index
        }
    }
```

yutannihilation:
`ALTREP_LENGTH` might have gone in the recent R API change.

``` 
   error[E0425]: cannot find function, tuple struct or tuple variant `ALTREP_LENGTH` in this scope
    --> extendr-api/src/wrapper/altrep.rs:83:27
     |
  83 |                 let len = ALTREP_LENGTH(x);
     |                           ^^^^^^^^^^^^^ not found in this scope
  
```

andy-thomason:
@yutannihilation did we establish a mechanism for using #[cfg] with libR-sys?

If so, I can screen out some of the newer features.

The `ALTREP_` functions are good shortcuts to calling altrep dispatches.

yutannihilation:
> did we establish a mechanism for using #[cfg] with libR-sys?

(Sorry, I forgot to answer...) Yes, for internal usages within extendr crates, we can set cfg in build.rs.

https://github.com/extendr/extendr/blob/6cf4c1b457d01ce22f2f7fe875a4974e799122eb/extendr-api/build.rs#L10-L17

> If so, I can screen out some of the newer features.

In case a feature is not available in the older versions of R, it sounds the right thing to do. But, this time, the case is that it's not available on the newer version, which means we'll need to remove it in future when the current versions are out of the range of our minimum supported version. So, I'm wondering if it's worth the effort...

clauswilke:
> But, this time, the case is that it's not available on the newer version, which means we'll need to remove it in future when the current versions are out of the range of our minimum supported version. So, I'm wondering if it's worth the effort...

Yeah, I had the same thoughts. If it's a feature that is going away we should just work without for all R versions. Or, if critical, reach out to the R core developers and ask them to not remove the feature.

yutannihilation:
For `ALTREP_LENGTH()`, it seems we can use `XLENGTH_EX()` instead.

https://github.com/wch/r-source/blob/ea1c48271846cd02ad5f93b5562152ad82778d8d/src/include/Rinlinedfuns.h#L202-L205

yutannihilation:
Or just `Rf_xlength()`? Sorry, I'm not familiar with the things around here...

andy-thomason:
Claus is right that we can't rely on the functions now in `defn.h`

The sum, min and max functions should work for code, so we can always
call them using `call!`

I still have to implement defaults duplicate and coerce, but then we can merge.

I am planning to finish off the last few SXP types as part of the grant contract.

PR #254, opened 2021-08-08
Remove TryFrom/Into conversion of HashMap, and implement into_hashmap() and from_hashmap()

### LINKED ISSUE #230: Conversions to and from hashmap
See https://github.com/extendr/extendr/pull/229#issuecomment-882090169

Should hashmaps be converted to lists or environments by default? I think both conversions should be available, and then one can be made the default if the user doesn't specify what result they want.

There's also the question what happens if an unnamed list is converted to a hashmap. It should probably fail with an error.

yutannihilation:
Close #230

- [x] `<List>::from_hashmap()`
- [x] `<List>::into_hashmap()`
- ~~`<HashMap<>>::from_list()`~~
- ~~`<HashMap<>>::into_list()`~~


andy-thomason:
Do you need some help with this?

yutannihilation:
@andy-thomason 
Can I merge this?

Issue #247, opened 2021-07-29
Is it intentional not to provide TryFrom<Robj> for &[T] ?


yutannihilation:
`FromRobj` implementation accepts `&[T]`, but `TryFrom` doesn't. Is this intentional? 

``` r
rextendr::rust_source(code = r"(
#[extendr(use_try_from = false)]
fn f1(x: &[i32]) -> i32 {
    x.iter().sum()
}
)", patch.crates_io = list(`extendr-api` = list(git = "https://github.com/yutannihilation/extendr", branch = "fix/coerce-real-to-int"))
)
#> ℹ build directory: '/tmp/RtmpvYlJzS/file79203ad54351'
#> ✓ Writing '/tmp/RtmpvYlJzS/file79203ad54351/target/extendr_wrappers.R'.

# cannot compile
rextendr::rust_source(code = r"(
#[extendr(use_try_from = true)]
fn f2(x: &[i32]) -> i32 {
    x.iter().sum()
}
)", patch.crates_io = list(`extendr-api` = list(git = "https://github.com/yutannihilation/extendr", branch = "fix/coerce-real-to-int"))
)
#> ℹ build directory: '/tmp/RtmpvYlJzS/file79203ad54351'
#> Error: Rust code could not be compiled successfully. Aborting.
```

<sup>Created on 2021-07-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

yutannihilation:
I also wondered how it's useful to use `&[T]` as input. As these are not the reference to the original values but to the converted values from Robj, so I guess it's no problem to consume the input?

andy-thomason:
We should cover everything.

We may need to work out what we are missing and then switch to try_from as default.

yutannihilation:
I see.

PR #258, opened 2021-08-13
TryFrom<Robj> for Int, Logical, Real

### LINKED ISSUE #252: Implement TryFrom<Robj> for Int, Logical, Real
These seems missing, compared to `FromRobj` implementation

https://extendr.github.io/extendr/extendr_api/trait.TryFrom.html#implementors

brendanrbrown:
Closes #252

### DIFF
--- a/extendr-api/src/robj/tests.rs
+++ b/extendr-api/src/robj/tests.rs
@@ -155,6 +155,18 @@ fn test_try_from_robj() {
         assert_eq!(<Vec::<f64>>::try_from(Robj::from(1.)), Ok(vec![1.]));
         assert_eq!(<Vec::<Bool>>::try_from(Robj::from(TRUE)), Ok(vec![TRUE]));
         assert_eq!(<Vec::<u8>>::try_from(Robj::from(0_u8)), Ok(vec![0_u8]));
+        
+        let v: Result<Real> = r!(NA_REAL).try_into();
+        let mut v: Vec<_> = v.unwrap().collect();
+        assert!(v.pop().unwrap().is_nan());
+        assert_eq!(<Real>::try_from(r!([1.0, 2.0])).unwrap().collect::<Vec<f64>>(), vec![1.0, 2.0]);
+        assert!(<Real>::try_from(r!([true])).is_err());
+
+        assert_eq!(<Int>::try_from(r!([1, 2])).unwrap().collect::<Vec<i32>>(), vec![1, 2]);
+        assert!(<Int>::try_from(r!([true])).is_err());
+
+        assert_eq!(<Logical>::try_from(r!([true, false])).unwrap().collect::<Vec<Bool>>(), vec![TRUE, FALSE]);
+        assert!(<Logical>::try_from(r!([1])).is_err());
 
         assert_eq!(<&[i32]>::try_from(Robj::from(1)), Ok(&[1][..]));
         assert_eq!(<&[f64]>::try_from(Robj::from(1.)), Ok(&[1.][..]));
--- a/extendr-api/src/robj/tests.rs
+++ b/extendr-api/src/robj/tests.rs
@@ -156,6 +156,10 @@ fn test_try_from_robj() {
         assert_eq!(<Vec::<Bool>>::try_from(Robj::from(TRUE)), Ok(vec![TRUE]));
         assert_eq!(<Vec::<u8>>::try_from(Robj::from(0_u8)), Ok(vec![0_u8]));
 
+        // TODO: once related todos resolved in try_from_robj.rs, add tests for
+        // Real to Int successful case (e.g. 1.0 to 1) and failing case
+        // Int to Real
+        // NA handling
         let v: Result<Real> = r!(NA_REAL).try_into();
         let mut v: Vec<_> = v.unwrap().collect();
         assert!(v.pop().unwrap().is_nan());
--- a/extendr-api/src/robj/tests.rs
+++ b/extendr-api/src/robj/tests.rs
@@ -155,7 +155,7 @@ fn test_try_from_robj() {
         assert_eq!(<Vec::<f64>>::try_from(Robj::from(1.)), Ok(vec![1.]));
         assert_eq!(<Vec::<Bool>>::try_from(Robj::from(TRUE)), Ok(vec![TRUE]));
         assert_eq!(<Vec::<u8>>::try_from(Robj::from(0_u8)), Ok(vec![0_u8]));
-        
+
         let v: Result<Real> = r!(NA_REAL).try_into();
         let mut v: Vec<_> = v.unwrap().collect();
         assert!(v.pop().unwrap().is_nan());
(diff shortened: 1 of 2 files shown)


clauswilke:
Not sure how this relates to #251 after most recent conversations. We need to figure out some coherent approach to handling numeric vectors that does appropriate type conversions and checking, like we are doing with scalars.

extendr-api/src/robj/tests.rs
```diff
@@ -156,6 +156,18 @@ fn test_try_from_robj() {
         assert_eq!(<Vec::<Bool>>::try_from(Robj::from(TRUE)), Ok(vec![TRUE]));
         assert_eq!(<Vec::<u8>>::try_from(Robj::from(0_u8)), Ok(vec![0_u8]));
 
+        let v: Result<Real> = r!(NA_REAL).try_into();
+        let mut v: Vec<_> = v.unwrap().collect();
+        assert!(v.pop().unwrap().is_nan());
+        assert_eq!(<Real>::try_from(r!([1.0, 2.0])).unwrap().collect::<Vec<f64>>(), vec![1.0, 2.0]);
+        assert!(<Real>::try_from(r!([true])).is_err());
+
+        assert_eq!(<Int>::try_from(r!([1, 2])).unwrap().collect::<Vec<i32>>(), vec![1, 2]);
```
clauswilke, line 165: What happens with:
```rust
<Int>::try_from(r!([1.0, 2.0])).unwrap().collect::<Vec<i32>>()
```
Does it work? Often, R numerics will be floats, even if they are meant to store integers.

brendanrbrown:
> Not sure how this relates to #251 after most recent conversations. 

Do you mean #252, which this closes? Or is the suggestion that the NA handling issue should be addressed before this one?

> We need to figure out some coherent approach to handling numeric vectors that does appropriate type conversions and checking, like we are doing with scalars.

Agreed. I see this PR as plugging a hole for now, following the behavior of similar implementations, but of course it should be to be changed as part of a move to better conversions in general.

extendr-api/src/robj/tests.rs
brendanrbrown: It fails with the appropriate error. This does raise the point that this behavior is not consistent with #243, but that probably should be addressed in the `as_{type}_iter` methods for `Robj`, which this PR's TryFrom conversions use.

clauswilke:
I meant #251, and specifically, I meant we need to have a coherent framework of how these things should work before continuing to implement.

@Ilia-Kosenkov suggested we should write a whitepaper outlining our data conversion model. This suggestion is on discord. I'll file an issue now here.

clauswilke:
See #259.

yutannihilation:
> I see this PR as plugging a hole for now as with #247, using the behavior of similar implementations.

Yeah, I thought the same at the time of adding "good first issue" label. Sorry for the confusion...

clauswilke:
How about something like the following. Instead of this conversion:
```rust
robj.as_integer_iter()
            .ok_or_else(|| Error::ExpectedInteger(robj))
```
could we implement something like this:
```rust
if let Ok(v) = robj.as_integer_iter() {
    v
} else {
    let rreal:Real = robj.try_into()?;
    let rint = Int::from_real(rreal);
    Ok(rint)
}
```
where `Int::from_real()` takes a `Real` and converts it into an `Int`, making a copy of the data in the process.

andy-thomason:
We could implement a "Cow" for numeric types.

A `Cow<[type]>` is either a slice or a vector depending on whether
a conversion has been done.

So `Cow<[i32]>` is `as_integer_slice` for and INTSXP and
a vector conversion for a REALSXP.

So `Cow<[f64]>` is `as_real_slice` for and REALSXP and
a vector conversion for a INTSXP.

andy-thomason:
I may re-work Int, Logical and Real to be just Robj wrappers and pull
out the iterator as a separate object.

We can implement GET_REGION for ALTREP objects which
would be faster.

andy-thomason:
But essentially, the logic will be the same.

andy-thomason:
I'm happy to approve this for now if @clauswilke is ok so that @brendanrbrown gets a contribution.

We can either implement `Cow` types for int/real mismatches or improve the iterators
to accept Altrep types and do the conversions as a separate step.

clauswilke:
Before merging, can we at least stub out an alternative path for now so we have a marker in the code that this isn't completed?

Something like so maybe?
```Rust
if let Ok(v) = robj.as_integer_iter() {
    Ok(v)
} else {
    // TODO: check if robj is another type (e.g. Real) that can be converted
    // into integer, and do the conversion
    Error::ExpectedInteger(robj)
}
```

Also, @andy-thomason, I'd suggest we wait with reworking how `Int`, `Real`, and `Logical` work until we have a bit more progress on an architecture whitepaper, #259. Let's think through all the problems first and then pick an implementation that is appropriate. To me, the question of whether we need types that can stand in for R's equivalent of `i32` and `f64` is unresolved. (We have something like it for `Logical` with `Bool`.)

brendanrbrown:
I added the todo comment for possible merging, should that be what makes sense here.

> could we implement something like this ... where `Int::from_real()` takes a `Real `and converts it into an `Int`

Yes, this seems reasonable, would not be difficult, and would be a step closer to being consistent (probably) with whatever future coherent conversion scheme, based on the conversation. If having it would make this hole-patching PR a better fit, I would be glad to add it.

> We could implement a "Cow" for numeric types.

`Cow` does seem a good choice for the `as_` conversions, for exactly the kind of 'maybe integer' conversions discussed above.

> I'm happy to approve this for now if @clauswilke is ok so that @brendanrbrown gets a contribution.

@andy-thomason: thanks for the thoughtfulness.

brendanrbrown:
Don't understand the CI failure here. Since the most recent passing tests I only moved text, but the isn't cargo-fmt related. Someone who knows more able to shed some light?

clauswilke:
> Don't understand the CI failure here.
It may be a probabilistic failure (i.e., works most of the time but breaks sometimes, randomly). I'll restart the tests to see if that fixes it.

clauswilke:
> If having it would make this hole-patching PR a better fit, I would be glad to add it.

If you don't mind I'd appreciate it. We could also break it into two separate PRs though.

I have a few more documentation requests, though:
- Could you add the same kind of TODO comment to the implementation for `Real`?
- In the test function, could you add a TODO comment that we will have to add tests for reals that can be converted to integers, reals that can't, integers converted to reals, and vectors containing `NA`s when reals get converted to ints and vice versa?

Thanks!

brendanrbrown:
@clauswilke My preference would be to address the additional conversions, e.g. using an `Int::from_real` method, in a separate PR. It'll be easier to review and comment on I think. Should be able to submit an initial PR for that this week.

clauswilke (approved):
Sounds good. Any idea why the one test is failing?

brendanrbrown:
It was the same failure as the previous case, in one of the windows builds, which appears inconsistently. I'm not quite sure what to make of it.

clauswilke:
Thanks!

PR #261, opened 2021-08-16
Marshaling rules

Ilia-Kosenkov:
This is a draft outlining some ideas of how to deal with vectors passed from `R` to `Rust`.
Proposed are several configuration flags that affect code generation.

So if a function is defined as `fn fn_1(x : Vec<i32>)`, default configuration will emit code that checks input for `NA`s, unfolds ALTREPs, safely coerces `double()` and `complex()` vectors if their values can be exactly represented with `i32`. Both `fn_1(c(1, 2, 3))` and `fn_1(1:3)` will do the same without errors.

Alternatively, a different configuration and different parameter type can be used to forbid any implicit conversions/validations, allowing the user to deal with (almost) raw arrays or altrep representations. Reduces to minimum overhead introduced by extendr.

Finally, a balanced approach is also possible. `fn fn_2(x : Vec<Rint>)` offloads validation onto user, but allowing for altrep unfolding.

This draft is just a... draft. Comments are welcomed, discussions are welcomed, ideas are welcomed.
It can be also rejected entirely if it is too complicated to implement/maintain, but it provides (in my opinion) idiomatic Rust way of handling complicated and unpredictable input.

NOTE: PR is made into a separate branch, not master.


clauswilke:
To me this is a good starting point. It might be worthwhile to integrate the materials from #244, as there is almost no overlap in material.

Ilia-Kosenkov:
It is not clear to me how to do that cleanly. Perhaps a PR can be made into this branch at some later point?

yutannihilation:
I too have no idea how to merge #244 into this, but I hope this document also covers non-`#[extendr]` conversions. While `#[extendr]` is the main part of extendr's ergonomics, it should be good to enumerate all types of conversions so that it might get easy to decide what to provide as the macro and what not to.

marshaling_rules.md
```diff
@@ -0,0 +1,121 @@
+# `R` <--> `Rust` marshaling rules
+
+This document outlines conversion and validation rules between `R` and `Rust` types.
+The aim is to maintain type safety on the `Rust` side without sacrificing usability on the `R` side.
+
+
+# Conversion and validation configuration
```
yutannihilation, line 7: Are these options necessary? Can't the configurations be determined uniquely by the type? I hope we can stick with `TryFrom` conversions for simplicity, if possible.

yutannihilation (commented):
Great work! Let me add comments mainly for discussion.

Ilia-Kosenkov:
@yutannihilation,
It is difficult to enumerate all possible type conversions even for vectors, but I hope we will be able to do this when we finalize all the rules.

marshaling_rules.md
```diff
@@ -0,0 +1,121 @@
...
+- `AltRepCmpl`
+- `AltRepStr`
+
+Altreps should be iterators of types `Rxxx`, so `AltRepInt` should produce `Rint`.
+If altreps cannot produce `NA`s, then they can iterate over primitive types like `i32` (i.e. with no `NA` validation mechanism).
+
+**Bare ALTREP types should not be allowed as function parameters, but may be used as return types**.
+
+### Handling both plain and ALTREP vectors
+To enable direct consumption of ALTREP vector wrappers (without explicit memory allocations), a simple approach can be used. The monad `Either` is an excellent solution for representing either plain vector or its ALTREP wrapper.
```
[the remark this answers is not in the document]
Ilia-Kosenkov: Why not? Not asking to expose `R` implementation of ALTREP, but a `Rust` wrapping iterator that only produces values?

I read a little bit about altreps, if I understood correctly, it is more or less an interface, altreps can represent not only simple things like sequences but also data stored on disk that cannot fit in memory (or should be read on demand).

If that is so, imagine a dot product method that accepts two vectors.

If it is a
```Rust
fn dot_product(x : Either<&[RFloat], AltRepFloat>, y  : Either<&[RFloat], AltRepFloat>) -> f64
```
then developer can decide at runtime if they are dealing with pre-allocated blocks of memory, which can be potentially read in chunks, verified against `NA` in chunks, and their product can be computed using some fuse-multiply-add and other intrinsic tricks, or if they are dealing with vector generators, which potentially wrap a very large data array with extremely poor support of random access. 

Of course, it is an advanced case, but I will not be surprised if such things can outperform some basic implementations of `R` functions, or be extremely memory-efficient.


At the same time, for general use-cases, just accept some `Vec<f64>`, and your altreps will be unfolded and each element verified against `NA` using simple checks. A lot of overhead but simple syntax, good for prototyping and `{rextendr}`.

marshaling_rules.md
Ilia-Kosenkov: I am sure they can, but we need to understand what types should be allowed in what regimes.

At some point when I was thinking about vectors, I realized there are three distinct problems that we must solve, which resulted in these configuration flags. 

When writing code, I prefer (and enjoy) two things: when syntactic sugar allows for simple, concise and readable code (with potential overhead under the hood), and when language features allow performant implementations using pointers and other tricks.
The first case is lazy `R` - `Rust` interop, when `Rust` function is called rarely, and provided input data are relatively small.
You implement `fn do_work(x : Vec<f64>)`, you call `do_work(1:100)`, and it just works.

However, if you want to speed up things and validate input yourself, you start using various `&[Rfloat]`, which just take ptr from `R` and use `std::slice::from_raw_parts`.

marshaling_rules.md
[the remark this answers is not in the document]
yutannihilation: For example, cpp11 handles it internally and the users don't need to care whether the input is an ALTREP or not. I feel this is easy to use.

https://github.com/r-lib/cpp11/blob/5fb3137c62c773b224bf1b3f5092bd556ea20db3/inst/include/cpp11/doubles.hpp#L30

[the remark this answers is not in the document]
Ilia-Kosenkov: I know, it is a reasonable solution, but afaik  cpp11 does not allow for `double[]` or `int[] `. How should we treat `Vec<T>` in extendr?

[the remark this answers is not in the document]
yutannihilation: > How should we treat `Vec<T>` in extendr?

I think I agree with you on this. We can make sure the vector is in the unfolded form and construct a `Vec<T>` or `&[T]` from the raw memory. That's totally fine.

What I'm suggesting is something like `Doubles` instead of `Either<&[RFloat], AltRepFloat>`.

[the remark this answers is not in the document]
yutannihilation: Honestly, I care little about performance and I'll probably stick with `Vec<T>` or `&[T]` for the sake of simplicity, so I'm not the right person to discuss this... Maybe @andy-thomason has some thoughts?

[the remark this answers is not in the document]
clauswilke: > What I'm suggesting is something like `Doubles` instead of `Either<&[RFloat], AltRepFloat>`.

I think that's an implementation detail, as `Doubles` could just be a wrapper for `Either`. However,  `Doubles` is preferable in my opinion, because it would present a single interface, so downstream users don't have to constantly special-case their code.

clauswilke:
@Ilia-Kosenkov To move things forward, let me suggest that for the issues for which there are differences in opinion or alternative options you present both in the document, possibly with pros and cons. I see this as a planning document, so let's record differing perspectives and possibilities rather than trying to solve all problems at once.

Ilia-Kosenkov:
OK I will add this.
I the mean time, what would be your opinion on handling `NA`s? If we accept `Vec<T>` of built in type, we have to iterate over it and panic.
`Vec<Option<T>>` is binary incompatible, I think, because we do not know internal memory layout of `Option`.
If we accept a `Vec<Rt>` (such as `Rint`), it is binary compatible with the plain data and it provides a single point of validation, and is `NA` aware.

clauswilke:
Yes, I like `Vec<Rt>`. It's an accurate representation of the underlying reality. Also, we could encapsulate all the various conversion rules into conversions from one `Rt` to another, e.g. `Rint` to `Rfloat`.

Ilia-Kosenkov:
@clauswilke,
The concept of `Rt` types compared to `T` (that is, `Rint` instead of `i32`) is the main reason I wanted to discuss this issue at all.
I did such tricks in other compiled languages, and if done carefully it works fine.

I will add this to the text now, also mentioning type-based conventions suggested by @yutannihilation.

Still, I believe we should support plain types like `Vec<i32>` for education/quick prototyping purposes. It will work nicely with `{rextend}` dynamic code compilation.

Ilia-Kosenkov:
@clauswilke,

>  Also, we could encapsulate all the various conversion rules into conversions from one `Rt` to another, e.g. `Rint` to `Rfloat`.

These conversions will be unlikely used by the consumers. The issue here is that if `Rust` expects `&[Rint]`, it cannot accept an `&[Rfloat]`, because they are binary incompatible in terms of size.  `R`, however, does not care. You can pass `c(1, 2, 3)` and `1:3`, both of which represent the same (from the `R` programmer perspective) entities, yet the first is a float vector, the second is an integer altrep.

Such type coercions should happen in the wrapper, i.e. before entering  `Rust` function, written by hand.
I see no value in conversions between `Rint` and `Rfloat` in user code. Need a float? Do `Rint` -> `Option<i32>` -> `f64`.

I would say `NA` handling is more or less understood. Either use a `NA`-aware binary compatible type, or use a regular type with additional runtime overhead that you have no control over.

Altrep also makes sense more or less, use `Iterator` that wraps any vector of given type, or use `Vec<T>` and pay for allocation for altrep under the hood.

The final problem is type conversion. One good example is how `{vctrs}` handles type casts [here](https://vctrs.r-lib.org/reference/theory-faq-coercion.html). `vctrs::vec_cast(1.0, integer())`, works, but `vctrs::vec_cast(1.5, integer())` does not, as it is lossy.
Type coercion in the wrapper defeats the idea of iterators, which produce little to no overhead and all processing happens when a user accesses elements of a vector. If we try to coerce types, we need to allocate memory. I dare to suggest to have something like `Numeric` iterator, which is a discriminated union of `Integer`, `Real` and `Complex`, and such type will be able to wrap any numeric input without extra allocations (because each union member itself wraps either a data array or an altrep).

Ilia-Kosenkov:
After reviewing suggestions, I still have no strong opinion on what to do if we expect `&[Rint]`, but get `double()`. This case is extremely common (thanks to `R` handling of literals). 
Perhaps type coercion can be left as a single (our of three originally suggested) switch in `#[extendr(...)]`? By default, coercion will happen (but limited, not like in base `R`), but a special switch can turn this mechanism off. 
However, if `Numeric` proposal is accepted, then all 'optimized' code can just work with `Numeric` and deal with the input in user code, and type coercion can be for all other use cases.

clauswilke:
> These conversions will be unlikely used by the consumers. 

I wasn't necessarily thinking that consumers would use them. Just that it's a logical, centralized place to implement conversions. You can implement `TryFrom` for `Rint` to `Rfloat` and vice versa and then all other API functions that need to do conversions would just use those traits and never have to think about details, conversion of `NA`s, etc.

Ilia-Kosenkov:
@clauswilke ,
Ok I agree, I will write this down.

Ilia-Kosenkov:
@clauswilke,
I have a question about conversions between `Rint` and `Rfloat` (and similar ones).
So, there are always safe casts, like `i32` -> `f64` -> `complex`, because the whole range of values of source type is representable by the target type. 
The reverse casts are mostly lossy. What should we do in case of a lossy cast? I like the approach of `{vctrs}`, where it errors if the value cannot be represented exactly, but this is not the case of basic `R` cast functions (`as.integer(1.5) == 1L`), nor the case of `Rust` `as` operator (`1.5 as i32 == 1`).

So, what would be the result of `std::convert::TryInto::<Rint>::try_convert(Rfloat(1.5))` ? 
- `Ok(Rint(1))`, the default behavior of `f64 as i32` or of `vctrs::allow_lossy_cast(vctrs::vec_cast(1.5, integer()))`
-  `Err(...)`, the default behavior of `vctrs::vec_cast(1.5, integer())`

clauswilke:
Now that I think about it, I'm not sure. R seems to implicitly convert, without error:
``` r
seq_len(3.24)
#> [1] 1 2 3
```

<sup>Created on 2021-08-19 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

Yet I feel in general that's a bad idea, because it can lead to unexpected and hard-to-find bugs. I'd rather error out as the default.

In general, I do think we need both lossy and lossless casts, so the only question is what should be the default. Since these functions will mostly be used to receive input via function calls, I think the lossless approach is safer.

Ilia-Kosenkov:
@clauswilke,
The lossless approach is the one similar to that of {vctrs}, which I personally prefer.

It will correctly cast `c(1, 2, 3)` to `Rint`, which would be the most frequent use case imo.

I will add this to the list

marshaling_rules.md
```diff
@@ -0,0 +1,148 @@
...
+Each vector can contain special `NA` values. None of the primitive types have built-in support for `NA` (including `f64`, which has notion of `NaN`, a different thing), so `R` treats one value from the range of allowed values as `NA`. For instance, `NA_integer_` is `i32::MIN`, which is `1i32 << 31 = -2147483648`. As a result, `x <- -2147483648L` results in an error in `R`.
+
+## `Rust` counterparts
+
+`R` objects passed to `Rust` require additional validation and transformation. Let us define the following types:
+- `struct Rint(i32)`
+- `struct Rbool(i32)`
+- `struct Rfloat(f64)`
+- `struct Rbyte(u8)`
+- `struct Rcmpl(f64, f64)`
```
clauswilke, line 68: I would spell it out. `Rcomplex`.

marshaling_rules.md
```diff
@@ -0,0 +1,148 @@
...
+ - `raw (u8)`
+ - `character (usize)` (collection of pointers to character arrays)
+
+Each vector can contain special `NA` values. None of the primitive types have built-in support for `NA` (including `f64`, which has notion of `NaN`, a different thing), so `R` treats one value from the range of allowed values as `NA`. For instance, `NA_integer_` is `i32::MIN`, which is `1i32 << 31 = -2147483648`. As a result, `x <- -2147483648L` results in an error in `R`.
+
+## `Rust` counterparts
+
+`R` objects passed to `Rust` require additional validation and transformation. Let us define the following types:
+- `struct Rint(i32)`
+- `struct Rbool(i32)`
```
yutannihilation, line 65: Will this replace `Bool`?

marshaling_rules.md
```diff
@@ -0,0 +1,148 @@
...
+
+- `Vec<i32>` triggers `NA` validation, altrep unfolding, and type coercion if compatible (so `1.0` converts to `1L`). Heavy on overhead and memory allocation, good for prototypes and testing things out.
+- `&[Rint]` is a responsible approach, which for array-based vectors of type `integer()` results in no allocations and no validation (zero overhead in the wrapper). **Unclear if type coercion should be performed**
+- `Integer` is an obscure wrapper of either `&[Rint]` or some `AltRepInt`. Acts as iterator with indexing capabilities, items are of type `Rint`, providing correct `NA` handling. Preferred way to handle vectors, mimics that of `{cpp11}`. **Unclear if type coercion should be performed**
+- (**Undecided**) `Numeric` is a discriminated union of `Integer | Real | Complex`. Accepts all numeric inputs, but leaves it up to the user to decipher what exactly was received from `R`. No runtime validation, no extra allocation, altreps remain altreps.
```
yutannihilation, line 15: I think complex is not a numeric in R.

``` r
is.numeric(1+0i)
#> [1] FALSE
```

<sup>Created on 2021-08-22 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

yutannihilation (commented):
minor comments

marshaling_rules.md
Ilia-Kosenkov: Agreed

marshaling_rules.md
Ilia-Kosenkov: Fair enough, we can exclude it.

Ilia-Kosenkov: However, even though it is not numeric, it seems to work nicely with basic `R` functions, including `mean`:

```R

mean(0:10 * 1i)
# > 0 + 5i
```

I would prefer to have `Numeric` include `Complex`.

Ilia-Kosenkov:
So far the API looks like this:
- (public) A set of primitive, binary compatible and `NA`-aware types, like `Rint`
- (private) A set of ALTREP wrappers for each type
- (public) Obscure iterators that contains either `&[Rt]` or `AltRepT`, with `<Item = Rt>` (`NA` - aware type)
- (public) Discriminated union `Numeric = Integer | Double`
- (public) Discriminated union `ComplexNumeric = Integer | Double | Complex` to support all numeric inputs (see e.g. `mean()`)

Type conversion traits are implemented for `NA`-aware primitives, iterator trait -- for obscure iterators, which under the hood either iterates over an array or goes through altrep.

We allow to implicitly convert 'smaller' primitive types to 'larger' (so `1L` can be passed as `&[Rfloat]`). We allow limited conversions from 'larger' to 'smaller' types (`1.0` can be passed as `&[Rint]`, but `1.5` will `panic!`). Strings, booleans and bytes have no conversions.

clauswilke:
@Ilia-Kosenkov Thanks for all your work on this! It's starting to come together.

Have you seen #266? I feel it's important to use names that indicate when something is an iterator or a proper wrapper and, in the latter case, represents a vector of multiple values.

Ilia-Kosenkov:
@clauswilke,
I explained in #266 the current naming convention I adopted, I suggest we come to some agreement there and then I will update all names here and we can proceed.

andy-thomason:
A quick note for ALTREP types:

ALTREP allows us to use deferred data which is calculated on the fly.
The compact range like `1:1000000000` is currently the most common instance of this.

Using a slice (as_integer_slice etc.) will call the `DATAPTR` function which
will allocate memory and manifest the values in the object. This is extremely
cheap for regular vectors, but very expensive for ALTREPs - and you may
run out of memory.

ALTREP has `GET_ELT` and `GET_REGION` functions which could be incorporated
into iterators. a `GET_REGION` of about 1k would make the evaluation quite efficient.

I'm leaning towards returning all vectors as ALTREPs as they have certain advantages:

1) We can use any data type as the backing store (u16, ULEB or fixed point for example).
2) We can wrap any movable closure or iterator into an ALTREP object.
3) ALTREP string vectors can be done with only one allocation.

Note that if you write a Vec<> as an input, a buffer needs to be allocated and data copied
to the buffer from R, whereas a slice has zero overhead.

You could write Rint like the Bool type (ie. four bytes), but i32 already supports the IsNA trait.

An ideal Int, Real, Complex or Numeric wrapper should be able to yield iterators that are
appropriate to the type.

We should maybe rename Character to Char and use "Character" for the string vector
wrapper with the same name for the RType.

Ilia-Kosenkov:
@andy-thomason,
> Note that if you write a Vec<> as an input, a buffer needs to be allocated and data copied
to the buffer from R, whereas a slice has zero overhead.

That is up to the user. `Vec<Rt>` mans allocation (all the time), `&[Rt]` should give just a slice from raw pointer if the vector is not altrep, and wrapper type like `Integers` should work with `&[Rt]` under the hood (again, if not wrapping altreps). I assume that we can control the lifetime of `&[Rt]` precisely so it does not leak outside of user code.

> You could write Rint like the Bool type (ie. four bytes), but i32 already supports the IsNA trait.

The trick here is to force people to not use built-in types (`i32` and `f64`) as-is. If one provides a wrapper, then conversion rules enforce all the validation. In short, if you get `Vec<i32>`, you may just read values and then end up with garbage results because you used `NA` in your computations. If instead, you get a `Vec<Rint>`, **you have no option** but to carefully check if it is `NA` or not. Same reason `Integers` should iterate over `Rints` (independently of underlying vector representation).
Just think of `i32` and `R`'s `integer()` as completely different types, with different value ranges:
- `{                 NA, -2147483647, ..., 0, ..., 2147483647 }`
- `{-2147483648, -2147483647, ..., 0, ..., 2147483647 }`

> I'm leaning towards returning all vectors as ALTREPs as they have certain advantages:

How would this work? What if the user code generates some (peudo-) random numbers and just needs to return a `Vec<Rfloat>` with concrete values?

yutannihilation:
> I'm leaning towards returning all vectors as ALTREPs

Oh, sounds interesting. It would be nice if we can build some nice interface.

> The trick here is to force people to not use built-in types (`i32` and `f64`) as-is.
> If instead, you get a `Vec<Rint>`, **you have no option** but to carefully check if it is `NA` or not.

Although I agree with you in that `IsNA` trait might not be very useful, I'm against this part. I believe users should have an option not to check for the sake of speed, as the `NA` check might be done externally. I think it's wrappers like `Integers` that can provide both something like:

* `Integers::to_slice()`, with `NA`-validation
* `Integers::to_slice_unchecked()`, without `NA`-validation
* `Integers::get()`, with `NA`-validation (I have no idea how to distinguish between `NA` and the end of the vector)
* `Integers::get_unchecked()`, without `NA`-validation

Ilia-Kosenkov:
@yutannihilation,
I agree about wrappers, we can do whatever we want there. However, how would you handle `to_slice*` if the underlying representation is ALTREP? 

I am starting to think we need to limit input types. Perhaps leave a generic `Vec<T>` with guaranteed allocation, which unfolds altrep and coerces values where applicable. This is best for quick prototyping when you just need to work with data in the memory and do not want to handle edge cases yourself. The second option is strongly typed wrapper types, similar to that of `{cpp11}`. As mentioned earlier, such wrapper type `Wrapper<T>`  contains either  `&[T]`  or `AltRep<T>` , exposing public interface. Here we can have some `unsafe_try_get_slice()`, which succeeds if there is slice to obtain.

An interesting observation is that `{cpp11}` wrapper types are strongly typed, so

``` r
cpp11::cpp_function("void cpp_test(doubles x) {}")

cpp_test(1.0)
cpp_test(1L)
#> Error: Invalid input type, expected 'double' actual 'integer'
```

If we decide to make our wrappers strongly typed (no coercion), they will be zero-allocation in all cases.
And when one wants to handle both ints and floats, one can use `Numerics`, which is either `Integers` or `Doubles`, each of which in turn wraps either altrep or slice.


So, to sum up, the simplest solution could be:
- `Vec<T>` for simplicity, education & prototyping. Always allocates, coerces types, checks for `NA` if `T` is not `NA` aware
- `Wrapper<T>` for all other cases, strongly typed, no allocation, `panic!`s if type mismatch.

<details>
<summary>Imagining how wrapper types may work</summary>

```rust
struct AltRepImpl<'a, T>(std::marker::PhantomData<&'a T>) where T : Sized;

enum Wrapper<'a, T> where T : Sized {
    AltRep(AltRepImpl<'a, T>),
    PlainVector(&'a [T])
}

type Integers<'a> = Wrapper<'a, i32>;
type Doubles<'a> = Wrapper<'a, f64>;

enum Numerics<'a> {
    Int(Integers<'a>),
    Dbl(Doubles<'a>)
}


fn check_input<'a>(x : Numerics<'a>) {
    let _ = match x {
        Numerics::Int(Integers::AltRep(_altrep)) => -1,
        Numerics::Int(Integers::PlainVector(int_vec)) => int_vec[0],
        Numerics::Dbl(Doubles::AltRep(_altrep)) => -2,
        Numerics::Dbl(Doubles::PlainVector(dbl_vec)) => dbl_vec[0] as i32
    };
}

```

</details>

yutannihilation:
> how would you handle `to_slice*` if the underlying representation is ALTREP?

We just can allocate by calling `REAL` or `INTEGER`, just as we currently do in `as_typed_slice()`. Sorry, I think I don't get your point.

Ilia-Kosenkov:
> We just can allocate by calling REAL or INTEGER, just as we currently do in as_typed_slice(). Sorry, I think I don't get your point.

I was trying to say there is no benefit in getting out a slice if we first need to allocate, which is the case of ALTREP. For regular plain vectors, getting a slice is basically a pointer cast. So (imo) getting a slice should be conditional, see example in `{cpp11}`:
https://github.com/r-lib/cpp11/blob/1b4beb25f56c8dca6e0448d5dc55fb64fddab5d1/inst/include/cpp11/integers.hpp#L34-L41

yutannihilation:
Ah, I see your point. I think it's nice the wrappers have that conditional version of method as well, but I still want a version that allocates as well, just like the R API has both `DATAPTR` and `Dataptr_or_null`. As a user, I don't want to care whether the vector is ALTREP or not when it contains just a small number of elements.

clauswilke:
I think all of this conversation shows that writing wrappers is the right approach. We can then add any number of accessor methods with different performance and safety guarantees. We don't necessarily have to settle on the exact set of available accessor methods today.

yutannihilation:
Yeah, I guess so. If we want a finer control over conversion, it can easily go beyond what `#[extendr]` can handle solely, and we'll need to rely on other explicit conversions. That was one of the motivations why I started writing #244 to cover all sorts of conversions, not just `#[extendr]`.

Ilia-Kosenkov:
@clauswilke,
I assume we should use `Strings` instead of `Characters`?

clauswilke:
Yes, I think `Strings` is preferable, following your argument that `Characters` sounds like a collection of characters (i.e., a single string).

Ilia-Kosenkov:
Once again, a quick summary:
So far we seem to agree to encourage two models:
1. Easy. Great for prototyping and learning. Accept arguments like `Vec<i32>`, get all conversions, allocations, `NA`-checks in the extendr-generated boilerplate code, but at runtime (with all the overhead).
2. Safe. Embrace `{cpp11}` conventions. Accept `Doubles` to get only `doubles()` (plain or altrep). Because of the natural support of discriminated unions, we can have a nice, strict, allocation-free, and verbose system. Want an `integer()` or a `double()`? Accept `Numerics` and use match syntax. In the [comment above](https://github.com/extendr/extendr/pull/261#issuecomment-904681673) I have a hidden code block that shows how such types can be used.
3. Forbid explicit slices (like `&[Rt]`). While slices **can** be obtained in the proposed model (e.g., `fn Integers::get_slice(&self) -> Result<&[Rint], ()>`), such function argument is useless in the general case. It is hard to predict what type of input will `R` provide, so if the underlying type is different from `Rt`, or if it is an altrep, such function call will fail. A safer approach is to accept a wrapper-iterator and carefully check what input type is passed and if it is an altrep.

marshaling_rules.md
```diff
@@ -0,0 +1,165 @@
...
+This can be the preferred solution when dealing with `R` plain vectors.
+
+For each supported primitive type `T` `Rt` would be its minimal wrapper. E.g., for `T = i32`, `Rt = Rint`.
+`extendr` prefers `Rt` over `T` types. Parameters that use `T`-derived types will require runtime `NA` validation, introducing implicit overhead.
+
+Type conversion traits for `Rt` are:
+- `Into<Option<T>>` (this is always a valid conversion)
+- `TryInto<T>`, errors on `NA`
+- `TryFrom<T>`, errors if provided argument equals to the value reserved for `NA`
+- `TryFrom<Option<T>>` for the same reason, as `Some(i32::MIN)` is invalid `Rint` value (instead, `None` should be used)
```
clauswilke, line 74: I don't entirely understand what you're trying to say here about `TryFrom<Option<T>>`. For the same reason what? Could you add 1-2 sentences to explain?

clauswilke:
I'm very happy with this latest version. Except for the one minor issue I pointed out with a comment, this is good to go as far as I am concerned.

marshaling_rules.md
Ilia-Kosenkov: Done, rephrased it in the last commit.

clauswilke: Thanks, I get it now.

clauswilke:
@andy-thomason You approved this PR, but could you read over the latest version and let us know whether you agree with the proposed naming scheme or have any concerns? Thanks!

yutannihilation (approved):
Looks good to me! Thank you for handling this difficult discussion.

Let me leave two minor comments.

* I'm still not sure what's the ideal way to deal with `complex()`. We might want to wrap it with some type e.g. [num::complex::Complex](https://autumnai.github.io/cuticula/num/complex/struct.Complex.html), putting behind a feature gate, but let's discuss this in future. I think the current proposal is good to go.
* "unfold" is fine (so you don't need to fix anything), but I think "expand" is more common term in the context of ALTREP.

clauswilke:
Since this is getting pretty solid and useful, I'm wondering whether we should merge into main rather than into a branch. Maybe we could create a top-level directory called `project_architecture` or `design_specifications` or similar and move this document there. We could also move the maintainers guide there.

To keep the chain of discussion and commits, though, we could also merge into the branch first and then immediately merge the branch.

Ilia-Kosenkov:
@yutannihilation,
I agree with the choice of 'expand' instead of 'unfold',  I will replace it.
Speaking of `complex()`, am I correct that `num` is actually a separate crate? If we do not reference it already, we may want to use a feature-guard, similar to the case of `ndarray`. At least, this was my intention.

@clauswilke,
I would prefer to merge it into current branch and then create another PR to master.

clauswilke:
> I would prefer to merge it into current branch and then create another PR to master.
Yes, let's do that.

yutannihilation:
>  Speaking of `complex()`, am I correct that `num` is actually a separate crate? If we do not reference it already, we may want to use a feature-guard, similar to the case of `ndarray`. At least, this was my intention.

Yes, that's what I meant. Like matrix, users might need some crate like `num` to do mathematical operations with complex numbers for efficiency. But, since complex is not what everyone uses every time, it sounds a good idea to use a feature-guard.

(Disclaimer is that I've never used complex in a serious way, so I'd like to hear from users who are experienced in using complex in their algorithms.)

Ilia-Kosenkov:
@clauswilke, @yutannihilation, @andy-thomason, and everyone interested in the topic,
Let's do one last and quick check on the state of this proposal before I merge it. 
As discussed above, the intention is to merge it to a separate branch and then merge that branch to master.

clauswilke:
I suggest you merge to the branch now, and then we can review (and possibly fine-tune) there before merging to master. This makes it easier to read the markdown and to catch any typos or other minor issues. Also this will allow us to merge in @yutannihilation's text from #244.

Ilia-Kosenkov:
@clauswilke, sounds fair, will do it now.

PR #264, opened 2021-08-21
Int, Real conversions follow-up

brendanrbrown:
A requested follow-up to #258.

Maintainers should feel free to mark this PR as draft (or I can) if deemed more appropriate: 

The implementation is very likely to change once the discussion in #261, #251 solidifies. In particular, wrapping types such as `Rint` if implemented would be used here. Better NA handling and error types in `TryFrom<Robj> for Int/Real` would also follow, which I think is needed.

### DIFF
--- a/extendr-api/src/iter.rs
+++ b/extendr-api/src/iter.rs
@@ -83,6 +83,30 @@ pub type NamedListIter = std::iter::Zip<StrIter, ListIter>;
 /// ```
 pub type Int = SliceIter<i32>;
 
+impl TryFrom<Real> for Int {
+    type Error = Error;
+
+    fn try_from(it_real: Real) -> Result<Self> {
+        // collect_robj does not handle Result<Robj>, so collect to Vec first.
+        let mut v = Vec::with_capacity(it_real.len());
+
+        // mirrors relevant part of impl_try_from_scalar_integer
+        for r in it_real {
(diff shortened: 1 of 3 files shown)


clauswilke:
Thanks! I think this PR highlights that types such as `Rint` are needed and should probably be implemented first. (Or you could do so as part of this PR.) There's no good reason to duplicate conversion code in multiple places, as we would be doing right now. I'll open a separate issue on this topic and link here.

clauswilke:
I have filed issues #265 and #266 that cover relevant discussions we should have before proceeding. I think we need at least consensus on #265 before we can finalize this PR.

brendanrbrown:
> I think we need at least consensus on #265 before we can finalize this PR.

Entirely agree, and as you suggest this PR is meant in part to point out some cases in which these changes are needed. Thanks for filing these issues and for marking this PR as draft.

brendanrbrown:
It looks to me as though this PR can be closed, in light of #284 and #274, which seem to fill the holes discussed above. If others agree I (or they) can close it.

This does leave open the question of what to do with the remaining `Real` and `Int` types -- but that is for a separate issue or PR I think.

clauswilke:
Yes, agreed. Please file issues for any items that you can think of that should be addressed. Thanks!

PR #268, opened 2021-08-27
Basic S4 support.

andy-thomason:
This PR adds the ability to create S4 objects and read and write their slots.

### DIFF
--- a/extendr-api/src/wrapper/s4.rs
+++ b/extendr-api/src/wrapper/s4.rs
@@ -1,0 +1,153 @@
+//! S4 class support.
+
+use super::*;
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct S4 {
+    pub(crate) robj: Robj,
+}
+
+impl S4 {
+    /// Make a new S4 type.
+    ///
+    /// Example:
+    /// ```
+    /// use extendr_api::prelude::*;
+    ///
+    /// test! {
+    ///     let robj = S4::new();
+    ///     assert_eq!(robj.rtype(), RType::S4);
+    /// }
+    /// ```
+    pub fn new() -> Self {
+        let robj = single_threaded(|| unsafe { new_owned(Rf_allocS4Object()) });
+        Self { robj }
+    }
+
+    /// Get a named slot from a S4 object.
+    ///
+    /// Example:
+    /// ```
+    /// use extendr_api::prelude::*;
+    ///
+    /// test! {
+    ///     let mut robj = S4::new();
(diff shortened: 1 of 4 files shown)


extendr-api/src/wrapper/s4.rs
```diff
@@ -0,0 +1,153 @@
...
+    ///
+    /// test! {
+    ///     let robj = S4::new();
+    ///     assert_eq!(robj.rtype(), RType::S4);
+    /// }
+    /// ```
+    pub fn new() -> Self {
+        let robj = single_threaded(|| unsafe { new_owned(Rf_allocS4Object()) });
+        Self { robj }
+    }
```
yutannihilation, line 25: Is it legal to use an S4 object without specifying the class? Sorry, I almost know nothing about S4, especially how to handle it via C API.

Rcpp does require the class.

https://github.com/RcppCore/Rcpp/blob/423a2ed79bfda9b4e5e06b5dc1c646b02d60eb42/inst/include/Rcpp/S4.h#L57-L62

andy-thomason: Good point.

Likewise, I'm not proposing using S4 classes, but we should at least be able to read/write their slots.

There is a lot of redundancy in the S4 data model. The S4 flag, the S4SXP type and the class
attribute may all need to be there together.

yutannihilation: > we should at least be able to read/write their slots.

I see, agreed. Then, how about removing `new()` method (or at least `pub`)? Tests can use an S4 object created by `R!()` macro.

andy-thomason:
I'm experimenting with following the same-ish model as S4.

The problem with R_do_MAKE_CLASS is that it may throw an R error of the class does not exist.

We also need some way of creating classes...

It may be simplest just to lose the new for now.

andy-thomason:
We needed some way to make S4 classes for tests, so i wrapped setClass() and new()
using R!.

Not the finest solution, but it should work.

yutannihilation (approved):
Thanks for investigating. This looks good for now. Could you add some TODO comment (or file an issue) for future improvment?

The CI failure on Windows + R-devel can be ignored. I'll take a look how to tweak the CI setting for retired 32-bit.

andy-thomason:
Shall I merge, despite the "tests_with_unsuccessful_outcomes" build failure?

clauswilke:
Yes, we should address that problem separately, probably by enforcing single-threaded execution as you pointed out in the relevant issue.

Issue #54, opened 2020-11-29
Clippy warnings


jbabyhacker:
There are numerous clippy warnings. Most are related to publicly exposed unsafe functions. I would like to go an fix this, but what is the guidelines in the project regarding public unsafe functions? Allow them or wrap them so no public functions are unsafe?

clauswilke:
Are these extendr-api functions or libR-sys functions? I think libR-sys needs publicly exposed unsafe functions, because that's its entire purpose. extendr-api probably shouldn't expose any unsafe functions.

jbabyhacker:
These are extendr-api functions. Unsafe makes sense for libR-sys.

jbabyhacker:
So I looked into this a little deeper. Because of the separation between extendr-api and extendr-macros, and the use of raw pointers, the only way to not have unsafe code publicly exposed would be to combine the two crates. Not sure if you want to do that.

clauswilke:
Rust procedural macros need to live in their own crate. Hence, there's no way to combine the two crates into one.

yutannihilation:
Sorry to ask a Rust newbie question here... How can I see these "Clippy warnings"? Is it just running `cargo clippy` on console? If so, I don't see any warnings. Do I need some additional linting configurations?

yutannihilation:
> I don't see any warnings

Now I see the warnings.

https://gist.github.com/yutannihilation/485ba77d2584784fd339c7979a786b22

yutannihilation:
I'm wondering whether we should monitor clippy warnings on CI or not.

As clippy's warnings may contain false-positives, I now feel the warnings should not make CI fail (the errors might, though). In terms of this, this GitHub Action seemed very useful in that this adds comments instead of failure.

https://github.com/actions-rs/clippy-check

But, its README says there's one big limitation:

> Due to token permissions, this Action WILL NOT be able to post clippy annotations for Pull Requests from the forked repositories.

Due to this, the action suppresses clippy warnings for pull requests from others, which makes many warnings unnoticed. So, I now feel it's not worth checking clippy warnings, considering the complexity.

Any thoughts?

clauswilke:
I agree, let's not have CI fail due to clippy warnings.

It would be good, though, if the clippy warnings could at least be output somewhere so we can monitor them easily if we want to. We have a lot of warnings that probably should be fixed, and I had never seen them before you posted them.

yutannihilation:
I see. I'll look for other options.

PR #274, opened 2021-08-30
Altrep Integers class.

andy-thomason:
This is a sample for an Integers class that uses Altrep
methods to handle very large vectors.

There are no iterators at present as we would need to implement
Rint first.

The USP of this is the ability to make vectors from iterators without
allocating memory.

We support elt(), get_region(), is_sorted() and no_na() correctly.

We do not have access to altrep min() and max() since these were removed
from Rinternals.h

Suggestions on a postcard...

### DIFF
--- a/extendr-api/src/scalar/rint.rs
+++ b/extendr-api/src/scalar/rint.rs
@@ -208,3 +208,18 @@ gen_unnop!(
     |lhs: i32| Some(!lhs),
     "Logical not a Rint value, overflows to NA."
 );
+
+impl std::iter::Sum for Rint {
+    /// Sum an integer iterator over Rint.
+    /// Yields NA on overflow of NAs present.
+    fn sum<I: Iterator<Item = Rint>>(iter: I) -> Rint {
+        iter.fold(Rint::from(0), |a, b| a + b)
+    }
+}
+
+impl PartialEq<i32> for Rint {
+    /// Compare a Rint with an integer. NA always fails.
+    fn eq(&self, other: &i32) -> bool {
+        !self.is_na() && self.0 == *other
+    }
+}
--- a/extendr-api/src/scalar/rint.rs
+++ b/extendr-api/src/scalar/rint.rs
@@ -169,6 +169,7 @@ macro_rules! gen_unnop {
     };
 }
 
+// Generate binary ops for +, -, * and /
 gen_binop!(
     Add,
     add,
@@ -193,6 +194,8 @@ gen_binop!(
     |lhs: i32, rhs| lhs.checked_div(rhs),
     "Divide two Rint values or an option of i32, overflows to NA."
 );
+
+// Generate unary ops for -, !
 gen_unnop!(
     Neg,
     neg,
--- a/extendr-api/src/scalar/rint.rs
+++ b/extendr-api/src/scalar/rint.rs
@@ -1,4 +1,5 @@
-use super::super::IsNA;
+use crate::*;
+use std::convert::TryFrom;
 use std::ops::{Add, Div, Mul, Neg, Not, Sub};
 
 /// Rint is a wrapper for i32 in the context of an R's integer vector.
@@ -15,6 +16,11 @@ impl Rint {
     /// Construct a NA Rint.
     pub fn na() -> Self {
         Rint(i32::MIN)
+    }
+
+    /// Get an integer or i32::MIN for NA.
+    pub fn inner(&self) -> i32 {
+        self.0
     }
 }
 
@@ -221,5 +227,58 @@ impl PartialEq<i32> for Rint {
     /// Compare a Rint with an integer. NA always fails.
     fn eq(&self, other: &i32) -> bool {
         !self.is_na() && self.0 == *other
+    }
+}
+
+impl TryFrom<Robj> for Rint {
+    type Error = Error;
+
+    fn try_from(robj: Robj) -> Result<Self> {
+        // Check if the value is a scalar
+        match robj.len() {
+            0 => return Err(Error::ExpectedNonZeroLength(robj)),
+            1 => {}
+            _ => return Err(Error::ExpectedScalar(robj)),
+        };
+
+        // Check if the value is not a missing value
+        if robj.is_na() {
+            return Ok(Rint::na());
+        }
+
+        // If the conversion is int-to-int, check the limits. This
+        // needs to be done by `TryFrom` because the conversion by `as`
+        // is problematic when converting a negative value to unsigned
+        // integer types (e.g. `-1i32 as u8` becomes 255).
+        if let Some(v) = robj.as_integer() {
+            if let Ok(v) = Self::try_from(v) {
+                return Ok(v);
+            } else {
+                return Err(Error::OutOfLimits(robj));
+            }
+        }
+
+        // If the conversion is float-to-int, check if the value is
+        // integer-like (i.e., an integer, or a float representing a
+        // whole number). This needs to be down with `as`, as no
+        // `TryFrom` is implemented for float types. `FloatToInt` trait
+        // might eventually become available in future, though.
+        if let Some(v) = robj.as_real() {
+            let result = v as i32;
+            if (result as f64 - v).abs() < f64::EPSILON {
+                return Ok(Rint::from(result));
+            } else {
+                return Err(Error::ExpectedWholeNumber(robj));
+            }
+        }
+
+        Err(Error::ExpectedNumeric(robj))
+    }
+}
+
+impl From<Rint> for Robj {
+    /// Convert am Rint into an robj.
+    fn from(value: Rint) -> Self {
+        Robj::from(value.0)
     }
 }
--- a/extendr-api/src/scalar/rint.rs
+++ b/extendr-api/src/scalar/rint.rs
@@ -1,0 +1,207 @@
+use super::super::IsNA;
+use std::ops::{Add, Div, Mul, Neg, Not, Sub};
+
+/// Rint is a wrapper for i32 in the context of an R's integer vector.
+///
+/// Rint can have a value between i32::MIN+1 and i32::MAX
+///
+/// The value i32::MIN is used as "NA".
+///
+/// Rint has the same footprint as an i32 value allowing us to use it in zero copy slices.
+#[derive(PartialEq, Eq)]
+pub struct Rint(pub i32);
+
+impl Rint {
+    /// Construct a NA Rint.
+    pub fn na() -> Self {
+        Rint(i32::MIN)
+    }
+}
+
+impl Clone for Rint {
+    fn clone(&self) -> Self {
+        Self(self.0)
+    }
+}
+
+impl Copy for Rint {}
+
+impl IsNA for Rint {
+    /// Return true is the is a NA value.
+    fn is_na(&self) -> bool {
+        self.0 == std::i32::MIN
+    }
+}
+
+impl From<i32> for Rint {
+    /// Construct a Rint from an integer.
+    /// i32::MIN gives an NA.
+    fn from(v: i32) -> Self {
+        Self(v)
+    }
+}
+
+impl From<Option<i32>> for Rint {
+    /// Construct an Rint from an optional integer.
+    /// None or Some(i32::MIN) gives an NA.
+    fn from(v: Option<i32>) -> Self {
+        if let Some(v) = v {
+            v.into()
+        } else {
+            Rint::na()
+        }
+    }
+}
+
+impl From<Rint> for Option<i32> {
+    /// Convert an Rint to an optional integer.
+    /// NA gives None.
+    fn from(v: Rint) -> Self {
+        if v.is_na() {
+            None
+        } else {
+            Some(v.0)
+        }
+    }
+}
+
+impl std::fmt::Debug for Rint {
+    /// Debug format an Rint.
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let z: Option<i32> = (*self).into();
+        if let Some(val) = z {
+            write!(f, "{}", val)
+        } else {
+            write!(f, "na")
+        }
+    }
+}
+
+macro_rules! gen_binop {
+    ($opname1 : ident, $opname2: ident, $expr: expr, $docstring: expr) => {
+        impl $opname1<Rint> for Rint {
+            type Output = Rint;
+
+            #[doc = $docstring]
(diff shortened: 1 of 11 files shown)


clauswilke:
Thanks for your efforts! Two comments:

- What about the non-altrep variant? I believe the idea was that `Integers` would be an opaque wrapper that can represent both altrep and regular vectors. @Ilia-Kosenkov suggested `AltIntegers` for an altrep-only wrapper for integers: https://github.com/extendr/extendr/blob/03d06173fd457c338bfe371a9d9f211c44f32051/marshaling_rules.md#naming-convention So maybe implement `AltIntegers` first and then use that to implement `Integers`?
- If `Rint` is needed for feature completeness should that be implemented first or at the same time? I assume it's relatively lightweight.

andy-thomason:
This should handle both the ALTREP and regular vectors.

The methods we call like get_region and elt should also work
for regular vectors.

When constructing vectors, I'm chosing to use regular vectors
for small ones and ALTVEC for larger ones. This is because there
is some overhead in creating the class and vector and we
need to generate the class every time.

The regular conversion rules should accept anything with INTSXP
as input. The macro in mod.rs converts from Robj to Integers and
vice versa.

extendr-api/src/wrapper/mod.rs
```diff
@@ -157,6 +159,8 @@ make_conversions!(Altrep, ExpectedAltrep, is_altrep, "Not an Altrep type");
 
 make_conversions!(S4, ExpectedS4, is_s4, "Not a S4 type");
 
+make_conversions!(Integers, ExpectedInteger, is_integer, "Not an integer type");
```
andy-thomason, line 162: This makes conversions...

clauswilke:
> The methods we call like get_region and elt should also work
for regular vectors.

I see, thanks!

extendr-api/src/scalar/rint.rs
```diff
@@ -0,0 +1,207 @@
...
+        let z: Option<i32> = (*self).into();
+        if let Some(val) = z {
+            write!(f, "{}", val)
+        } else {
+            write!(f, "na")
+        }
+    }
+}
+
+macro_rules! gen_binop {
```
clauswilke, line 80: Could you add just one or two sentences here commenting what this does? E.g., something like: "Programmatically derive binary operators"? Also, if this hardcodes the `Rint` type maybe call it `gen_binop_Rint`, since we'll need something similar for `Rfloat`?

extendr-api/src/scalar/rint.rs
```diff
@@ -0,0 +1,207 @@
...
+                        return Rint::from(res);
+                    }
+                }
+                Rint::na()
+            }
+        }
+    };
+}
+
+macro_rules! gen_unnop {
```
clauswilke, line 136: Same comments as `gen_binop`.

clauswilke:
The `Rint` prototype looks good to me. We'll probably want to add `TryFrom` traits eventually, but it's a good starting point.

extendr-api/src/wrapper/integers.rs
```diff
@@ -0,0 +1,101 @@
...
+                robj
+            };
+            Self { robj }
+        })
+    }
+
+    /// Get a single element from the vector.
+    pub fn elt(&self, index: usize) -> i32 {
+        unsafe { INTEGER_ELT(self.get(), index as R_xlen_t) }
+    }
```
yutannihilation, line 82: What I've been arguing on #261 is that this kind of method should be named as `elt_unchecked()` (and probably marked as `unsafe` function), and implement another `elt()` that does validation. We might need another variant that returns `Rint`.

yutannihilation (commented):
Thanks, this pull request is a really nice starting point for discussion.

extendr-api/src/wrapper/integers.rs
clauswilke: If we mark it as `unsafe` we can't have it part of the public API though, can we? (Actually, I still think we will eventually need unsafe functions in the public API for maximum usability.)

yutannihilation: To be clear, I think this function `elt_unchecked()` is "safe" enough to be included in the public API, in the sense that it won't destroy some memory region or panic suddenly. But, I thought it should be labelled as `unsafe` when it's users' responsibility to ensure the input is valid like this one:

https://doc.rust-lang.org/std/primitive.f32.html#method.to_int_unchecked

I'm yet to figure out the semantics of `unsafe`, so I might be wrong here.

clauswilke: I was just commenting on the fact that we recently removed an unsafe public function because some linter complained, but I think a library such as extendr needs unsafe public functions. So I'm on board with marking this as unsafe.

yutannihilation: Ah, https://github.com/extendr/extendr/issues/54? But I didn't succeed to reproduce the warning at that time (https://github.com/extendr/extendr/issues/54#issuecomment-757086375).

This might be the one. This indicates it's fine to have `unsafe` public function as long as it has `Safety` section.

```
warning: unsafe function's docs miss `# Safety` section
  --> extendr-api/src/wrapper/integers.rs:80:5
   |
80 | /     pub unsafe fn elt(&self, index: usize) -> i32 {
81 | |         INTEGER_ELT(self.get(), index as R_xlen_t)
82 | |     }
   | |_____^
   |
   = note: `#[warn(clippy::missing_safety_doc)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#missing_safety_doc
```

andy-thomason: Yes, I don't think elt will ever be efficient, so returning a validated Result always may be the best bet.

It should probably return a Result<Rint>, also.

get_region can do all its checks once for a number of elements and call "skip" on the iterator
which may take time depending on optimisation levels. It is better for performing vector operations.

We might want a get_region_repeat which repeats the vector when doing operations on different size vectors.

andy-thomason: I'm keen to implement parallel region operations, also.

clauswilke: What's going to be the simple interface when somebody doesn't care about performance and has a smallish vector. An iterator?

andy-thomason: Just give it an iterator like this:

```
let x : Vec<i32> = ...
let y : Vec<i32> = ...
let plus_one = Integers::from_values(x.iter().map(|&x| x + 1);
let xplusy = Integers::from_values(x.iter().zip(y.iter())map(|(&x, &y)| x + y);

```

I'd need to check them.

clauswilke: Thanks, this looks good. I was thinking about the other direction, though, e.g. add two R vectors and store the result in a rust vector.

clauswilke: Yes, correct.

andy-thomason: The `FromIterator` is what we need here.

https://doc.rust-lang.org/std/iter/trait.FromIterator.html

andy-thomason: Done.

extendr-api/src/wrapper/integers.rs
```diff
@@ -98,4 +106,57 @@ impl Integers {
...
+    /// Return an iterator for a non-altrep object.
+    /// Suitable for short vectors.
+    /// ```
+    /// use extendr_api::prelude::*;
+    /// test! {
+    ///     let vec = Integers::from_values(0..3);
+    ///     assert_eq!(vec.iter().cloned().sum::<i32>(), 3);
+    /// }
+    /// ```
+    pub fn iter(&self) -> impl Iterator<Item = &i32> {
```
clauswilke, line 119: Apologies for being nitpicky: Shouldn't the item type be `Rint`? In the general case, we don't know that we can return an `i32` for sure (could be `NA`).

yutannihilation: I thought this would return `SliceIter<T>`. Or, is the idea that doing `NA`-validation beforehand at once, which is similar to this comment above? If so, `i32` might make sense.

> do all its checks once for a number of elements

andy-thomason: Good point...

andy-thomason: I've implemented Rint where possible.

I'm in two minds about get_region as we would like to support SIMD arithmetic but can do this too if required.

andy-thomason: The SliceIter should probably go. Its only useful feature is that the lifetime
is effectively static.

extendr-api/src/scalar/rint.rs
```diff
@@ -0,0 +1,284 @@
...
+gen_unnop!(
+    Not,
+    not,
+    |lhs: i32| Some(!lhs),
+    "Logical not a Rint value, overflows to NA."
+);
+
+impl std::iter::Sum for Rint {
+    /// Sum an integer iterator over Rint.
+    /// Yields NA on overflow of NAs present.
```
yutannihilation, line 220: > Yields NA on overflow of NAs present.

If this sentence means to say no `NA`-validation is needed because of this, I don't think it's true. (e.g. `c(-1, NA)`)

extendr-api/src/scalar/rint.rs
```diff
@@ -0,0 +1,284 @@
...
+        if robj.is_na() {
+            return Ok(Rint::na());
+        }
+
+        // If the conversion is int-to-int, check the limits. This
+        // needs to be done by `TryFrom` because the conversion by `as`
+        // is problematic when converting a negative value to unsigned
+        // integer types (e.g. `-1i32 as u8` becomes 255).
+        if let Some(v) = robj.as_integer() {
+            if let Ok(v) = Self::try_from(v) {
```
yutannihilation, line 254: I think we can skip this `try_from` in this case. `Rint` is 32-bit, so it should never overflow.

extendr-api/src/wrapper/integers.rs
```diff
@@ -0,0 +1,163 @@
...
+                }
+                robj
+            };
+            Self { robj }
+        })
+    }
+
+    /// Get a single element from the vector.
+    /// Note that this is very inefficient in a tight loop.
+    pub fn elt(&self, index: usize) -> Rint {
```
yutannihilation, line 89: Sorry for repeating the same comment, but I'd argue `elt()` should do `NA`-validation and return `i32` or (`Result<i32>`).

andy-thomason: In this case, `Rint` itself does the result validation as it has a conversion to `Option<i32>`.

The `i32` value is guaranteed to not be `i32::MIN`.

extendr-api/src/scalar/rint.rs
andy-thomason: /of/or/

extendr-api/src/scalar/rint.rs
andy-thomason: Yes.

extendr-api/src/scalar/rint.rs
yutannihilation: Ah, I see. So, anyway we need `NA`-validation here, right?

extendr-api/src/wrapper/integers.rs
yutannihilation: Hmm, okay. I hoped this returns `Option<i32>` directly, but I guess I'm a minority here.

andy-thomason: If you have a strong opinion, we can change it. elt() is not that performance critical.

extendr-api/src/wrapper/integers.rs
```diff
@@ -0,0 +1,163 @@
...
+        unsafe { INTEGER_ELT(self.get(), index as R_xlen_t).into() }
```
yutannihilation, line 90: In the case of non-ALTREP, cpp11 accesses to the underlying memory directly without `INTEGER_ELT`. Can we do the same thing? Doesn't this matter much on the performance?

https://github.com/r-lib/cpp11/blob/1b4beb25f56c8dca6e0448d5dc55fb64fddab5d1/inst/include/cpp11/integers.hpp#L29-L32

extendr-api/src/wrapper/integers.rs
```diff
@@ -0,0 +1,163 @@
...
+            INTEGER_GET_REGION(self.get(), index as R_xlen_t, dest.len() as R_xlen_t, ptr);
+        }
+    }
+
+    /// Return TRUE if the vector is sorted, FALSE if not, or NA_BOOL if unknown.
+    pub fn is_sorted(&self) -> Bool {
+        unsafe { INTEGER_IS_SORTED(self.get()).into() }
+    }
+
+    /// Return TRUE if the vector has NAs, FALSE if not, or NA_BOOL if unknown.
```
yutannihilation, line 106:
```suggestion
    /// Return TRUE if the vector has no NAs, FALSE if not, or NA_BOOL if unknown.
```

extendr-api/src/wrapper/integers.rs
yutannihilation: I don't have a strong opinion. I think I'll be fine with `Rint`.

andy-thomason:
Just so you know that we all very much respect your views.

yutannihilation:
Thanks, I just feel I don't have any idea worth arguing about there for now :) Let's move forward.

extendr-api/src/scalar/rint.rs
yutannihilation: Okay, I thought this would return `i32::MIX - 1`, but I found this is handled correctly thanks to `<Rint>::checked_sub()`. Sorry for my misunderstanding!

``` r
rextendr::rust_source(code = r"(
use extendr_api::scalar::Rint;

fn do_sum(x: Integers) -> Rint {
  x.iter().sum()
}
)", patch.crates_io = list(`extendr-api` = list(git = "https://github.com/extendr/extendr"))
)
#> ℹ build directory: '/tmp/Rtmp7RG5fJ/file77e25279e1bf0'
#> ✓ Writing '/tmp/Rtmp7RG5fJ/file77e25279e1bf0/target/extendr_wrappers.R'.

do_sum(c(NA_integer_, -1L))
#> [1] NA
```

<sup>Created on 2021-09-12 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

PR #275, opened 2021-08-31
ndarray improvements

multimeric:
Changes:
* Bump ndarray. This is needed because the current 0.13.x version that we are using is not compatible with projects using the more modern 0.15.x versions, and interop between arrays from different versions of the library will fail in confusing ways
* Implement `From<ArrayBase<S, D>> for Robj` ie conversion from any ndarray to an Robj, with tests

Outstanding questions:
* ~~How can we actually compare R objects in Rust? My tests are failing I assume because the pointers are not the same, even though the string representation seems identical~~
* ~~Why do the arrays created using `R!` not have any classes set when I `a.get_attrib(class_symbol())`?~~

### DIFF
--- a/extendr-api/src/robj_ndarray.rs
+++ b/extendr-api/src/robj_ndarray.rs
@@ -6,8 +6,8 @@ use crate::prelude::{class_symbol, dim_symbol};
 use crate::*;
 
 impl<'a, T> FromRobj<'a> for ArrayView1<'a, T>
-    where
-        Robj: AsTypedSlice<'a, T>,
+where
+    Robj: AsTypedSlice<'a, T>,
 {
     /// Convert an R object to a `ndarray` ArrayView1.
     fn from_robj(robj: &'a Robj) -> std::result::Result<Self, &'static str> {
@@ -58,10 +58,10 @@ make_array_view_2!(f64, "Not a floating point matrix.");
 //     }
 // }
 impl<A, S, D> From<ArrayBase<S, D>> for Robj
-    where
-        S: Data<Elem=A>,
-        A: Copy + ToVectorValue,
-        D: Dimension,
+where
+    S: Data<Elem = A>,
+    A: Copy + ToVectorValue,
+    D: Dimension,
 {
     fn from(arr: ArrayBase<S, D>) -> Self {
         let dims: Vec<i32> = arr.shape().iter().map(|x| *x as i32).collect();
@@ -159,15 +159,19 @@ fn test_from_robj() {
 }
 
 fn compare_robj_arrays<'a, T>(a: &Robj, b: &Robj)
-    where T: 'a + std::fmt::Debug + PartialEq,
-          Robj: AsTypedSlice<'a, T>
+where
+    T: 'a + std::fmt::Debug + PartialEq,
+    Robj: AsTypedSlice<'a, T>,
 {
     assert!(a.is_vector() || a.is_array());
     assert!(b.is_vector() || b.is_array());
     assert_eq!(a.len(), b.len());
     // assert_eq!(a.get_attrib(class_symbol()), b.get_attrib(class_symbol()));
     assert_eq!(a.get_attrib(dim_symbol()), b.get_attrib(dim_symbol()));
-    assert_eq!(AsTypedSlice::<T>::as_typed_slice(a), AsTypedSlice::<T>::as_typed_slice(b));
+    assert_eq!(
+        AsTypedSlice::<T>::as_typed_slice(a),
+        AsTypedSlice::<T>::as_typed_slice(b)
+    );
 }
 
 #[test]
--- a/extendr-api/src/robj_ndarray.rs
+++ b/extendr-api/src/robj_ndarray.rs
@@ -1,6 +1,6 @@
 #[doc(hidden)]
 use ndarray::prelude::*;
-use ndarray::Data;
+use ndarray::{Data, IntoDimension, ShapeBuilder};
 
 use crate::prelude::{class_symbol, dim_symbol};
 use crate::*;
@@ -60,26 +60,28 @@ make_array_view_2!(f64, "Not a floating point matrix.");
 impl<A, S, D> From<ArrayBase<S, D>> for Robj
 where
     S: Data<Elem = A>,
-    A: Copy + ToVectorValue,
+    A: Copy + ToVectorValue + std::fmt::Debug,
     D: Dimension,
 {
     fn from(arr: ArrayBase<S, D>) -> Self {
-        let dims: Vec<i32> = arr.shape().iter().map(|x| *x as i32).collect();
-        let mut ret = arr
-            .as_slice()
-            .ok_or("Could not convert")
-            .unwrap()
-            .iter()
-            .copied()
-            .collect_robj();
+        let dims: Vec<i32>;
+        let inverse = arr.t();
+        let reshaped: ArrayView<A, D>;
+        if arr.is_standard_layout() {
+            // If we have a standard layout array, we have to keep the current dimensions
+            // but transpose the memory layout
+            dims = arr.shape().iter().map(|x| *x as i32).collect();
+            reshaped = inverse;
+        } else {
+            // If we have a fortran layout array, we can keep the current memory layout, but
+            // transpose the dimensions
+            reshaped = arr.view();
+            dims = inverse.shape().iter().map(|x| *x as i32).collect();
+        }
+        let mut ret = reshaped.iter().copied().collect_robj();
 
         if dims.len() > 1 {
             ret = ret.set_attrib(dim_symbol(), dims.clone()).unwrap();
-            if dims.len() == 2 {
-                ret = ret.set_attrib(class_symbol(), ["matrix", "array"]).unwrap()
-            } else {
-                ret = ret.set_attrib(class_symbol(), ["array"]).unwrap()
-            }
         }
 
         ret
@@ -158,36 +160,26 @@ fn test_from_robj() {
     }
 }
 
-fn compare_robj_arrays<'a, T>(a: &Robj, b: &Robj)
-where
-    T: 'a + std::fmt::Debug + PartialEq,
-    Robj: AsTypedSlice<'a, T>,
-{
-    assert_eq!(a.rtype(), b.rtype());
-    assert_eq!(a.len(), b.len());
-    // assert_eq!(a.get_attrib(class_symbol()), b.get_attrib(class_symbol()));
-    assert_eq!(a.get_attrib(dim_symbol()), b.get_attrib(dim_symbol()));
-    assert_eq!(
-        AsTypedSlice::<T>::as_typed_slice(a),
-        AsTypedSlice::<T>::as_typed_slice(b)
-    );
-}
-
 #[test]
 fn test_to_robj() {
     test! {
-    compare_robj_arrays::<i32>(
+    assert_eq!(
         &Robj::from(array![1., 2., 3.]),
         &R!("c(1, 2, 3)").unwrap()
     );
 
-    compare_robj_arrays::<i32>(
-        &Robj::from(array![[1., 2., 3.], [4., 5., 6.]]),
+    assert_eq!(
+        &Robj::from(Array::from_shape_vec((2, 3), vec![1., 2., 3., 4., 5., 6.]).unwrap()),
         &R!("matrix(c(1, 2, 3, 4, 5, 6), nrow=2, byrow=T)").unwrap()
     );
 
-    compare_robj_arrays::<i32>(
-        &Robj::from(array![[[1, 2], [3, 4]], [[5, 6], [7, 8]]]),
+    assert_eq!(
+        &Robj::from(Array::from_shape_vec((3, 2).f(), vec![1., 2., 3., 4., 5., 6.]).unwrap()),
+        &R!("matrix(c(1, 2, 3, 4, 5, 6), nrow=2, byrow=T)").unwrap()
+    );
+
+    assert_eq!(
+        &Robj::from(array![[[1, 5], [3, 7]], [[2, 6], [4, 8]]]),
         &R!("array(1:8, dim=c(2, 2, 2))").unwrap()
     );
     }
--- a/extendr-api/src/robj_ndarray.rs
+++ b/extendr-api/src/robj_ndarray.rs
@@ -182,10 +182,10 @@ fn test_to_robj() {
         &R!("c(1, 2, 3)").unwrap()
     );
 
-    // compare_robj_arrays::<i32>(
-    //     &Robj::from(array![[1, 2, 3], [4, 5, 6]]),
-    //     &R!("matrix(c(1, 2, 3, 4, 5, 6), nrow=2, byrow=T)").unwrap()
-    // );
+    compare_robj_arrays::<i32>(
+        &Robj::from(array![[1, 2, 3], [4, 5, 6]]),
+        &R!("matrix(c(1, 2, 3, 4, 5, 6), nrow=2, byrow=T)").unwrap()
+    );
 
     compare_robj_arrays::<i32>(
         &Robj::from( array![[[1, 2], [3, 4]], [[5, 6], [7, 8]]]),
--- a/extendr-api/src/robj_ndarray.rs
+++ b/extendr-api/src/robj_ndarray.rs
@@ -163,9 +163,7 @@ where
     T: 'a + std::fmt::Debug + PartialEq,
     Robj: AsTypedSlice<'a, T>,
 {
-    assert!(a.is_vector() || a.is_array());
-    assert!(b.is_vector() || b.is_array());
-    assert_eq!(a.sexptype(), b.sexptype());
+    assert_eq!(a.rtype(), b.rtype());
     assert_eq!(a.len(), b.len());
     // assert_eq!(a.get_attrib(class_symbol()), b.get_attrib(class_symbol()));
     assert_eq!(a.get_attrib(dim_symbol()), b.get_attrib(dim_symbol()));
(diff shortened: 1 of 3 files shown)


clauswilke:
You'll probably have to test for equality of each individual element, just like the other tests in the same source file.

multimeric:
Oh is **that** why all the other tests did that. Is there any discussion about implementing a `PartialEq` for Robj against Robj that compares by value?

yutannihilation:
> Is there any discussion about implementing a `PartialEq` for Robj against Robj that compares by value?

I can't find any. My guess is that `Robj` is basically a pointer to a data owned by R, so the comparison is done in pointer-level. But I agree it's handy if we can compare by value somehow.

multimeric:
> I can't find any. My guess is that Robj is basically a pointer to a data owned by R, so the comparison is done in pointer-level. But I agree it's handy if we can compare by value somehow.

Well I'll open an issue for it. Surely the R API has a way to compare by value? Or if not we can just write comparison functions like I have done here for vectors/arrays

yutannihilation:
Thanks, I think we need to write it by ourselves.

multimeric:
Turns out I'm wrong, actually. We're using the `R_compute_identical` API function in our implementation: https://github.com/extendr/extendr/blob/a06962c7e42e97bdc078a2ed4e4a47b54749f5d1/extendr-api/src/robj/mod.rs#L956-L967

And this function quite neatly compares all aspects of any Robj by value: 
https://github.com/SurajGupta/r-source/blob/a28e609e72ed7c47f6ddfbb86c85279a0750f0b7/src/main/identical.c#L86

I think my code was not working with this function only because the attribute lists were different (which I still don't understand).

extendr-api/src/robj_ndarray.rs
```diff
@@ -55,6 +57,34 @@ make_array_view_2!(f64, "Not a floating point matrix.");
...
+            .collect_robj();
+
+        if dims.len() > 1 {
+            ret = ret.set_attrib(dim_symbol(), dims.clone()).unwrap();
+            if dims.len() == 2 {
+                ret = ret.set_attrib(class_symbol(), ["matrix", "array"]).unwrap()
+            } else {
+                ret = ret.set_attrib(class_symbol(), ["array"]).unwrap()
+            }
+        }
```
yutannihilation, line 83: `matrix` and `array` are not S3 classes. I think you can just set `.Dim` attribute unconditionally (I mean, even when `dim.len() == 1`).

``` r
f <- function(dim) {
  x <- array(1:6, dim = dim)
  
  print(x)
  
  # check class
  print(is(x))
  
  # check the serialized form (confirm there's no class)
  dput(x)
  
  # check the internal representation
  .Internal(inspect(x))
}

# vec
f(6L)
#> [1] 1 2 3 4 5 6
#> [1] "array"     "structure" "vector"   
#> structure(1:6, .Dim = 6L)
#> @55a9750033f8 13 INTSXP g0c3 [REF(1),ATT] (len=6, tl=0) 1,2,3,4,5,...
#> ATTRIB:
#>   @55a975ddc598 02 LISTSXP g0c0 [REF(1)] 
#>     TAG: @55a974082ac0 01 SYMSXP g0c0 [MARK,REF(2694),LCK,gp=0x4000] "dim" (has value)
#>     @55a975f1e9b0 13 INTSXP g0c1 [REF(65535)] (len=1, tl=0) 6

# matrix
f(c(2L, 3L))
#>      [,1] [,2] [,3]
#> [1,]    1    3    5
#> [2,]    2    4    6
#> [1] "matrix"    "array"     "structure" "vector"   
#> structure(1:6, .Dim = 2:3)
#> @55a977c09e98 13 INTSXP g0c3 [REF(1),ATT] (len=6, tl=0) 1,2,3,4,5,...
#> ATTRIB:
#>   @55a9766f52a8 02 LISTSXP g0c0 [REF(1)] 
#>     TAG: @55a974082ac0 01 SYMSXP g0c0 [MARK,REF(2732),LCK,gp=0x4000] "dim" (has value)
#>     @55a9775b4ef0 13 INTSXP g0c1 [REF(65535)] (len=2, tl=0) 2,3

# array
f(c(2L, 3L, 1L))
#> , , 1
#> 
#>      [,1] [,2] [,3]
#> [1,]    1    3    5
#> [2,]    2    4    6
#> 
#> [1] "array"     "structure" "vector"   
#> structure(1:6, .Dim = c(2L, 3L, 1L))
#> @55a977c24ce8 13 INTSXP g0c3 [REF(1),ATT] (len=6, tl=0) 1,2,3,4,5,...
#> ATTRIB:
#>   @55a97678e768 02 LISTSXP g0c0 [REF(1)] 
#>     TAG: @55a974082ac0 01 SYMSXP g0c0 [MARK,REF(2735),LCK,gp=0x4000] "dim" (has value)
#>     @55a977c0f268 13 INTSXP g0c2 [REF(65535)] (len=3, tl=0) 2,3,1
```

<sup>Created on 2021-09-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

yutannihilation:
Oh, I didn't notice it, sorry.

yutannihilation (changes requested):

yutannihilation:
@andy-thomason @clauswilke @Ilia-Kosenkov 
Though I'm not familiar with the implementation around ndarray, this pull request basically looks good to me. We might need to tweak some implementation in association with the ongoing changes on conversions, but I expect this won't be a burden as the implementation is simple. Any major concerns?

extendr-api/src/robj_ndarray.rs
multimeric: I'm not entirely sure what this means. Are you saying that they don't have a class symbol set ever?

yutannihilation: Yes.

clauswilke: Yes, correct. Compare these two. First does not have a class symbol, second does.

``` r
x <- array(1:6, dim = 2)
dput(x)
#> structure(1:2, .Dim = 2L)

x <- grid::unit(4, "in")
dput(x)
#> structure(4, unit = 2L, class = c("simpleUnit", "unit", "unit_v2"
#> ))
```

<sup>Created on 2021-08-31 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

clauswilke:
I agree.

extendr-api/src/robj_ndarray.rs
multimeric: > matrix and array are not S3 classes. I think you can just set .Dim attribute unconditionally (I mean, even when dim.len() == 1).

This doesn't appear to be true. If I set dim unconditionally the comparison (implemented in R's API) fails.

multimeric:
I just pushed a refactor, it should be much cleaner now.

Other than what is being discussed in the review, I'd appreciate review on:

* The way that I'm cloning the data before conversion to R. Is that correct?
* The way that I'm handling memory layouts. It occurred to me that since `ndarray` supports both C and F layouts, while R supports only F (I think?), there needed to be some kind of handling. I've since implemented this with a test, but it's doing my head in and I would appreciate someone double checking that I've done this correctly

extendr-api/src/robj_ndarray.rs
yutannihilation: I guess that's because you are using `c()` for testing 1D array. I'm suggesting returning a 1D array instead of a vector.

multimeric: Right, but I think that R users will expect 1D outputs from any function to be as a vector rather than a 1D array. 1D arrays are basically unheard of to my knowledge.

multimeric: Or would you rather that this conversion always returns an array, for consistency?

yutannihilation: Yes, for the sake of consistency. But, I'm not fully sure what's the right thing to do here. I'd like to wait for others' comments.

> 1D arrays are basically unheard of to my knowledge.

I think that's because 1D arrays can be handled just the same as usual vectors transparently. 1D array is probably not useful, but should be harmless.

multimeric: You could be right, they're probably harmless.

multimeric: Yeah look I'm tending towards agreeing now. It means a simpler implementation. And if `extendr` users want to return an R vector they can just return [`arr.to_vec()`](https://docs.rs/ndarray/0.15.3/ndarray/struct.ArrayBase.html#method.to_vec) to do this explicitly (I assume we already have a conversion between R and Rust vectors).

extendr-api/src/robj_ndarray.rs
```diff
@@ -127,3 +158,28 @@ fn test_from_robj() {
...
+
+    assert_eq!(
+        &Robj::from(Array::from_shape_vec((3, 2).f(), vec![1., 2., 3., 4., 5., 6.]).unwrap()),
+        &R!("matrix(c(1, 2, 3, 4, 5, 6), nrow=2, byrow=T)").unwrap()
+    );
+
+    assert_eq!(
+        &Robj::from(array![[[1, 5], [3, 7]], [[2, 6], [4, 8]]]),
+        &R!("array(1:8, dim=c(2, 2, 2))").unwrap()
+    );
```
Ilia-Kosenkov, line 183: It would be nice to have a test for something like `array(1:6, c(1, 2, 3))`, i.e. when the size of one of the dimensions is 1.

multimeric: Fair enough, will do.

multimeric: Fair enough, will do.

Ilia-Kosenkov:
@multimeric,
I enabled CI for you so you can test your changes en masse.

Ilia-Kosenkov:
@multimeric 
> The way that I'm cloning the data before conversion to R. Is that correct?

I can answer this. We have recently devised some [ground rules](https://github.com/extendr/extendr/blob/master/design_specifications/marshaling_rules.md) on how to handle the marshaling of `R` objects into `Rust`, starting with `R` vectors (not arrays of dim 1 though). As a result, the way we handle arrays will likely change in the future, but for now, I believe, any reasonably working solution will do fine. 

>Do we approve of using as i32 to convert between usize and i32? I gather that it truncates so that if we have a dimension that's too long to represent as an i32 then it will just fail silently and instead return the maximum i32 value.

I am not really familiar with `ndarray`, but I assume you are talking about converting array dimensions to `i32` used to set `dim`.
I do not like such unchecked casts, if a `Rust` array is 'too large', I doubt `R` can handle it correctly anyway, so overflowing casts should just fail.
I suggest to at least leave a comment in the code mentioning this issue.

multimeric:
> I am not really familiar with ndarray, but I assume you are talking about converting array dimensions to i32 used to set dim.
I do not like such unchecked casts, if a Rust array is 'too large', I doubt R can handle it correctly anyway, so overflowing casts should just fail.

Okay, I've changed this such that it will panic in this (rare) scenario. I could also change it to throw an error, but then we would have to implement `TryFrom` instead of `From`, but I don't think any other conversions do this?

clauswilke:
> Okay, I've changed this such that it will panic in this (rare) scenario. I could also change it to return an error, but then we would have to implement TryFrom instead of From, but I don't think any other conversions do this?

We're moving to `TryFrom` throughout to exactly avoid these kinds of issues. We're just not there yet, you caught us in the middle of this transition. I don't like random panics or unexpected data corruption, even if extremely rare.

multimeric:
If there's another branch I should target in order to use `TryFrom`, I happily will.

clauswilke:
Therefore, I would suggest moving to `TryFrom` now rather than leaving the code half-finished and hoping somebody will fix it later. It'll cause 100x more pain later on, when people are actually using the API and experience breaking changes.

clauswilke:
No, development is happening in master.

multimeric:
Okay. Do we have some kind of error scheme I should be following? Just strings?

Ilia-Kosenkov:
> 
> 
> Okay. Do we have some kind of error scheme I should be following? Just strings?

I think this is the one
https://github.com/extendr/extendr/blob/master/extendr-api/src/error.rs

multimeric:
Okay. Do I just add new errors to that enum? Because the kind of errors I expect are like "axis too long", and whatever kind of error `set_attrib` returns.

clauswilke:
Maybe just use `Other` with a string? Since this is going to be an extremely rare and obscure error, I'm not sure it needs its own type.

Ilia-Kosenkov:
`Other` with a string is probably preferable because this may be reworked in the nearest future (not entirely, of course), and other Error types may emerge. I am not entirely sure about the usage of the word `axis` though, perhaps array dimension? 
Coming from `R`, I'd expect to receive `"axis too long"` when making plots, not when dealing with arrays.

multimeric:
Done.

Ilia-Kosenkov:
LGTM, but I'd like to suggest to test one other edge case -- 'empty' arrays. As far as I can see, we have no such tests.
I propose something like `matrix(integer(), nrow = 10)`, which actually produces an empty matrix. 
I believe it is a valid use case as `R` matrix can be sliced into such 'empty' object.

PS: I was under the impression the CI will continue working after I approved it once, but it seems it requires manual triggering for now.

clauswilke:
I believe CI needs to be approved manually until somebody has made at least one accepted contribution to a repo: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

It doesn't say so in the github docs, but I believe this is primarily to prevent cryptocurrency mining via github actions.

extendr-api/src/robj_ndarray.rs
```diff
@@ -127,3 +159,61 @@ fn test_from_robj() {
...
+        assert_eq!(
+            &Robj::try_from(&array![1., 2., 3.])?,
+            &R!("array(c(1, 2, 3))")?
+        );
+
+        // We give both R and Rust the same 1d vector and tell them both to read it as a matrix
+        // in C order. Therefore these arrays should be the same.
+        assert_eq!(
+            &Robj::try_from(&Array::from_shape_vec((2, 3), vec![1., 2., 3., 4., 5., 6.]).unwrap())?,
+            &R!("matrix(c(1, 2, 3, 4, 5, 6), nrow=2, byrow=T)")?
```
clauswilke, line 175: One minor nitpick: Please write out `TRUE` and `FALSE`. `T` and `F` are bad style and it's best to get used to never use them. (They are just variables that have been assigned `TRUE` and `FALSE`, respectively.)

extendr-api/src/robj_ndarray.rs
```diff
@@ -127,3 +159,61 @@ fn test_from_robj() {
...
+        );
+
+        // We give both R and Rust the same 1d vector and tell them both to read it as a matrix
+        // in fortran order. Therefore these arrays should be the same.
+        assert_eq!(
+            &Robj::try_from(&Array::from_shape_vec((2, 3).f(), vec![1., 2., 3., 4., 5., 6.]).unwrap())?,
+            &R!("matrix(c(1, 2, 3, 4, 5, 6), nrow=2, byrow=F)")?
```
clauswilke, line 182: See above. `byrow = FALSE`.

multimeric:
> LGTM, but I'd like to suggest to test one other edge case -- 'empty' arrays. 

Done.

multimeric:
Only the MacOS test is failing but I don't think it's due to my PR?

clauswilke:
I'm re-running the tests. Not clear to me why OSX is failing.

clauswilke:
The OSX failure has persisted for a while now. Maybe @yutannihilation knows what's going on?

clauswilke (approved):

Ilia-Kosenkov:
> Only the MacOS test is failing but I don't think it's due to my PR?

CRAN sometimes incorrectly distributes `R` binaries/sources,  especially `devel` ones. In our case, it fails at the `R` installation step with an HTTP 404 error while trying to download the newest version, which is pretty much self-explanatory :)
We can ignore this failure for now as there seem to be no other issues on other platforms, including macOS.

Ilia-Kosenkov (approved):
Looks good.
Few suggestions:
- First, I'd recommend to update the `CHANGELOG.md` with a one-line entry mentioning the support of `TryFrom` for arrays. You not only implemented new features, but also bumped the version of an (optional) dependency, so I'd say it is worth writing down.
- Second, if you feel like doing some `R` coding, it could be nice to have some array tests in `tests/extendrtests`. This package showcases `extendr` features (and also tests them), so having, say, a Rust function that accepts two matrices, multiplies them, and then returns the product back to `R` would be nice. This is a topic for a separate PR and can be implemented by someone else.

yutannihilation:
> The OSX failure has persisted for a while now. Maybe @yutannihilation knows what's going on?

No idea, but let's ignore.

```
Error: Failed to get R devel: Failed to get R devel: Failed to download version devel: Error: Unexpected HTTP response: 404
```

yutannihilation (approved):

multimeric:
> First, I'd recommend to update the CHANGELOG.md 

Done.

I also got some helpful comments from one of the `ndarray` devs here: https://github.com/rust-ndarray/ndarray/issues/1060 which I used to inform a slight optimisation.

> Second, if you feel like doing some R coding, it could be nice to have some array tests in tests/extendrtests. This package showcases extendr features (and also tests them), so having, say, a Rust function that accepts two matrices, multiplies them, and then returns the product back to R would be nice. This is a topic for a separate PR and can be implemented by someone else.

Sounds good, but yes I would rather leave this to another PR.

Ilia-Kosenkov:
@multimeric,
Seems all is good, thank you for your contribution. I am merging now.

multimeric:
Hooray! Thanks for the merge. I look forward to the next major release.

PR #279, opened 2021-09-06
Graphics functions

andy-thomason:
This PR adds low level graphics functions.

Stretch goal: maybe the ability to define new graphics devices.

The task is made more challenging by the varying support given to each
version of R.

We will likely have to put this behind a feature gate.

### DIFF
--- a/extendr-api/src/graphics/mod.rs
+++ b/extendr-api/src/graphics/mod.rs
@@ -6,7 +6,7 @@ pub struct Context {
 }
 
 #[derive(Clone, Debug, PartialEq)]
-pub struct DevDesc {
+pub struct Device {
     inner: pGEDevDesc,
 }
 
@@ -15,7 +15,7 @@ pub struct Pattern {
     inner: Robj,
 }
 
-impl DevDesc {
+impl Device {
     pub(crate) fn inner(&self) -> pGEDevDesc {
         self.inner
     }
@@ -43,11 +43,11 @@ pub enum LineType {
     Twodash,
 }
 
-pub enum Mode {
-    Off,
-    On,
-    InputOn,
-}
+// pub enum Mode {
+//     Off,
+//     On,
+//     InputOn,
+// }
 
 pub enum FontFace {
     PlainFont,
@@ -59,24 +59,28 @@ pub enum FontFace {
 
 impl Context {
     pub fn new() -> Self {
-        // See GInit()
-        let inner = R_GE_gcontext {
-            col: -1,
-            fill: -1,
-            gamma: 1.0,
-            lwd: 5.0,
-            lty: 0,
-            lend: R_GE_lineend_GE_ROUND_CAP,
-            ljoin: R_GE_linejoin_GE_ROUND_JOIN,
-            lmitre: 10.0,
-            cex: 1.0,
-            ps: 1.0,
-            lineheight: 1.0,
-            fontface: 0,
-            fontfamily: [0; 201],
-            // patternFill: R_NilValue,
-        };
-        Self { inner }
+        #[allow(unused_unsafe)]
+        unsafe {
+            let inner = R_GE_gcontext {
+                col: -1,
+                fill: -1,
+                gamma: 1.0,
+                lwd: 5.0,
+                lty: 0,
+                lend: R_GE_lineend_GE_ROUND_CAP,
+                ljoin: R_GE_linejoin_GE_ROUND_JOIN,
+                lmitre: 10.0,
+                cex: 1.0,
+                ps: 1.0,
+                lineheight: 1.0,
+                fontface: 0,
+                fontfamily: [0; 201],
+
+                #[cfg(use_r_patternfill)]
+                patternFill: R_NilValue,
+            };
+            Self { inner }
+        }
     }
 
     pub fn rgb(red: u8, green: u8, blue: u8) -> i32 {
@@ -154,10 +158,10 @@ impl Context {
         self
     }
 
-    pub fn char_extension(&mut self, cex: f64) -> &mut Self {
-        self.inner.cex = cex;
-        self
-    }
+    // pub fn char_extra_size(&mut self, cex: f64) -> &mut Self {
+    //     self.inner.cex = cex;
+    //     self
+    // }
 
     pub fn font_face(&mut self, fontface: FontFace) -> &mut Self {
         use FontFace::*;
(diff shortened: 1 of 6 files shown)


andy-thomason:
The current challenge is to find some values for `R_GE_gcontext` that actually draws a line!

clauswilke:
You'll probably have to set the line type (e.g. 1), line width (e.g. 1), and color (e.g. "blue"). See the documentation and examples for the `gpar()` function, which is mostly a 1-1 mapping to the graphics context.

https://www.rdocumentation.org/packages/grid/versions/3.6.2/topics/gpar

clauswilke:
It may also be helpful to look at the grid source code, e.g. the code creating graphics contexts from gpar objects:
https://github.com/wch/r-source/blob/79298c499218846d14500255efd622b5021c10ec/src/library/grid/src/gpar.c#L306

andy-thomason:
I think that I'm getting there.

Would anyone like to guess what "cex" stands for. It seems to be a fount size multiplier.

clauswilke:
From the documentation of `par()` (in base R):

> `cex` A numerical value giving the amount by which plotting text and symbols should be magnified relative to the default. This starts as 1 when a device is opened, and is reset when the layout is changed, e.g. by setting mfrow.

> Note that some graphics functions such as plot.default have an argument of this name which multiplies this graphical parameter, and some functions such as points and text accept a vector of values which are recycled.

I usually just set it to 1 and keep it there.

andy-thomason:
Roadmap for this (deadline 1st October):

* Implement graphics::Context (R_GE_context) - write-only attributes, from_par.
* Implement graphcis::Device - line, polyline, text etc.
* Provide methods to select standard drivers: pdf, postscirpt.
* Test using the postscript device.

Stretch goal:

* Implement a trait for user-defined graphics devices.
* Write a simple SVG renderer.

andy-thomason:
We also need to resolve version issues with `#[cfg]s. 

The pattern parameter was added recently causing many of the build failures.

andy-thomason:
The postscript driver is working quite well now.

Screenshot from 2021-09-08 17-55-33

But the X11 driver is segfaulting when drawing filled regions (circle).

clauswilke:
Looks promising! Could we get access to font metrics? Function `GEStrMetric()`.

andy-thomason:
Text is next...

andy-thomason:
I suspect that the X11 crash is due to the clipping state not being set. I'll try to diagnose this.

clauswilke:
> Text is next...

Super, thanks!

Btw., it'll be fairly easy (I think) to write a set of visual tests as part of the integration tests, using vdiffr: https://cran.r-project.org/web/packages/vdiffr/index.html
We just have to write code that generates graphical output, and vdiffr will take care of storing snapshots and comparing for changes.

andy-thomason:
The segfaulting issue was because of a version mismatch between the headers and my R library.

I built R from source - fairly easy on unix - and the problem went away.

It may be worth comparing the version in the headers with the version returned by the library.

andy-thomason:
Screenshot from 2021-09-16 19-57-23

andy-thomason:
We should probably call this done for now.

I would be interested to see someone use it for plotting.

We will come back to this later.

clauswilke (approved):
This is great! I think it's an excellent starting point for anybody who wants to write Rust code to interact with graphics devices, and I agree we can leave it like this for now and come back to it when there is meaningful usage of the code.

Ilia-Kosenkov:
I have one comment:
I feel like having all color-related functions at the top level is not the best idea.
Maybe I am spoiled by heavy-oop languages, but  `color::Colors::aliceblue()` looks more appropriate. All color related functions should be under `color` namespace, but predefined colors should reside in their own type or module.

andy-thomason:
Maybe a `predefined` module: `use color::predefined::*`.

Ilia-Kosenkov:
It is up to you, I won't oppose leaving all these functions at the top level. But after doing some desktop UI (using another stack), I came to appreciate when there is some `Color` representing just one color, and some `Colors`, members of which are pre-defined, well, *colors*. Sounds easy.

So if I have `struct Color`, I would definitely first look for `fn Colors::white() -> Color` at the same level.
Examples: .NET's [`Color`](https://docs.microsoft.com/en-us/dotnet/api/system.windows.media.color?view=net-5.0) and [`Colors`](https://docs.microsoft.com/en-us/dotnet/api/system.windows.media.colors?view=net-5.0), `R`'s [`colors()`](https://www.rdocumentation.org/packages/grDevices/versions/3.6.2/topics/colors),

Ilia-Kosenkov (approved):
Now it feels better structured IMO.
`graphics::color::predefined::white()` is self-explanatory and does not pollute `graphics::color::*`.

extendr-api/src/graphics/mod.rs
```diff
@@ -0,0 +1,759 @@
...
+            descent: tm.descent / self.scalar,
+            width: tm.width / self.scalar,
+        }
+    }
+}
+
+#[allow(non_snake_case)]
+impl Device {
+    /// Get the current device.
+    pub fn current() -> Device {
```
yutannihilation, line 343: Can this fail? If so, can we catch it and return `Option`?

extendr-api/src/graphics/mod.rs
```diff
@@ -0,0 +1,759 @@
...
+
+    /// Enable device rendering.
+    pub fn mode_on(&self) {
+        unsafe {
+            GEMode(1, self.inner());
+        }
+    }
+
+    /// Disable device rendering and flush.
+    pub fn mode_off(&self) {
```
yutannihilation, line 359: I think this fails when the current device is the null device. If so, can we catch it and return `Result`?

yutannihilation (approved):
I might be wrong, but I think the operations on graphic device are generally fallible, so we probably need some more consideration to catch them nicely. But I think we can improve later and so this looks good to merge.

extendr-api/src/graphics/mod.rs
```diff
@@ -0,0 +1,759 @@
...
+                ljoin: R_GE_linejoin_GE_ROUND_JOIN,
+                lmitre: 10.0,
+                cex: 1.0,
+                ps: 14.0,
+                lineheight: 1.0,
+                fontface: 1,
+                fontfamily: [0; 201],
+
+                #[cfg(use_r_patternfill)]
+                patternFill: R_NilValue,
```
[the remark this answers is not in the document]
andy-thomason: We are using the R data structure which uses patternFill.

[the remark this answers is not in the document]
yutannihilation: Ah, got it. Sorry.

extendr-api/src/graphics/mod.rs
andy-thomason: These two now have an error return in the case of `Rf_NoDevices()`

Issue #283, opened 2021-09-11
Proposal: Introduce `paste` dependency in `extendr-api`


Ilia-Kosenkov:
### Description
[`paste`](https://crates.io/crates/paste) is a very simple yet useful tool for advanced macros creation.
While I do not claim to be an expert on Rust and especially on macros, it seems the standard `macro_rules!` lacks certain features, while `proc_macro` is too over-engineered for the majority of cases.
`paste` allows simple manipulation with the tokens, including *concatenation* of identities using `paste!` macros and special syntax of `[< ... >]`.
### Justification
In `extendr`, when working with, e.g., `R` vectors, a lot of functions and types are named using the underlying `SEXP` type name. `R` functions have typical prefixes/suffixes of the form `REAL`, `INT`, etc. With the help of `paste`, all of these cases can be covered by providing one argument to `extendr` macros, which is then used to generate function names (like `as_real_slice`, `as_int_slice`). `paste` allows for limited control of capitalization, so it has potential to simplify already existing macros, like these ones (`Add` and `add` can be represented by a single parameter):
https://github.com/extendr/extendr/blob/7e9cac34187d333abd83b79b0e13887249fb7d42/extendr-api/src/scalar/rint.rs#L178-L184

If `paste` allows us to construct function and type names on the fly, we can work with vector types and their counterparts entirely through macros, duplicating **no** code.

As mentioned above, `paste` is much easier than `proc_macro`, and can be easily integrated into existing `macro_rules!` macros.

`paste` has no external dependencies (as far as I can tell, see its [`Cargo.toml`](https://github.com/dtolnay/paste/blob/master/Cargo.toml)), so it should not bloat `extendr` dependency tree.

### Drawbacks
**+1** dependency

yutannihilation:
This sounds good to me.

One more possible drawback of relying macro heavily is build time. I bet 99% this won't make the build massively slow, because the amount of code in extendr is modest. So, we can forget this, but let me leave this comment just in case we found some problem related to the build time.

c.f. "Compilation Model: Macros And Pipelining" section of https://matklad.github.io/2021/09/04/fast-rust-builds.html

Ilia-Kosenkov:
[...]

Issue #282, opened 2021-09-11
`S4` doctests potentially fail on Windows


Ilia-Kosenkov:
While working on the implementation of `Doubles`, I set up a local environment to run all tests, including doc tests.
On Windows, we do cross-compilation, which does not support doc tests. The Nightly version, however, has experimental support for `-Zdoctest-xcompile`, which runs all tests.

I observe the following issue on Windows, after running:
```pwsh
$(rustup default; cargo --version; git log  -n 1 --oneline ; cargo test --manifest-path .\extendr-api\Cargo.toml --target x86_64-pc-windows-gnu --features tests-all -Zdoctest-xcompile) 2>&1 > win.log
```

```
failures:
    src\wrapper\s4.rs - wrapper::s4::S4::get_slot (line 47)
    src\wrapper\s4.rs - wrapper::s4::S4::has_slot (line 104)
    src\wrapper\s4.rs - wrapper::s4::S4::new (line 30)
    src\wrapper\s4.rs - wrapper::s4::S4::set_class (line 14)
    src\wrapper\s4.rs - wrapper::s4::S4::set_slot (line 74)

test result: FAILED. 149 passed; 5 failed; 3 ignored; 0 measured; 0 filtered out; finished in 65.15s

```

Full test output is available here:

<details>
<summary>win.log</summary>

```pwsh
nightly-x86_64-pc-windows-msvc (default)
cargo 1.56.0-nightly (e96bdb0c3 2021-08-17)
7e9cac34 Altrep Integers class. (#274)
    Finished test [unoptimized + debuginfo] target(s) in 0.11s
     Running unittests (C:\Users\[redacted]\AppData\Local\Temp\cargo_temp\x86_64-pc-windows-gnu\debug\deps\extendr_api-76396fb6968d5bf3.exe)

running 22 tests
test robj::tests::output_iterator_test ... ok
test robj::tests::test_try_from_robj ... ok
test robj_ndarray::test_from_robj ... ok
test robj_ndarray::test_round_trip ... ok
test robj_ndarray::test_to_robj ... ok
test tests::big_r_macro_works ... ok
test tests::class_wrapper_test ... ok
test tests::export_test ... ok
test tests::metadata_test ... ok
test tests::pairlist_macro_works ... ok
test tests::r_output_test ... ok
test tests::slice_test ... ok
test tests::test_na_str ... ok
test wrapper::matrix::matrix_ops ... ok
test wrapper::symbol::test::test_constant_symbols ... ok
[...]

PR #284, opened 2021-09-11
[WIP] Implements `Rfloat` and `Doubles` using macros

### LINKED ISSUE #280: Implement "Doubles" wrapper
Given the work on "Integers" in #274 I would like one of our contributors to make a PR
for the "Doubles" equivalent.

@Ilia-Kosenkov might be a good fit!

What do you think?

Ilia-Kosenkov:
Fixes #280.

This turned out to be a much larger contribution than I expected.
After implementing `Doubles`, I decided to introduce `paste` as a dependency (see #283) and then further improved code generation. 

Major topics:
1. `Rfloat` as a thin `NA`-aware wrapper around `f64`. I generalized macros so that both `Rint` and `Rfloat` are generated using the same macros, and the only thin written by hand is `TryFrom` that also allows coercion (i.e., floats that can be represented as ints can be converted to `Rint`).
2. `Doubles`. `Doubles` is an almost exact copy of `Integers`, so, with the help of `paste`, both `Doubles` and `Integers` are generated using a single macros. I expect this approach will work as well for other primitive types. If not, at least in these cases we do not write the same code twice.
3. Moved tests from doc strings to `test` module. When members are generated using macros, it is even harder to generate meaningful tests. In this case, I took out test cases, and put them in a separate module (under `crate::wrappers::{integers, doubles}::tests`). These tests are guarded with `#[cfg(test)]` and are now executed in all build scenarios (see #282).

Takeways:
- Prefer 'horizontal' development to 'vertical' (i.e., deal with one abstraction layer fully instead of implementing support of one type), which encourages macros usage instead of heavy copy-pasting
- Prefer meaningful tests in a separate module to reduce comment pollution around `impl` members
- Carefully address edge cases (marked with `TODO`)

TODO:
- [x] Meaningful tests in `extendr-api/tests`
- [x] Meaningful tests in `tests/extendrtests`
- [x] Find a way to keep docs when generating members using macros


clauswilke:
This is great!

One comment about doc strings: I understand the reasoning that they shouldn't be used to test the code, and I agree. However, I think 1-2 examples per function are super helpful and make the code more accessible for a new user. So I wouldn't just delete all these examples. Rather assess whether the example is actually helpful or not.

Ilia-Kosenkov:
@clauswilke, I agree about examples. The challenge is -- when members are generated using macros, we have two options -- generate examples the same way we generate all these methods right now, which can be tricky, or add additional parameters to the macros, which are translated into docs. 
This is challenging, and I will look into it (and encourage others with the knowledge of macros to do the same). I certainly do prefer macros with weaker documentation but no code repetition compared to a lot of code repetition but slightly better docs.

clauswilke:
> The challenge is -- when members are generated using macros, we have two options -- generate examples the same way we generate all these methods right now, which can be tricky, or add additional parameters to the macros, which are translated into docs.

I see. Maybe it's better to write documentation by hand and to just put all the relevant info and examples into a single documentation block for the entire type. In other words, put the documentation and examples for `Doubles` here: 
https://github.com/extendr/extendr/blob/65492817e92eca226b5755a877ae370eb94fa6d4/extendr-api/src/wrapper/doubles.rs#L4

Ilia-Kosenkov:
@clauswilke,
Yeah, it is probably better to write the whole documentation block in one place.
However, we can provide some limited support for doctest (and examples) by generating docstrings within the macros. 
The last commit 27c0858 demonstrates such an example. The generated test case also participates in the testing.

clauswilke:
I see. That's great also. Whichever way you prefer.

Ilia-Kosenkov:
So I propose to auto-generate some simple examples using macros (like for `new` or `is_na`, which are identical for all types) and write more detailed examples in the type documentation -- this can be different for each type.

andy-thomason:
Looking good I am usually cautious about using extra dependencies, but the problems with `concat_ident!` require this for now.
There is a PR to fix this in the language but don't forget to exhale.

It is worth thinking about Logicals/Rbool (aka Bool) but not in this PR.

This shares much with Integers.

We may need to implement more operators in the future (such as &, |, ^) for integer and bool.

Ilia-Kosenkov:
I have identified several issues with `Rfloat` that I need to address.
The main problem is that `NA` checks and `PartialEq` do not work in the current (macros) implementation, because `NA` is represented with `f64::NaN`, and `f64::NaN != f64::NaN` (there are multiple `NaN`s, and `NA` uses one value AFAIK).

clauswilke:
> Downside - can no longer compare NAs directly and in assertions,

I don't see this as a downside. As you say, it's exactly how R handles the same case.

andy-thomason:
I agree that Rint::na() should not (partial) compare with another Rint::na()

This behavious makes it easy to compare floats.

tests/extendrtests/src/rust/src/lib.rs
```diff
@@ -22,40 +20,53 @@ fn do_nothing() {
...
+    x
+}
 
 // Convert an int scalar to itself
 // x a number
 #[extendr]
-fn int_scalar(x: i32) -> i32 { x }
+fn int_scalar(x: i32) -> i32 {
+    x
+}
```
Ilia-Kosenkov, line 32: These formatting changes were unintentional, I hooked `cargo fmt` to OnSave of my CLion, so once I edited this file, it automatically applied formatting and I did not notice it.

extendr-api/src/wrapper/macros.rs
```diff
@@ -0,0 +1,154 @@
...
+                #[doc = "   assert!(vec.elt(10).is_na());"]
+                #[doc = "}"]
+                #[doc = "```"]
+                pub fn elt(&self, index: usize) -> $type_elem {
+                    // Defensive check for oob
+                    if(index >= self.len()) {
+                        <$type_elem>::na()
+                    } else {
+                        unsafe { [<$r_type _ELT>](self.get(), index as R_xlen_t).into() }
+                    }
```
yutannihilation, line 94: `*_ELT` will check the bounds as of the next major release of R; I mean, the internal, macro version of `*_ELT` don't, but the function version of them do, and the internal version won't be available anymore. 

c.f. https://twitter.com/dvaughan32/status/1417647704289824770

<details>

<summary>e.g. function version of `INTEGER_ELT`</summary>

https://github.com/wch/r-source/blob/77d6c4371eaa690f151f36d5767be892e8a7b334/src/include/Rinlinedfuns.h#L393-L397

CHECK_VECTOR_INT_ELT
https://github.com/wch/r-source/blob/77d6c4371eaa690f151f36d5767be892e8a7b334/src/include/Rinlinedfuns.h#L277-L282

CHECK_BOUNDS_ELT
https://github.com/wch/r-source/blob/77d6c4371eaa690f151f36d5767be892e8a7b334/src/include/Rinlinedfuns.h#L266-L269

</details>

So, if we want to avoid the double check, we might want to access to the underlying pointer directly in the case of non-ALTREP, just like what cpp11 does.

https://github.com/r-lib/cpp11/blob/1b4beb25f56c8dca6e0448d5dc55fb64fddab5d1/inst/include/cpp11/integers.hpp#L29-L32

yutannihilation (commented):
Minor comments.

extendr-api/src/wrapper/macros.rs
Ilia-Kosenkov: I agree, but I did not change the functions that were already used in the case of `Integers`.
I agree that your solution is better, so perhaps you can make a follow-up PR to fix this?
This PR already covers a lot, I do not want to make more changes here.

yutannihilation: > I did not change the functions that were already used in the case of `Integers`.

I commented this because you added the bound check, but, okay, I don't intend to block the PR for this. This comment was just in case you didn't notice my reply on Discord. Let's discuss later :)

yutannihilation (approved):

clauswilke (approved):
I feel this exceed my level of technical understanding so please take my approval as nothing more than saying I'm very much in favor of this work and am glad you're taking the lead.

One minor comment: I always find it hard to read macro code and I would encourage you to be extra careful with documenting. It may seem all obvious to you today as you're writing this but a few more comments probably wouldn't hurt. At a minimum, I think it would be good for every chunk of macro code (i.e., every `impl` section) to provide in a comment a concrete example of what it implements. For example: "Implement binary mathematical operators such as + or -: a + b"

andy-thomason (approved):

PR #287, opened 2021-09-16
Reduce robj to a single pointer

andy-thomason:
This PR make Robj (and most wrappers like Rstr) the same layout as a SEXP.

This will enable read-only slices of Robs for lists and slices of Rstr for strings.

There is a small extra overhead as system objects have to be reference counted
by `ownership::protect()` but the ability to reference list may counteract this.

We may want to check system objects in `ownership`, especially `RNilValue`.

### DIFF
--- a/extendr-api/src/functions.rs
+++ b/extendr-api/src/functions.rs
@@ -83,7 +83,7 @@ pub fn find_namespace<K: Into<Robj>>(key: K) -> Result<Environment> {
 /// }
 /// ```
 pub fn current_env() -> Environment {
-    unsafe { new_owned(R_GetCurrentEnv()).try_into().unwrap() }
+    unsafe { Robj::from_sexp(R_GetCurrentEnv()).try_into().unwrap() }
 }
 
 /// The "global" environment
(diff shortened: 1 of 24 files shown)


yutannihilation:
I'll take a look in a few days, though I'm not confident I can review well...

A quick comment. I think `extendr-api/src/graphics/mod.rs` will conflict with #279. In what order do you think we should merge?

clauswilke:
Same here. I think it would be great if @Ilia-Kosenkov could review.

andy-thomason:
This is in two parts (with two commits).

The first changes the enum to a struct.

The second renames `new_owned` and `new_sys` to `Robj::from_sexp`.

The next PR will include `to_typed_slice` for `List` and `Strings`.

Ilia-Kosenkov:
Will review in the next couple of hours.

Ilia-Kosenkov (approved):
Looks good to me.
To double-check that it does not break anything, I tested `extendr-api` using this PR and nightly pipeline with `-Zdoctest-xcompile` to enable doc tests while cross-compiling.

A surprising side-effect is that this PR seems to fix problems in `S4` tests that I reported in #282.
Or at least it is what I observe while switching between this PR & master.

andy-thomason:
I have left `new_owned` in for the while as it is used by #279 and the function wrappers.

Without the `s/new_owned/Robj::from_sexp/` change, there are relatively few differences.

By protecting more objects, we may have helped with the S4 tests. It is difficult to say. I have also
folded `Rf_AllocateVector` into a `single_threaded` block by using the `Robj::allocVector` call.

yutannihilation (approved):

yutannihilation:
@andy-thomason 
Sorry for reviewing late, I think now I get the point of this pull request. Looks good.

One thing, what I wanted to ask you by this comment was if you accidentally commit `extendr-api/src/graphics/mod.rs` here, which should exist after #279 gets merged. Is it intentionally included here? If so, it's fine.

> A quick comment. I think `extendr-api/src/graphics/mod.rs` will conflict with #279. In what order do you think we should merge?

andy-thomason:
It should not now contain changes to `graphics/mod.rs` but required ninja-like rebase skills.

yutannihilation:
Yeah, that was my concern :) Glad that you are a ninja of it!

PR #288, opened 2021-09-21
Updating `IsNA` trait

Ilia-Kosenkov:
Trait `IsNA` is modified and renamed to`CanBeNA`.
New static function (i.e. function with no `self` arg) is introduced::
`fn CanBeNA::na() -> Self`
which can be used to obtain `NA` value of a given type.

Modified code in key places to switch to `<type>::na()` instead of helper functions.

I did not remove implementations of `CanBeNA` for primitive types (like `f64` or `i32`).
This may be deprecated in the future.

### DIFF
--- a/extendr-api/src/na.rs
+++ b/extendr-api/src/na.rs
@@ -1,5 +1,5 @@
 use libR_sys::{R_IsNA, R_NaReal};
-use crate::{Bool, na_str};
+use crate::Bool;
 /// Return true if this primitive is NA.
 pub trait CanBeNA {
     fn is_na(&self) -> bool;
@@ -45,4 +45,18 @@ impl CanBeNA for &str {
     fn na() -> Self {
         na_str()
     }
+}
+
+/// Special "NA" string that represents null strings.
+/// ```
+/// use extendr_api::prelude::*;
+/// test! {
+///     assert!(na_str().as_ptr() != "NA".as_ptr());
+///     assert_eq!(na_str(), "NA");
+///     assert_eq!("NA".is_na(), false);
+///     assert_eq!(na_str().is_na(), true);
+/// }
+/// ```
+pub fn na_str() -> &'static str {
+    unsafe { std::str::from_utf8_unchecked(&[b'N', b'A']) }
 }
--- a/extendr-api/src/na.rs
+++ b/extendr-api/src/na.rs
@@ -1,0 +1,31 @@
+use libR_sys::R_IsNA;
+use crate::{Bool, na_str};
+/// Return true if this primitive is NA.
+pub trait IsNA {
+    fn is_na(&self) -> bool;
+}
+
+impl IsNA for f64 {
+    fn is_na(&self) -> bool {
+        unsafe { R_IsNA(*self) != 0 }
+    }
+}
+
+impl IsNA for i32 {
+    fn is_na(&self) -> bool {
+        *self == std::i32::MIN
+    }
+}
+
+impl IsNA for Bool {
+    fn is_na(&self) -> bool {
+        self.0 == std::i32::MIN
+    }
+}
+
+impl IsNA for &str {
+    /// Check for NA in a string by address.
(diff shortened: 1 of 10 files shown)


extendr-api/src/na.rs
```diff
@@ -0,0 +1,76 @@
...
+///     assert_ne!(<&str>::na().as_ptr(), "NA".as_ptr());
+///     assert_eq!(<&str>::na(), "NA");
+///     assert_eq!("NA".is_na(), false);
+///     assert_eq!(<&str>::na().is_na(), true);
+/// }
+/// ```
+impl CanBeNA for &str {
+    /// Check for NA in a string by address.
+    fn is_na(&self) -> bool {
+        self.as_ptr() == <&str>::na().as_ptr()
```
clauswilke, line 70: I know this was copied over from the code that was there earlier, but I don't understand how this works. We're comparing whether the pointer is equal to a special `NA` string we have made. But how would this work for `NA`s generated by R?

clauswilke: Oh, I get it, this is for the Rust type `&str`, not for R character vectors.

extendr-api/src/robj/from_robj.rs
```diff
@@ -126,7 +126,7 @@ impl<'a> FromRobj<'a> for Vec<String> {
         } else if let Some(v) = robj.as_string_vector() {
             let str_vec = v.to_vec();
             // check for NA's in the string vector
-            if let Some(_str) = str_vec.iter().find(|&s| *s == na_str()) {
+            if let Some(_str) = str_vec.iter().find(|&s| *s == <&str>::na()) {
```
clauswilke, line 129: Should this use `is.na()`? We should generally avoid comparison to `NA` values, just like in R.

extendr-api/src/na.rs
Ilia-Kosenkov: The concept here as far as I understand is that we create a special string. Original `na_str()` returned `&'static str`, so I assume that this special string is stored as a binary resource and our method always returns a slice from the same block of memory. As a result, by value this slice is identical to `"NA"`, but their in memory locations are different. So if we compare using ptr, only `"NA"` that was created using our method will have correct ptr and will test positive for `NA`, while any other string won't. 

This is a bit of a hack though, but my goal was to update trait, not to rethink implementation, which will likely change as soon as we get scalar and vector types for strings. Same applies for `Bool` used here.

extendr-api/src/robj/from_robj.rs
Ilia-Kosenkov: I checked it and now I am not sure we can replace this. It is an old implementation, and `&s` here is `&String`.
It seems there are no tests to verify that if I change it to something like `(&s[..]).is_na()`, it will still work correctly.
I suggest keeping it 'as-is' and later deprecate these conversions.

Ilia-Kosenkov: I mean, in this case it seems that we search for literal `"NA"` in the string vector, so `<&str>::is_na()` will not work.

clauswilke: Thanks for checking. Could you maybe add a brief comment describing what is happening here?

andy-thomason:
Nice. Good to have a type-specific NA generator.

You can probably lose the turbofish `<&std>::` in many cases as the type inference may do
this for you.

Ilia-Kosenkov:
@andy-thomason,
Actually, it does not seem to work. `&str::na()` for instance, is treated as `&(str::na())`, and there is no `CanBeNA` implementation for `str`. I, however, removed unnecessary `<>` for non-reference types like `f64`, but only in `NA`-related code.

Ilia-Kosenkov:
CI fails due to `404` while getting `R-devel` for mac. I hate that whenever there is a version bump on CRAN side, it is always some random `404` on one of the platforms.
Will re-run the CI later.

clauswilke (approved):

PR #289, opened 2021-09-30
Implement Serializer to convert Serialize objects to Robj.

andy-thomason:
This PR adds serde support to extendr.

This will convert (almost) any Serde object to a Robj and vice versa.

This should provide a serialize/deserialize mechanism for Altrep
and other objects.

We should also update the external pointer type so that we can hook
the load/save calls to embed Rust in RData objects.

```
       use extendr_api::prelude::*;
        use extendr_api::serializer::to_robj;
        use serde::Serialize;
        test! {
            #[derive(Serialize)]
            struct Test {
                int: i32,
                seq: Vec<&'static str>,
            }

            let test = Test {
                int: 1,
                seq: vec!["a", "b"],
            };

            let expected = list!(int=1, seq=list!("a", "b"));
            assert_eq!(to_robj(&test).unwrap(), Robj::from(expected));
        }
 ```

### DIFF
--- a/extendr-api/tests/serializer_tests.rs
+++ b/extendr-api/tests/serializer_tests.rs
@@ -1,0 +1,58 @@
+#[cfg(feature = "serde")]
+mod test {
+    use extendr_api::prelude::*;
+    use extendr_api::serializer::to_robj;
+    use serde::Serialize;
+
+    ////////////////////////////////////////////////////////////////////////////////
+
+    #[test]
+    #[cfg(feature = "serde")]
+    fn test_serialize_struct() {
+        test! {
+            #[derive(Serialize)]
+            struct Test {
+                int: i32,
+                seq: Vec<&'static str>,
+            }
+
+            let test = Test {
+                int: 1,
+                seq: vec!["a", "b"],
+            };
+
+            let expected = list!(int=1, seq=list!("a", "b"));
+            assert_eq!(to_robj(&test).unwrap(), Robj::from(expected));
+        }
+    }
+
+    #[test]
+    #[cfg(feature = "serde")]
+    fn test_serialize_enum() {
+        test! {
+            #[derive(Serialize)]
+            enum E {
+                Unit,
+                Newtype(i32),
+                Tuple(i32, i32),
+                Struct { a: i32 },
+            }
+
+            let u = E::Unit;
+            let expected = list!(Unit=NULL);
+            assert_eq!(to_robj(&u).unwrap(), expected);
+
+            let n = E::Newtype(1);
+            let expected = list!(Newtype=1);
+            assert_eq!(to_robj(&n).unwrap(), expected);
+
+            let t = E::Tuple(1, 2);
+            let expected = list!(Tuple=list!(1, 2));
+            assert_eq!(to_robj(&t).unwrap(), expected);
+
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by yutannihilation** [comment]
Wow, this is cool! I was going to implement some simple API server that receive JSONs and return plot images in order to to test graphic device features. Will review in a few days.

File: extendr-api/tests/serializer_tests.rs [hunk 1]
```diff
@@ -0,0 +1,58 @@
...
+
+            let test = Test {
+                int: 1,
+                seq: vec!["a", "b"],
+            };
+
+            let expected = list!(int=1, seq=list!("a", "b"));
+            assert_eq!(to_robj(&test).unwrap(), Robj::from(expected));
+        }
+    }
```
Ilia-Kosenkov, line 27: Perhaps I do not fully understand `serde` or our marshaling rules, but why `Vec<&'static str>` is serizlized as `list("a", "b")` and not `c("a", "b")`?

andy-thomason: Serde has its own marshalling rules and we don't get to see the type until the first array element arrives.

As a result, we collect everything as Robs and lists.

We could perform and extra step to convert lists of identical types to their vector types, but it is never
going to be as efficient as converting with a priori knowledge of the type.

We can look at this transformation as a later step.

andy-thomason:
There are perhaps three more things to do here:

* Implement deserialize to convert Robjs to structures.
* Implement Serialize for Robj so we can convert Robjs to JSON and other formats.
* Implement Deserialize for Robj so we can convert JSON and others to Robjs.

extendr-api/tests/serializer_tests.rs
Ilia-Kosenkov: Then I suggest writing it down somewhere -- in the docs or comments, for people like myself -- that if on the Rust side we work with `Vec<T>`, on the R side we should expect `list()`.

Ilia-Kosenkov:
@andy-thomason,
How difficult is it to implement deserialization from R objects to Rust structs (the opposite of what we now have)?
If it is not that difficult given your current progress, maybe it can be included in this PR and then we can test that the (de-)serialziation is roundtrippable.

Otherwise, I am ok with merging the PR as it is now.

I suggest to open a separate issue for JSON supoprt to track the progress there.

andy-thomason:
There are four classes in Serde:

* Serializer - done for now for Derive types
* Deserializer - to do for Derive types
* Serialize - to do for Robj
* Deserialize - to do for Robj.

A Deserializer is a bit more complex as it can take a broad set of inputs.

I need to read a little more on Serialize and Deserilaize.

We could also make our own `TryFrom<Robj>` and `Into<Robj>` proc macro
generator this would be more efficient than Serde.

```
#[Derive(ToRobj, FromRobj)]
struct { ... }
```

andy-thomason:
You are right that we should do this in parts and should merge this now.

Ilia-Kosenkov:
One last thing -- shouldn't `tests-all` feature also now include `serde`, together with `ndarray` and `use-bindgen`?

andy-thomason:
Yes!

Ilia-Kosenkov (approved):
Ok, no more questions. LGTM.

PR #302, opened 2021-10-11
Add comments to macros in scalar/macros.rs

ericwburden:
Since the comments I included in the macro from my last PR went over so well, I thought that adding similar comments to the other macros in `scalar/macros.rs` might be useful.


Ilia-Kosenkov:
@ericwburden,
Yeah, something like this, looks good!

Issue #297, opened 2021-10-06
Graphic Device methods and mutability


yutannihilation:
Currently all methods requires only `&self`. But, I feel some of them mutate the state of the device in actual (e.g. `new_page` creates a fresh page and increments the page number). Is it legal not to use `&mut self`?

https://extendr.github.io/extendr/extendr_api/graphics/struct.Device.html

clauswilke:
I think it's best not to pretend graphics devices are not mutable when they are. Not sure if this runs into some issues on the Rust side, though.

Ilia-Kosenkov:
Moving the discussion from #324 here. 
I proposed to research if `Drop` is applicable to `Device`. I personally like such resource management language features.
However, it raises a question, as I see two `Device` usage scenarios:
1. We fully control a device. We create, draw, and close it from Rust. In this scenario, `Drop` can represent the equivalent of `dev.off()`, because as soon as device handle in Rust goes out of scope, we also should dispose of the R device.
2. We temporarily get a handle to an existing (or newly created) device, In this case, we may not want to `dev.off()` it once the handle goes out of scope. This, I believe, is what @andy-thomason suggested, and it allows to put Rust code in the middle of a larger drawing pipeline.

Depending on what scenario we embrace, the implementaiton will differ, but it could be nice to discuss it first.

andy-thomason:
Absolutely, devices should be mutable as file handles are in Rust.

Currently, Device is only a reference to a device, so Drop does not make sense.

I would like to be able to *implement* devices in the future. In this case Drop is important
but for a different struct, like `DeviceImpl`.

andy-thomason:
We should look at a guard model to encapsulate mode.on and mode.off.

PR #324, opened 2021-11-22
Implement Device::close()

### LINKED ISSUE #299: Implement <Device>::close()
I think `dev.new()` needs to be called with `R!` ayway, but probably `dev.off()` can be done via C API; we can probably use `GEkillDevice`.

https://github.com/wch/r-source/blob/a7671ab01ac59e08e47dfa0aac0355739be8f0fe/src/library/grDevices/src/devices.c#L109-L114
https://github.com/wch/r-source/blob/a7671ab01ac59e08e47dfa0aac0355739be8f0fe/src/main/devices.c#L332-L335

yutannihilation:
Close #299 

I chose `close()` because `off()` might be confusing with `mode_off()`, but I'm open to any suggestions.

### DIFF
--- a/extendr-api/src/graphics/mod.rs
+++ b/extendr-api/src/graphics/mod.rs
@@ -374,6 +374,18 @@ impl Device {
         }
     }
 
+    /// Close the current device (equivalent to `dev.off()` on R).
+    pub fn close(&self) -> Result<()> {
+        unsafe {
+            if Rf_NoDevices() != 0 {
+                Err(Error::NoGraphicsDevices(r!(())))
+            } else {
+                GEkillDevice(self.inner());
+                Ok(())
+            }
+        }
+    }
+
     /// Get the device number for this device.
     pub fn device_number(&self) -> i32 {
         unsafe { GEdeviceNumber(self.inner()) }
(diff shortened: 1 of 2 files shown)


Ilia-Kosenkov:
It feels like `Device` can make use of `Drop`.  In my understanding, `Drop` is designed for managing limited resources, and works well with Rust lifetimes. This will give us a failsafe mechanism -- any `Device` will be eventually `drop()`ed once all its references are out of scope.
Of course, *if* there are no caveats and *if* `Drop` is suitable for our task.

yutannihilation:
Hm, sounds fair, thanks. I just wanted to moving away from using `R!("dev.off()")`, but it seems we'll want to move to "guard model", which we discussed on https://github.com/extendr/extendr/issues/297#issuecomment-941048527, anyway. I'm closing this for now. Let's discuss there.

Ilia-Kosenkov:
@yutannihilation, You should not close this PR IMO, perhaps I should not have started the discussion here and discouraged you. 

I believe it is a good solution for a problem for now, and it implements an obviously missing functionality, so I suggest to re-open it. 

I will move the broader discussion to the mentioned issue.

yutannihilation:
No worries, honestly, I wasn't motivated much to write this pull request. It's just that, because #299 is the issue I filed, I thought it's me who should fix it. 

Actually, on implementing this, I feel I'm getting lost in the graphic APIs in that what `dev.off()` does. I should have figured it out first before diving into this...

PR #327, opened 2021-11-23
Add Deref implementations for wrappers.

andy-thomason:
This PR adds derefs for wrapper types.

Vectors deref to slices of their value (`Rint /Rfloat/Rbool/Rstr/Robj`).

Some Vectors deref to mutable slices (`Rint/Rfloat/Rbool`) but not `Rstr` and `Robj` which need set_elt to work.

ExternalPtr<T> derefs to &T so we can use it like Box.

Rstr derefs to &str

We could deref Robj to &Rany - answers on a postcard.

### DIFF
--- a/extendr-api/src/lib.rs
+++ b/extendr-api/src/lib.rs
@@ -406,8 +406,8 @@ pub enum RType {
 }
 
 /// Enum use to unpack R objects into their specialist wrappers.
-/// Todo: convert all Robj types to wrappers.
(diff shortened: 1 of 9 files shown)


andy-thomason:
This will enable us to do:

```
fn fred(x: Integers) -> Rint {
   x[2]
}

fn fred(x: Integers) -> Rint {
   x[2]
}
```

yutannihilation:
Looks very useful!

PR #328, opened 2021-11-24
Fixing `NA_real_` comparison on windows-x86

### LINKED ISSUE #321: 32-bit Windows cannot handle Rfloat::na()
c.f. https://github.com/extendr/extendr/pull/318#issuecomment-964282455 

``` 
failures:

---- src\scalar\rfloat.rs - scalar::rfloat::Rfloat (line 35) stdout ----
Test executable failed (exit code 101).

stderr:
thread 'main' panicked at 'assertion failed: (<Rfloat>::na()).is_na()', src\scalar\rfloat.rs:6:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


---- src\wrapper\doubles.rs - wrapper::doubles::Doubles::elt (line 25) stdout ----
Test executable failed (exit code 101).

stderr:
thread 'main' panicked at 'assertion failed: vec.elt(10).is_na()', src\wrapper\doubles.rs:9:4
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

For this reason, one ALTREP-related test is disabled. Let's not forget to re-enable this as well.

https://github.com/extendr/extendr/pull/318/commits/473ac0597683aee9e2c2aae9497f8f1d794ceff3

Ilia-Kosenkov:
Closes #321.


andy-thomason:
I did check in a simple change to master that used the R function instead of the bit comparison.

Yours is probably better, but the test should be done in `f64::is_na()`. Also yours has documentation!

I suggest merging with master which is essentially a one-liner.

Ilia-Kosenkov:
I merged master because I had conflicts. 
I forgot about plain `f64`, I will revisit it.
Could you point me to the changes that you mention? So I can incorporate them in this solution.

andy-thomason:
Really just this: https://github.com/extendr/extendr/commit/c22729a48c73b36c73472fd61fe27e8518960411

Although I noticed that the test lacked a test! {} wrapper also.

Ilia-Kosenkov:
@andy-thomason,
I think your solution is the correct approach.
While the `NA_real_` constant is a fixed value, `is.na` chacks are more flexible. 
Essentially R uses `isnan()` from `math.h` to check if the value is `NaN` and then verifies that lower `DWORD` is 1954, so no need for precise bitwise comparison.
https://github.com/wch/r-source/blob/39ab860ff4db69b926ca51d2761b52f17aaef010/src/main/arithmetic.c#L126-L131
I will move all the logic to `f64` and use `f64` implementation in `Rfloat`.

andy-thomason (approved):
Looks great.

We can always look at using the bit compare for non-windows targets in the future
as the call to R will not vectorise.

PR #335, opened 2021-12-08
Allow renaming of R functions

andy-thomason:
This PR adds the ability to rename the R functions using the "r_name" attribute.

We also introduce the "mod_name" attribute which may be useful for specialisations
in the future.

Currently, this only works for non-impl functions.

```rust
#[extendr(
    use_try_from = true,
    r_name = "test.rename.rlike",
    mod_name = "test_rename_mymod"
)]
fn test_rename() {
}

extendr_module! {
    mod mymod;
    fn test_rename_mymod;
}
```


clauswilke:
I think this one would also do well with an integration test.

PR #338, opened 2021-12-08
Implement Logicals

andy-thomason:
This PR does the following:

Add Rbool as a type - Bool is still there but will be deprecated.
Add Logicals as a type.
Redesign the `gen_vector_wrapper_impl` macro a little, but keeping Ilia's original design.
Make sure everything works with scalar types including Integers and Doubles.
Implement ALTREP logicals support.
Add Logicals to Rany.
Fix *many* *many* tests.


extendr-api/src/robj/mod.rs
```diff
@@ -770,6 +770,7 @@ macro_rules! make_typed_slice {
     }
 }
 
+make_typed_slice!(Rbool, INTEGER, LGLSXP);
 make_typed_slice!(Bool, INTEGER, LGLSXP);
```
Ilia-Kosenkov, line 774: What would be the strategy for deprecating `Bool`?

extendr-api/src/scalar/macros.rs
```diff
@@ -743,7 +766,7 @@ macro_rules! gen_trait_impl {
             impl PartialEq<$type_prim> for $type {
                 /// NA always fails.
                 fn eq(&self, other: &$type_prim) -> bool {
-                    !self.is_na() && self.0 == *other
+                    <Option<$type_prim>>::try_from(self.clone()) == Ok(Some(*other))
```
Ilia-Kosenkov, line 769: While I can't find any combination of the inputs that produce an incorrect result, I feel like `Some(*other)` is not the best choice here. What if `other` contains a value that is `NA`? Then `Some(NA)` does not actually make sense, it should be `None`. Is it designed so only to ensure that `NA` always fails? Wouldn't it be better for maintainability to explicitly check for `NA` and then perform an equality check on `Result<Option<$type_prim>, ()>`? For that we need something like `impl TryFrom<Option<$type_prim>> for $type_prim`  though, and I am not sure it is a good idea.

extendr-api/src/scalar/rbool.rs
```diff
@@ -0,0 +1,71 @@
...
+
+impl Rbool {
+    gen_impl!(Rbool, i32);
+
+    pub fn t() -> Rbool {
+        Rbool(1)
+    }
+    pub fn f() -> Rbool {
+        Rbool(0)
+    }
```
Ilia-Kosenkov, line 22: I really do not like `t` and `f`, but I guess there is no better option. `T` or `F` are against Rust naming convention, and something like `r#true` and `r#false` look ugly.

Ilia-Kosenkov:
What about a few trivial tests on R side here https://github.com/extendr/extendr/blob/master/tests/extendrtests/src/rust/src/lib.rs?
E.g., test for `NA` and `elt` access, similar to `Integer` and `Doubles`?

extendr-api/src/robj/mod.rs
andy-thomason: Lets keep it for a while, but stop documenting it.

extendr-api/src/scalar/macros.rs
andy-thomason: If both are NA then try_from will produce Err() and the test will fail.

If NA comparisons always fail, this is correct.

extendr-api/src/scalar/rbool.rs
andy-thomason: I'm not a fan either.

If we can ditch Bool, we can re-purpose TRUE, FALSE and NA_BOOLEAN.

extendr-api/src/scalar/macros.rs
Ilia-Kosenkov: Ok, fair enough.

extendr-api/src/scalar/rbool.rs
andy-thomason: I am tempted to do that now, I must admit.

extendr-api/src/scalar/rbool.rs
```diff
@@ -14,12 +14,25 @@ pub struct Rbool(i32);
...
+    pub const fn t() -> Rbool {
         Rbool(1)
     }
-    pub fn f() -> Rbool {
+
+    pub const fn f() -> Rbool {
         Rbool(0)
     }
+
+    pub const fn na_const() -> Rbool {
```
Ilia-Kosenkov, line 25: But we have it as a part of `CanBeNA` trait? What's the difference?

andy-thomason: Surprisingly, trait methods can't be const!

andy-thomason: I'm working hard to avoid `Rbool(pub i32)`

andy-thomason: You could take a peek at:

https://docs.rs/num-complex/latest/num_complex/struct.Complex.html

We could have this on a feature.

extendr-api/src/scalar/rbool.rs
clauswilke: I'm in favor. Let's not add suboptimal naming to a new API because the optimal names clash with outdated parts. We're too early in the overall project development cycle.

As an alternative, though, you could also consider `true_value()` and `false_value()` or `get_true()` and `get_false()` or `logical_true()` and `logical_false()` or similar. Basically just combine true and false with one other word.

clauswilke:
I think dropping `Bool` now is fine. The current version of extendr is basically a complete rewrite of the previous version anyways, so I don't see any reason to keep backwards compatibility.

andy-thomason:
Good. I'll do this then.

andy-thomason:
We may need to review the current case that:

`NA_LOGICAL != NA_LOGICAL`

Because `PartialEq` never returns true for NAs on Rbool, Rint or Rfloat.

Ilia-Kosenkov:
> We may need to review the current case that:
> 
> `NA_LOGICAL != NA_LOGICAL`
> 
> Because `PartialEq` never returns true for NAs on Rbool, Rint or Rfloat.

Can you explain this? It seems that if `PartialEq` never returns `true` for `NA`, then `NA_LOGICAL != NA_LOGICAL`, or am I missing something?

clauswilke:
On the R side, you cannot compare an `NA` to an `NA`. The result is neither `TRUE` nor `FALSE` but rather `NA`. So returning `TRUE` when comparing `NA` to `NA` would definitely be problematic.

Ilia-Kosenkov:
I am aware of this feature of R. Why can't we treat all `NA`s as floating-point `NaN`s, i.e. any equality operation involving an `NA` results in `false`, even `NA == NA`. This is somewhat consistent with R implementation, as I doubt anyone treats logical `NA` obtained during comparison as `TRUE`.

clauswilke:
I agree, that's the most reasonable choice. It does imply `NA_LOGICAL != NA_LOGICAL`, though. Maybe I misunderstood what you were trying to say earlier. I though you considered it a problem.

Ilia-Kosenkov:
I was trying to say `NA != NA` is the approach I prefer. Sorry for confusion.

Ilia-Kosenkov (approved):

PR #345, opened 2021-12-25
Refactoring

andy-thomason:
This PR moves several things forward.

Debug is now supported directly for wrapper types.

Expression is renamed to Expressions.

Serialize now works for more types.

This seems like quite a big change, but it is all just refactoring moving towards our next formal release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## extendr devel
 
+- Extra Serialize support.
+
+- Direct Debug support for wrappers.
+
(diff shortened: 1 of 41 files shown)


andy-thomason:
```
==> Running `brew cleanup checkbashisms`...
Error: The operation was canceled.
```

I'm guessing this is a transient failure on actions behalf, perhaps a network failure.

Ilia-Kosenkov:
It seems you were unlucky and caught yet another transient error. I rescheduled tests, let's see if it works this time.

PR #347, opened 2021-12-28
Add IntoRList derive macro

### LINKED ISSUE #333: Derive a conversion trait from Rust struct to R list
So the existing way that `extendr` handles Rust structs seems to be by creating an `env` with a pointer to the struct, and then providing wrapper functions for the methods in the annotated `impl` block. This is a great baseline behaviour, but sometimes I want a different behaviour from my structs; to convert them recursively into an R data structure. For instance, if my function has a complicated result struct, rather than forcing my users to call getter functions, I would rather just convert the whole result struct into some native R types.

For this reason I'm wondering if we should create a `derive` attribute, such that we can:

```rust
#[derive(extendr_list)]
struct Foo {
    a: str
    b: usize
}
```

Which would expand to:

```rust
impl TryFrom<Foo> for Robj {
    fn try_from(value: Foo) -> Result<Robj, Error> {
            let mut ret = List::new(2);
            ret.set_elt(0, value.a.into())?;
            ret.set_elt(1, value.b.into())?;
            ret.as_robj_mut().set_names(["a", "b"])?.as_list()?
    } 
}
```
(there might be a more idiomatic way to create a named list but I don't know it)

multimeric:
* Closes #333.
* Refer to the documentation comments for more details.
* Feedback welcome at this stage


multimeric:
Okay this seems to work. The only issue I've noticed is that, since I'm implementing the conversion on the struct itself, converting to an R list consumes the original struct. I tried implementing it on the `&Struct` but then I have to borrow all the fields, and we don't have conversions from say a `&Vec<_> -> Robj`. So I ultimately left it.

multimeric:
So my test is passing, it's just some unrelated test failure that's breaking the CI.

extendr-macros/tests/test_derive_into_list.rs
```diff
@@ -0,0 +1,26 @@
...
+use extendr_macros::IntoRList;
+
+#[test]
+fn test_derive_list() {
+    test! {
+        #[derive(IntoRList)]
+        struct Foo {
+            a: usize,
+            b: String,
+            c: Vec<f64>
```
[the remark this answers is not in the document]
multimeric: Yeah so it will work with anything that we have implemented `Into<Robj>` for. I can demonstrate this in a test, sure.

multimeric:
Oh, I've just seen #355. Does that PR replace this? If it can do struct → Robj list then I'm happy to close in favour of that.

Ilia-Kosenkov:
It does something similar indeed, but is much more complicated. Let's wait for #355 and @andy-thomason's opinion.

andy-thomason:
Can I suggest we call this `IntoRobj` and plan to make a `TryFromRobj` derive also.

They should ultimately work for both `struct` and `enum`.

I have done some conversion work in the serde support, but that is perhaps less
useful generally than this. I had hoped that serde would be a better method as
serde is supported by many types, but the primitives that `Serializer` supports
are too basic to be generally useful.

Can we also have tests in the `extendr/tests/ directory`? I'm falling out of favour
of making doctests as the integration tests work better and faster.

multimeric:
> Can I suggest we call this IntoRobj

Although this is more-or-less implementing `Into<Robj>`, I wanted to name the *macro* something more specific, to make it clear that this wasn't a general purpose macro, but rather one that specifically created a new R list with the same data as your struct, and it didn't just return a pointer to Rust memory like many other `extendr` macros. Ultimately up to you, though.

> plan to make a TryFromRobj derive also

Good idea. I suppose I should make assumptions to allow for a round trip, namely that the Robj is a named list?

> They should ultimately work for both struct and enum.

How would you envisage this working for an enum? Would it just call `into()` on the inner type?

andy-thomason:
Ok. Keep the `IntoRList` name if you like. Derives tend to be named after the
trait they implement `Serialize`, `Deserialize` etc.

Currently, derives only exist for structs, enums and unions.

It would be worth checking what happens in this case:

#[derive(IntoRlist)]
struct A { ... {

#[derive(IntoRlist)]
struct B { x: Vec<A> }
```

**Comment by andy-thomason** [comment]
We may also want something to handle data frame rows at some point, a bit like the CSV
derive crate.

**Comment by andy-thomason** [comment]
Just as an aside.

I'm using this trick:

```
git fetch origin pull/347/head:into-r-list-macro
git checkout into-r-list-macro
```
To test and review @multimeric 's code.

**Comment by multimeric** [comment]
Okay, so I've implemented the reverse (R to any struct) as requested by @andy-thomason, and I think this is ready for review.

Something else I did was change the impl to work on borrowed structs/Robj. This is neat because it no longer consumes the inputs, but I had to hack into some of the extendr internals to do this. I'm pretty happy with `impl_from_iter! {&Vec<T>}` because this is the same behaviour we do with `Vec<T>` now. I'm less sure about `Into<Robj> for &List` because it ends up cloning the robj and I have no idea what that actually means. Is it just cloning the pointer, or is this cloning the entire data structure in memory?

**Comment by andy-thomason** [comment]
It is better to use `&Robj` if we can as it is zero-cost. Clones do just copy the
pointer, rather like `Rc` and `Arc` but the `Robj`s are preserved by adding the SEXP reference to a global list.
When we clone, we call `Robj::from_sexp(sexp)` on the pointer to add an extra reference.
This is not cost-free as the code is thread-safe(ish) and uses a hash map.

When a variable containing a `Robj` goes out of scope, the drop() method decrements
the reference count on the object.

I was planning to provide `TryFrom<&Robj> for T` for all types to make function
call overhead lower. Currently, we call `Robj::from_sexp(sexp)` on the incoming
arguments, but R protects these itself.

Sometimes TryFrom<Robj> is the correct answer, for example when wrapping a
generic Robj.

Slices of `&[Robj] and `&[Rstr]` from lists and strings are preserved by the owning vector
and don't need to be reference counted. You can get a `&Robj` for free from a list, but a clone
is more expensive.

**Comment by andy-thomason** [comment]
This table may help.

### Zero cost:

source|dest|reason
--|--|--
`Robj`|`List` etc|SEXP is moved.
`&Robj`|`&List` etc|use `robj.as_any()`
`List`|`Robj`|use `from()/into()`
`Robj`|`List`|use `try_from()/try_into()`
`&Integers`|`&[Rint]`|use like a slice
`Integers`|`&[Rint]`|use like a slice

### Full cost - clone and destruction.

source|dest|reason
--|--|--
`&Robj`|`List` etc|use `robj.as_any() and clone()`
`SEXP`|`Robj` etc|use `Robj::from_sexp()`

`Vec` will always have a full duplication cost although we
could wrap them in ALTREP.

**Comment by multimeric** [comment]
Right, this is quite confusing. Is there a 0-cost way to get `Robj` from `&List`? Because that's the scenario needed in this PR.

File: extendr-api/src/wrapper/list.rs [hunk 2]
```diff
@@ -405,6 +405,13 @@ impl Deref for List {
     }
 }
 
+// A borrowed list can be converted to an owned Robj by cloning the Robj
```
andy-thomason, line 408: We could define this globally for everything that supports the `GetSexp` trait.

```
impl<T : GetSexp> Into<Robj> for &T {
    fn into(self) -> Robj {
        self.get_robj().to_owned()
    }
}
```

This would cover all wrapper types.

extendr-macros/src/lib.rs
```diff
@@ -168,3 +169,43 @@ pub fn R(item: TokenStream) -> TokenStream {
...
+/// ```ignore
+/// let foo: Foo = R!("list(a = 5, b = 'bar')").unwrap().try_into().unwrap();
+/// ```
+/// # Details
+/// Note, the `From<Struct> for Robj` behaviour is different from what you get by applying the standard `#[extendr]` macro
+/// to an `impl` block. The `#[extendr]` behaviour returns to R a **pointer** to Rust memory, and generates wrapper functions for calling
+/// Rust functions on that pointer. The implementation from `#[derive(ConvertRList)]` actually converts the Rust structure
+/// into a native R list, which allows you to manipulate and access the internal fields, but it's a one-way conversion,
+/// and converting it back to Rust will produce a copy of the original struct.
+#[proc_macro_derive(ConvertRList)]
```
andy-thomason, line 208: We should have to separate derives.

I would suggest `ToList, ToRobj, FromList and FromRobj`

The List versions use `List` instead of `Robj`.

We may want to implement some of the traits manually.

extendr-macros/src/list_struct.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    // The only thing we emit from this macro is the conversion trait impl
+    TokenStream::from(quote!(
+        impl std::convert::From<&#struct_name> for Robj {
+            fn from(value: &#struct_name) -> Self {
+                extendr_api::List::from_pairs([#(#into_robj_tokens),*]).into()
+            }
+        }
+
+        impl std::convert::TryFrom<&Robj> for #struct_name {
```
andy-thomason, line 44: FromRobj should convert from both `&Robj and Robj`.

extendr-macros/src/lib.rs
multimeric: What is the advantage of a whole trait to convert to/from `List` when you can get one in one function call from a `Robj`?

extendr-api/src/wrapper/list.rs
multimeric: Unfortunately that gives me:

```
error[E0210]: type parameter `T` must be covered by another type when it appears before the first local type (`robj::Robj`)
   --> extendr-api/src/robj/into_robj.rs:582:6
    |
582 | impl<T :GetSexp> Into<Robj> for &T
    |      ^ type parameter `T` must be covered by another type when it appears before the first local type (`robj::Robj`)
    |
    = note: implementing a foreign trait is only possible if at least one of the types for which it is implemented is local, and no uncovered type parameters appear before that first local type
    = note: in this case, 'before' refers to the following order: `impl<..> ForeignTrait<T1, ..., Tn> for T0`, where `T0` is the first and `Tn` is the last
```

I guess we can't actually blanket implement for `T` when we only have a trait bound for `T` (which admittedly is a local trait, so I feel like Rust should allow this)

multimeric: Oh, it *also* gives:
```
error[E0119]: conflicting implementations of trait `std::convert::Into<robj::Robj>` for type `&_`
   --> extendr-api/src/robj/into_robj.rs:582:1
    |
582 | impl<T :GetSexp> Into<Robj> for &T
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: conflicting implementation in crate `core`:
            - impl<T, U> Into<U> for T
              where U: From<T>;
```

extendr-macros/src/lib.rs
multimeric: In the latest commit I have separated the derives, although I have not made a separate trait for `List`s.

extendr-api/src/wrapper/list.rs
andy-thomason: Rust doesn't yet have specialisation selection of overlapping traits and may never have!

Next best would be to include it in `make_conversions`.

multimeric: Done.

andy-thomason (approved):
Looks pretty good.

We should merge it.

extendr-macros/src/lib.rs
```diff
@@ -168,3 +169,67 @@ pub fn R(item: TokenStream) -> TokenStream {
...
+/// #[derive(TryFromRobj)]
+/// struct Foo {
+///     a: usize,
+///     b: String
+/// }
+/// ```
+/// Then, you can create an instance of your struct from an R expression:
+/// ```ignore
+/// let foo: Foo = R!("list(a = 5, b = 'bar')").unwrap().try_into().unwrap();
+/// ```
```
Ilia-Kosenkov, line 193: I think we have never used 'second person' in comments/examples. It might be better if it is 'us'/'ours' (if directed towards us the maintainers) or perhaps passive (like the next docs block) -- neutral and directed to consumers?

extendr-macros/tests/test_derive_into_list.rs
```diff
@@ -0,0 +1,48 @@
...
+        // We define the objects "natively", ie the struct in Rust, and the list in R
+        let native_rust = Foo {
+            a: 5,
+            b: String::from("bar"),
+            c: vec![1., 2., 3.],
+            d: list!(a = 1., b = true)
+        };
+        let native_r = R!("list(a = 5L, b = 'bar', c = as.double(1:3), d = list(a = 1, b = TRUE))").unwrap();
+
+        // Check the R → Rust conversion used a Robj reference
```
Ilia-Kosenkov, line 25: 'using'

multimeric:
I converted my examples to doctests because although they're a touch less readable now, at least their correctness is automatically checked now.

yutannihilation:
Why are the Windows checks failing? Do you need any help on investigating this? (in case you don't have Windows machine)

multimeric:
I haven't got a clue why they're failing. Any help would be appreciated, I can't easily develop on Windows.

yutannihilation:
I couldn't find the cause, but confirmed the error is reproducible on my Windows. Considering the same code can pass in tests module, I think this is some problem in cross-compiling feature of doctest. I'd suggest to skip the doctests on Windows with `cfg!()` for now. We can investigate later.

andy-thomason:
I merged it anyway as the Windows link problem seems to be a separate issues.

multimeric:
Yes I mean the macro was broadly correct, but I think this will cause all builds on master to fail until we sort out #372. Would you mind reading through that issue (sorry, yes it's long), and providing some input there? We're kind of at an impasse.

PR #348, opened 2021-12-28
Fix typo

fdrennan:
An, embarrassingly small pr to introduce myself. :)

Maybe small things like this can spark discussion around different parts of the code that nobody has talked about in a while - if anything comes to mind, let me know!

### DIFF
--- a/extendr-api/src/lib.rs
+++ b/extendr-api/src/lib.rs
@@ -252,7 +252,7 @@ pub use std::ops::DerefMut;
 pub use robj::Robj;
 
 //////////////////////////////////////////////////
-// Note these pub use statements are deprectaed
+// Note these pub use statements are deprecated
 //
 // `use extendr_api::prelude::*;`
 //


fdrennan:
Hopefully, this means that someone tells me the correct workflow, or points me to the doc I failed to read. :P

Ilia-Kosenkov:
We have a standard [CONTRIBUTING.md](https://github.com/extendr/extendr/blob/master/CONTRIBUTING.md). As a general recommendation, it is better to first open an issue for discussion and then submit PRs.
My personal recommendation would be to perhaps first list all of the small issues that you have found (e.g. in the issue) and then submit one PR fixing several typos instead of making one PR per each line fixed. It is easier to review and keep track of the changes.
I will launch CI checks on both of your PRs.

fdrennan:
Will open issue first :)

yutannihilation:
Oh, sorry for the late reply.

@Ilia-Kosenkov 
Do we really require an issue for fixing a single typo? If we are to review some length of documents, it's worth an issue for discussion, but I think we can just merge a simple typo fix that has no room for discussion.

Ilia-Kosenkov:
No, but I believe things like typos can be fixed en masse within several larger PRs instead of many one-line fixes. Plus it is a good practice anyway to first file an issue and then a PR, especially for a new contributor. Also, there is no need to file an issue per typo, but one place of discussion that can be referenced is a good idea I think.

yutannihilation:
My feeling is that fixing a typo is usually another topic than fixing another typo, and every pull request should only have a single topic. Actually, `CONTRIBUTING.md` states:

> In general, PRs should:
> * Address a single concern in the least number of changed lines as possible.

yutannihilation:
But, I get your point. I'm fine with the rule for now. Let's revisit here when needed.

Ilia-Kosenkov:
I am not trying to create a rule, but it is just a matter of quantity. I'd personally prefer to have 10 typos fixed in one PR rather than in 10 PRs. But I am OK with any approach and did not ask to close this PR.

yutannihilation:
Yeah, I agree it's a matter of quantity. I feel 10 is a bit too much considering the amount of our documentations, but it sounds good to have such a criteria. Let's see how it goes in https://github.com/extendr/extendr/issues/353.

@fdrennan 
Feel free to try finding 10 typos :)

PR #350, opened 2021-12-28
Complexes

andy-thomason:
This PR implements the Complexes type and provides complex number support via num_complex.

`Rcplx` wraps `num_complex::Complex<f64>` the same way that `Rfloat` wraps f64. 

```
fn mycplx(x: Rcplx) -> Rcplx {
  -x
}

fn mycplxvec(x: Complexes) -> Complexes {
  x
}

fn mycplxvec2() -> Complexes {
  Complexes::from_values([(0., 0.), (1., -1.)])
}
```

This PR provides only basic support.


extendr-api/src/scalar/rcplx.rs
```diff
@@ -0,0 +1,186 @@
+use crate::scalar::macros::*;
+use crate::scalar::Rfloat;
+use crate::*;
+use std::convert::TryFrom;
+use std::ops::{Add, Div, Mul, Neg, Sub};
+use std::ops::{AddAssign, DivAssign, MulAssign, SubAssign};
+
+// #[cfg(feature="num-complex")]
+pub type C64 = num_complex::Complex<f64>;
```
Ilia-Kosenkov, line 9: I know it is against the naming convention, but shouldn't it be `c64` inline with `f64` and `i64`/`u64`?

extendr-api/src/wrapper/complexes.rs
```diff
@@ -0,0 +1,137 @@
...
+    primitive_type: C64,
+    r_prefix: COMPLEX,
+    SEXP: CPLXSXP,
+    doc_name: complex,
+    altrep_constructor: make_altcomplex_from_iterator,
+);
+
+impl Complexes {
+    /// Get a region of elements from the vector.
+    pub fn get_region(&self, index: usize, dest: &mut [Rcplx]) -> usize {
```
Ilia-Kosenkov, line 35: It feels like this can be a macro. Vector types other than `Logicals` may benefit from it as well.

extendr-api/src/wrapper/mod.rs
```diff
@@ -197,6 +199,7 @@ make_conversions!(S4, ExpectedS4, is_s4, "Not a S4 type");
 make_conversions!(Integers, ExpectedInteger, is_integer, "Not an integer type");
 make_conversions!(Logicals, ExpectedLogical, is_logical, "Not a logical type");
 make_conversions!(Doubles, ExpectedReal, is_real, "Not a floating point type");
+make_conversions!(Complexes, ExpectedComplex, is_complex, "Not a complex type");
```
Ilia-Kosenkov, line 202: Maybe 'complex number'? 'Complex type' can be misleading.

extendr-api/Cargo.toml
```diff
@@ -22,12 +22,14 @@ ndarray = { version = "0.15.3", optional = true }
 lazy_static = "1.4"
 paste = "1.0.5"
 serde = { version = "1.0", features = ["derive"], optional = true }
+num-complex = { version = "0.4", optional = true }
 
 [features]
-default = ["serde"]
+# change this to [] in master.
+default = ["num-complex"]
```
Ilia-Kosenkov, line 29: Is this something that should be fixed before merging?

extendr-api/src/scalar/rcplx.rs
```diff
@@ -0,0 +1,186 @@
+use crate::scalar::macros::*;
+use crate::scalar::Rfloat;
+use crate::*;
+use std::convert::TryFrom;
+use std::ops::{Add, Div, Mul, Neg, Sub};
+use std::ops::{AddAssign, DivAssign, MulAssign, SubAssign};
+
+// #[cfg(feature="num-complex")]
```
Ilia-Kosenkov, line 8: Should this be uncommented?

extendr-api/src/scalar/rcplx.rs
andy-thomason: Rust likes types to start with a capital letter. But we can override this with a #[allow()]

extendr-api/src/wrapper/complexes.rs
andy-thomason: It was originally a macro, but Doubles and Complexes differ in the methods they support.

extendr-api/src/wrapper/mod.rs
andy-thomason: Ok.

extendr-api/Cargo.toml
andy-thomason: Yes. I intend to write a simple alternative that does not use num-complex
as the default. I am mindful that every crate we add as a non-optional dependency
costs on build times.

extendr-api/src/scalar/rcplx.rs
andy-thomason: The next step is to make rcplx_full.rs and rcplx_default.rs

These will be two alternative implementations with varying levels of support.

andy-thomason: Note that num-complex also supports a wide range of operations on complex numbers such as logs,
trig, arg and magnitude.

extendr-api/src/wrapper/mod.rs
andy-thomason: Also, next step is to ditch `FromRobj`

extendr-api/src/scalar/rcplx.rs
Ilia-Kosenkov: I know, I am just curious and asking for opininons. I am ok with either options.

andy-thomason:
Are you good with this @clauswilke @yutannihilation @Ilia-Kosenkov ?

clauswilke:
I gave this a quick look and it generally seems fine to me. One issue that caught my eye is `Rcomplex` vs. `Rcplx`. It's not immediately clear to me why there are two types, which one an enduser would be supposed to use, and whether the difference is documented. I only found some documentation for `Rcplx`.

If there is a good reason for having these two types, then I think this should be documented somewhere. Also, users should be told which one to use when.

andy-thomason:
> One issue that caught my eye is `Rcomplex` vs. `Rcplx`.

Currently, we either wrap num_complex or a simple struct with the same layout in Rcplx.

num_complex gives us complex arithmetic, functions, nan tests and more but
is not the same as Rcomplex even though it has the same layout.

We could have implemented arithmetic for Rcomplex, but it would have been a lot
more work and would involve a lot of `transmute` calls.

clauswilke:
> Currently, we either wrap num_complex or a simple struct with the same layout in Rcplx.

That's fine, and I have no concerns with the technical aspects, but I think this should be stated clearly in the docs somewhere.

If I understand correctly, endusers would never normally use `Rcplx`, only `Rcomplex`. So I would suggest adding these thoughts to the documentation:

- In the documentation for `Rcplx`, mention that endusers would typically use `Rcomplex`.
- In the documentation for `Rcomplex`, mention that it wraps two different types, depending on use case. Also mention how this can be configured via features.

yutannihilation (approved):
I'm not very familiar with complex math, but this generally looks good to me.

PR #351, opened 2021-12-30
Polars wrapper

### LINKED ISSUE #331: `data.frame` → Polars `DataFrame` wrapper
I'm wondering if there is interest in a feature-gated conversion between Polars and R data frames, in the same way we are doing for `ndarray`. So the module would implement the appropriate To/From Robj traits on `DataFrame`.

multimeric:
* Closes #331 
* Includes:
    * Polars -> R conversion
    * A test
* TODO:
    * R -> Polars conversion
    * Conversion for non-numeric Series types


multimeric:
I'm closing this in favour of #357 which is much more complete and more robust, but if this somehow ends up being helpful for someone I could finish it.

PR #355, opened 2022-01-05
Serde Deserialize and Deserializer conversion from Robjs.

andy-thomason:
This PR adds Deserialize and Deserializer for Robjs.

This enables any objects declared with `#[derive(Deserialize)]` to be converted
from a Robj.

We can also make an Robj from JSON and other formats.

This is the counterpoint for Serialize and Serializer.

Example:

```
            #[derive(Deserialize, PartialEq, Debug)]
            struct Struct {
               a: i32,
               b: f64
            }
            assert_eq!(from_robj::<Struct>(&r!(list!(a=1, b=2))), Ok(Struct{ a: 1, b: 2.0 }));
```

Note that this is somewhat similar to #347

### DIFF
--- a/extendr-api/tests/deserializer_tests.rs
+++ b/extendr-api/tests/deserializer_tests.rs
@@ -126,6 +126,19 @@ mod test {
             let j = r!(list!(Struct=list!(a=1)));
             let expected = Enum::Struct { a: 1 };
             assert_eq!(expected, from_robj(&j).unwrap());
+
+            // Many things will generate a Robj.
+            // But note that the original Robj will not be copied verbatim.
+            // The Deserialize trait for Robj can also be used to generate
+            // JSON and other formats for Robj.
+            #[derive(Deserialize, PartialEq, Debug)]
+            struct ROBJ(Robj);
+            assert_eq!(from_robj::<ROBJ>(&r!(TRUE)), Ok(ROBJ(r!(TRUE))));
+            assert_eq!(from_robj::<ROBJ>(&r!(1)), Ok(ROBJ(r!(1))));
+            assert_eq!(from_robj::<ROBJ>(&r!(1.0)), Ok(ROBJ(r!(1.0))));
+            assert_eq!(from_robj::<ROBJ>(&r!("xyz")), Ok(ROBJ(r!("xyz"))));
+
+            // assert_eq!(from_robj::<ROBJ>(&r!([TRUE, FALSE])), Ok(ROBJ(r!([TRUE, FALSE]))));
         }
     }
 }
--- a/extendr-api/tests/deserializer_tests.rs
+++ b/extendr-api/tests/deserializer_tests.rs
@@ -5,10 +5,19 @@ mod test {
     use serde::Deserialize;
 
     ////////////////////////////////////////////////////////////////////////////////
-
+    ///
+    /// Deserialize from a Robj.
+    /// 
+    /// Like JSON, we can use a Robj as a storage format.
+    /// 
+    /// For example if creating vectors from a RDS file or returning a structure.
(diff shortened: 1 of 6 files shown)


extendr-api/tests/deserializer_tests.rs
```diff
@@ -0,0 +1,174 @@
...
+            #[derive(Deserialize, PartialEq, Debug)]
+            struct Null;
+            assert_eq!(from_robj::<Null>(&r!(NULL)), Ok(Null));
+            assert_eq!(from_robj::<Null>(&r!(1)), Err(Error::ExpectedNull(r!(1))));
+
+            #[derive(Deserialize, PartialEq, Debug)]
+            struct Int(i32);
+            assert_eq!(from_robj::<Int>(&r!(1)), Ok(Int(1)));
+            assert_eq!(from_robj::<Int>(&r!(1.0)), Ok(Int(1)));
+            assert_eq!(from_robj::<Int>(&r!(NULL)).is_err(), true);
```
Ilia-Kosenkov, line 31: Is there a reason for using `assert_eq!` when testing functions returning `bool`? Isn't it easier to use `assert!(expr)`?

extendr-api/tests/deserializer_tests.rs
```diff
@@ -0,0 +1,174 @@
...
+            // Sequences are always converted to lists.
+            assert_eq!(from_robj::<ROBJ>(&r!([TRUE, FALSE])), Ok(ROBJ(r!(list!(TRUE, FALSE)))));
+            assert_eq!(from_robj::<ROBJ>(&r!([1, 2])), Ok(ROBJ(r!(list!(1, 2)))));
+
+            // If you use a wrapper type, conversions are more specific.
+            #[derive(Deserialize, PartialEq, Debug)]
+            struct RIntegers(Integers);
+            assert_eq!(from_robj::<RIntegers>(&r!(1)), Ok(RIntegers(Integers::from_values([1]))));
+            assert_eq!(from_robj::<RIntegers>(&r!([1, 2])), Ok(RIntegers(Integers::from_values([1, 2]))));
+            assert_eq!(from_robj::<RIntegers>(&r!(1.0)).is_err(), true);
```
Ilia-Kosenkov, line 150: I am not sure I understand why we can deserialize `r!(1.0)` as `Rint` but not as `Integers` (see assertion on line 36).

Ilia-Kosenkov:
It is hard to review because of numerous changes, but the test coverage shows the implementation is likely correct.
I remember I asked when you implemented Serialize if roundtripping is possible, so I will ask again - is roundtripping possible (at least for primitive types)? I believe it is quite important. If it is possible, perhaps we can have a few tests that call `.from_robj()` and `.to_robj()` and compare results?

andy-thomason:
The whole effect is that we can use serde to do conversions to/from R objects.

The only downside with serde is that I can't see a way to "bypass" the serializer
and convert Robjs to other Robjs directly. I may find a way to do this, for example
using the custom bounds.

Writing a custom Derive as in #347 would be more direct for conversion and
we should support both methods.

andy-thomason:
I'm going to merge this if no one objects too much @Ilia-Kosenkov in particular.

We can refine it in the future, but it may not fly as a conversion method. It should cover
an awful lot of "foreign" type cases where we don't or can't have TryFrom/Into.

andy-thomason:
As the next PR, I'll provide a wrapper for the R load/save to std::io::Read/Write.

Ilia-Kosenkov (approved):

yutannihilation:
Can we merge this?

andy-thomason:
Yes, I think its good to go.

PR #357, opened 2022-01-12
Polars Arrow Wrapper

### LINKED ISSUE #331: `data.frame` → Polars `DataFrame` wrapper

multimeric:
* Closes #331
* Features
    * Adds an `impl` for converting a Polars `DataFrame` into a `RecordBatch` in R
    * Adds an `impl` for converting a `RecordBatch` in R into a Polars `DataFrame`
    * This requires the `arrow` package to be installed in R, but there is a check for this
    * Includes a thorough test for each conversion
* Run tests using `cargo test --features polars-wrapper --package extendr-api --lib wrapper::polars ` 
* TODO
    * [x] The reverse process: R → Rust
    * [x] Once `polars` and `arrow` have pushed stable releases with the features I'm using, use these instead of the git dependencies


multimeric:
Can anyone help me with finding a good way to install the `arrow` package on the CI? I normally use conda, and it's a bit tricky to install on several different OSes (since it requires some non-R dependencies).

andy-thomason (commented):
I would expect us to make a separate crate for this, perhaps part of the polars
library itself, otherwise we are going to get quite crowded with all our integrations.

It would be worth looking at some PyO3 integrations for comparision.

andy-thomason:
We  are also reserving `DataFrame` for a `robj` wrapper in the future.

yutannihilation:
> I would expect us to make a separate crate for this, perhaps part of the polars
> library itself, otherwise we are going to get quite crowded with all our integrations.

Yeah, I too feel so. It seems the consensus of Andy, [Ilia](https://github.com/extendr/extendr/issues/331#issuecomment-1008667859) and I is not to have this Polars wrapper within extendr-api itself (while my take is a bit different in that I'm rather hesitant to have dataframe-related features at all in fear of feature bloat). Can we close this pull request?

Again, to be clear, I don't mean to say the Polars wrapper is worthless at all. Rather, I'd like to see more extension crates like this developed outside of extendr org. Actually, I have several ideas of such extensions in my mind. extendr should mention these 3rd party crates on README (e.g. "Extentions" section).

> Can anyone help me with finding a good way to install the `arrow` package on the CI?

You can use RSPM to install binaries for Linux. See https://github.com/r-lib/actions/tree/v2/setup-r for `use-public-rspm` option.

PR #359, opened 2022-01-13
Register `[[.rust_class` just as `$.rust_class`

shrektan:
Hi,

I believe it's necessary to register `[[` S3 method for exported Rust Class as well. 

The reason is that, only so, we can call the method dynamically, e.g., `method <- "method_abc"; RustClass[[method]]()`. 

This is not possible for `$`.

Thanks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## extendr devel
 
+- Register S3 method `[[.rust_class` just as `$.rust_class`.
+
 - Extra Serialize support.
 
 - Direct Debug support for wrappers.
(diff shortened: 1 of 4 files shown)


yutannihilation:
Sounds good to me.

andy-thomason:
This is great. Good to update the Change log, too.

We may export trait methods to s3 methods in the future in which case we may optionally overload `[[` as `Index`, but we
can cross that bridge when we come to it.

yutannihilation:
Thanks for updating. One more thing, could you add a small test to verify `x[["foo"]]()` works as the same as `x$foo()` around here?

https://github.com/extendr/extendr/blob/a59d65c50e001810dadba2d8624517a36b4725d2/tests/extendrtests/tests/testthat/test-classes.R#L5

shrektan:
@yutannihilation added.

yutannihilation (approved):
Looks good, thanks! I'll wait for a while before merging to see if there's any concern from other maintainers.

yutannihilation:
I'm merging now. Thanks for the contribution!

PR #360, opened 2022-01-15
Create a graphic device

yutannihilation:
#279 adds functions to control a graphic device, but extendr don't have the functionality to create a graphic device yet. This pull request is an attempt to implement it. Note that this is at the very early stage, and I'm not sure I can successfully finish this.

Btw, I'm wondering if we should move the graphic device-related implementations to a separate crate because probably most of us don't need this.


clauswilke:
> Btw, I'm wondering if we should move the graphic device-related implementations to a separate crate because probably most of us don't need this.

I think this makes a lot of sense. Just like extendr-engine. There's no reason to have this be part of the regular API, and it's easy enough to include if somebody wants it.

yutannihilation:
Thanks, then I'll address it in a separate pull request (because this pull request is already a bit massive...). Let's file an issue for now.

andy-thomason:
This looks good.

It would probably be worth defining a trait like:
```
impl DeviceDriver {
    fn abc();
    fn def();
}
```

and have the default impl connect the callbacks to the functions as we do in Altrep.

If you write defaults for many of the functions, say converting everything into path, stroke and text
primitives, it should be easy to write a device driver in a few lines of code; we could do SVG as
a simple example.

yutannihilation:
Thanks, I was thinking if I can use a Trait here. One thing I'm still not sure is how to left the optional methods unimplemented, instead of filling empty default functions. Are there any trick to know whether the method is actually implemented? Can I compare the function pointers?

andy-thomason:
In the Altrep version I just generated C functions in the class constructor and mapped all of them.

Many of the functions have defaults with only elt and length being required.

For example here

https://github.com/extendr/extendr/blob/master/extendr-api/src/wrapper/altrep.rs#L51-L71

we make a default and here

https://github.com/extendr/extendr/blob/master/extendr-api/src/wrapper/altrep.rs#L538-L686

We create C functions to call the trait methods.

I will have to profile the Altrep class creation at some point to see if it needs to be cached, but with a device
driver, I doubt it would matter.

yutannihilation:
> mapped all of them.

Unlike ALTREP, many of the graphic device methods are optional, and I feel it's supposed to be left unimplemented if the device doesn't have that functionality.

e.g. https://github.com/wch/r-source/blob/9f284035b7e503aebe4a804579e9e80a541311bb/src/main/engine.c#L1499-L1502

yutannihilation:
Note to self: if a `panic!()` occurs inside a callback function, it seems it crashes the R session. I need to add some note on this.

yutannihilation:
Reading the source code of R, I'm now feeling the default implementation should make no problem in most of the cases. If we find some methods that needs opt-out, we can add a flag as an associated const flag.

yutannihilation:
Though this is far from perfect, I think this is ready for review now. Feel free to comment. For simplicity, I'll separate a pull request to support newer features. This one implements only the basic methods.

Here's an example code.

### Rust code

``` rust
use extendr_api::{
    graphics::{device_descriptor::DeviceDescriptor, device_driver::DeviceDriver, DevDesc},
    prelude::*,
};

struct YadnDevice<'a> {
    welcome_message: &'a str,
}

impl<'a> DeviceDriver for YadnDevice<'a> {
    fn activate(self, _dd: DevDesc) {
        let welcome_message = self.welcome_message;

        rprintln!("🎉🍕🍰📺🍓✨🍣🐈🎿🎉🍕🍰📺🍓✨🍣🐈🎿");
        rprintln!("");
        rprintln!("   {welcome_message}   ");
        rprintln!("");
        rprintln!("🎉🍕🍰📺🍓✨🍣🐈🎿🎉🍕🍰📺🍓✨🍣🐈🎿");
    }
}

/// Return string `"Hello world!"` to R.
///
/// @export
#[extendr]
fn yadngd(welcome_message: String) -> i32 {
    let device_driver = YadnDevice {
        welcome_message: welcome_message.as_str(),
    };

    let device_descriptor = DeviceDescriptor::new();
    let device = device_driver.create_device::<YadnDevice>(device_descriptor, "yadndgd");
    device.device_number()
}

// Macro to generate exports.
// This ensures exported functions are registered with R.
// See corresponding C code in `entrypoint.c`.
extendr_module! {
    mod yadngd;
    fn yadngd;
}
```

### R

``` r
yadngd(welcome_message = "◆祝◆ device activated!!! ◆祝◆")
#> 🎉🍕🍰📺🍓✨🍣🐈🎿🎉🍕🍰📺🍓✨🍣🐈🎿
#> 
#>    ◆祝◆ device activated!!! ◆祝◆   
#> 
#> 🎉🍕🍰📺🍓✨🍣🐈🎿🎉🍕🍰📺🍓✨🍣🐈🎿
#> [1] 1
```

andy-thomason:
Looks pretty cool :cool:

yutannihilation:
Thanks! The ALTREP implementation was really helpful.

yutannihilation:
I think the design is almost ready now. The main tasks left now is to add more tests (including the integrated tests) and documents...

andy-thomason:
Getting there.

yutannihilation:
I think I need a better understanding on the device capabilities. I filed a separate issue: #367.

yutannihilation:
@andy-thomason 
Now that I added a minimal integrated test and minimal module level document, it's ready for your review now. Could you review when you have time?

@clauswilke 
I think you are interested in graphic devices, so I'd like to hear your thoughts on the design. A simple example code can be found here: https://github.com/extendr/extendr/pull/360/files#diff-0277d1da8af341ba6cd25fc277d1b8602b9725f6640e86c4c55287372fdf7029R51-R77

@Ilia-Kosenkov 
I'm not sure the graphic device related things are within your main interest, but I'd appreciate your comments if you have time to take a look.

yutannihilation:
Curious. For some reason the integration test fails on R-release on Windows. It seems the failure is reproducible on my local, but cannot reproduce if I use R-devel...

Ilia-Kosenkov:
I confirm the R process crashes upon starting graphics device tests in `tests/extendrtests` (R-4.1.2x64 @ Windows).

yutannihilation:
Thanks for confirming. It seems to crash on closing the device (`dev.off()`). I tried removing the `close()` function at all, but it still crashes. I have no idea about the cause...

Ilia-Kosenkov:
I tried print-debugging by inserting `println!()` into each of the device functions you define in extendr-api. I managed only like 4-5 main functions. This may help to locate the problematic function, but it is a lot of work.

yutannihilation:
Thanks. I think the problem is not in the content of any of the device functions. If I put `std::fs::write("foo.txt", "foo!")` in `close()`, the `foo.txt` is created so probably the problem happens after the line:

```c
	}
	g->dev->close(g->dev);
	GEdestroyDevDesc(g);
	R_Devices[devNum] = NULL;
    }
}
```
https://github.com/wch/r-source/blob/6b22b60126646714e0f25143ac679240be251dbe/src/main/devices.c#L326

But most of the lines around here are from more than 10 years ago and I couldn't find any changes that might affect here.

Ilia-Kosenkov:
I am unfortunately very bad at debugging unmanaged code, but I will make another attempt. 
It is very hard to debug R calling into Rust, perhaps you can reproduce this problem the other way around -- running Rust and starting an R session inside it? Will you be able to call your custom graphics functions in such a setup?

yutannihilation:
Good idea, I added `dev.off()` on the test on Rust's side. This succeeds with R-devel on my local but crashes when it's with R-4.1.2.

yutannihilation:
To prove I'm sane, I added a Windows runner with R-devel. It should pass without errors...

yutannihilation:
I just learned how to get a crash dump and the stack trace using WinDbg. It seems the error is that it tried to free an already-freed heap (code c0000374 means `STATUS_HEAP_CORRUPTION`).

```
(5d38.5c80): Unknown exception - code c0000374 (!!! second chance !!!)
ntdll!RtlReportFatalFailure+0x9:
00007ffb`3df0be99 eb00            jmp     ntdll!RtlReportFatalFailure+0xb (00007ffb`3df0be9b)
0:000> k
 # Child-SP          RetAddr               Call Site
00 00000000`0441cb70 00007ffb`3df0be63     ntdll!RtlReportFatalFailure+0x9
01 00000000`0441cbc0 00007ffb`3df14c82     ntdll!RtlReportCriticalFailure+0x97
02 00000000`0441ccb0 00007ffb`3df14f6a     ntdll!RtlpHeapHandleError+0x12
03 00000000`0441cce0 00007ffb`3df1f455     ntdll!RtlpHpHeapHandleError+0x7a
04 00000000`0441cd10 00007ffb`3df14b90     ntdll!RtlpLogHeapFailure+0x45
05 00000000`0441cd40 00007ffb`3de26865     ntdll!RtlpAnalyzeHeapFailure+0x2fc
06 00000000`0441cda0 00007ffb`3de281fe     ntdll!RtlpFreeHeap+0xcc5
07 00000000`0441cf00 00007ffb`3de27551     ntdll!RtlpFreeHeapInternal+0x79e
08 00000000`0441cfc0 00007ffb`3c30c6ac     ntdll!RtlFreeHeap+0x51
09 00000000`0441d000 00000000`6c7bcf99     msvcrt!free+0x1c
0a 00000000`0441d030 00000000`6c79e7de     R!GEdestroyDevDesc+0x59
0b 00000000`0441d080 00000000`6fc828e9     R!GEcurrentDevice+0x37e
0c 00000000`0441d0f0 00000000`6c7a15fa     grDevices!devoff+0x59
0d 00000000`0441d130 00000000`6c7ed189     R!Rf_NewFrameConfirm+0x1eca
0e 00000000`0441d600 00000000`6c7fcbf1     R!R_initAssignSymbols+0x72c9
0f 00000000`0441e200 00000000`6c7fe907     R!Rf_eval+0x171
10 00000000`0441e4a0 00000000`6c7ffb6a     R!R_cmpfun1+0x4a7
11 00000000`0441e810 00000000`6c7fcd9c     R!Rf_applyClosure+0x1ba
12 00000000`0441e8a0 00000000`6c827291     R!Rf_eval+0x31c
13 00000000`0441eb40 00000000`6c827601     R!Rf_ReplIteration+0x2a1
14 00000000`0441eba0 00000000`6c82769d     R!Rf_ReplIteration+0x611
15 00000000`0441fc20 00000000`6c82773e     R!run_Rmainloop+0x4d
16 00000000`0441fc50 00000000`0040171a     R!Rf_mainloop+0xe
17 00000000`0441fc80 00000000`0040158a     Rterm+0x171a
18 00000000`0441fe00 00000000`004013c5     Rterm+0x158a
19 00000000`0441fe30 00000000`0040152b     Rterm+0x13c5
1a 00000000`0441ff00 00007ffb`3c5154e0     Rterm+0x152b
1b 00000000`0441ff30 00007ffb`3de0485b     KERNEL32!BaseThreadInitThunk+0x10
1c 00000000`0441ff60 00000000`00000000     ntdll!RtlUserThreadStart+0x2b
```

As it says `GEdestroyDevDesc`, probably the problem happens either of these `free()`. But I still have no idea why it happens only on R-release on Windows.

``` c
void GEdestroyDevDesc(pGEDevDesc dd)
{
    int i;
    if (dd != NULL) {
	for (i = 0; i < MAX_GRAPHICS_SYSTEMS; i++) unregisterOne(dd, i);
	free(dd->dev);
	dd->dev = NULL;
	free(dd);
    }
}
```
https://github.com/wch/r-source/blob/8ebcb33a9f70e729109b1adf60edd5a3b22d3c6f/src/main/engine.c#L78-L87

yutannihilation:
I asked for help on the R-package-devel mailing list.

https://stat.ethz.ch/pipermail/r-package-devel/2022q1/007644.html

While I still appreciate any help on this, I think I can find the cause eventually, and, even if I cannot, I think it's acceptable to require R >= 4.2 on graphics-related features. Sorry for the noise...

yutannihilation:
Unfortunately, I found another error when using grid, which is reproducible on other environments.

After struggling with gdb, it seems the error happens around `<pGEDevDesc>->gesd[]->systemSpecific` struct in `doSetViewport()`. Sorry, it seem I still have a long way to get this ready...

```
(gdb) n
353             UNPROTECT(1);
(gdb) n
358         if (LOGICAL(gridStateElement(dd, GSS_RESOLVINGPATH))[0]) {
(gdb) s
gridStateElement (dd=dd@entry=0x1654cfb3c80,
    elementIndex=elementIndex@entry=16) at state.c:154
154         return VECTOR_ELT((SEXP) dd->gesd[gridRegisterIndex]->systemSpecific,
```

- https://github.com/wch/r-source/blob/9f284035b7e503aebe4a804579e9e80a541311bb/src/library/grid/src/grid.c#L358
- https://github.com/wch/r-source/blob/2269cd94d37af1957ef54162f5ae5cde87e31c23/src/library/grid/src/state.c#L152-L156

Ilia-Kosenkov:
Don't be discouraged, graphics support is important for R <-> Rust interop.
I am unfamiliar with the device internals, so cannot really help you here.

Ilia-Kosenkov:
Ok, here is a wild guess -- mark all the structs that are somehow used on C side with `#[repr(C)]`. Rust does not guarantee the struct layout unless you explicitly demand it. I doubt it will help but it *may* solve at least some problems.

yutannihilation:
Thanks for the advice. Sorry, my comment above was just my gdb-fu was not enough to identify the line of the error... It happened in `setMask()` simply because it tried to call the non-existent function `setMask`. I was thinking these functions are optional, but my assumption was wrong. I added placeholder functions to fix.

yutannihilation:
Using `libc::malloc()` *seems to* solve the problem on my local on real R session so I feel this is the right direction. But, it won't on tests on Rust, as all the GitHub action CI fail.

yutannihilation:
Hmmmm... I think the CI result of the recent two commits confirmed the problem is somehow related to the mismatch of the memory allocator. I see both commits are fine when I use it when compiled within an R package. The error happens only on Rust tests.

1. Using `libc::calloc()`: fails on R-devel on Windows on Rust tests.
2. Using `R_calloc_gc()`: fails on R-release on Windows on Rust tests.

IIUC, in case 1, it depends on the compile settings what allocator will be used, so it can fail. But, I have no idea why it fails on case 2. Are there any possibility that `R_calloc_gc()` can use different allocator than the one that R uses...?

Ilia-Kosenkov:
It is good that we now have some understanding of what is causing the problem. Yet I am concerned that it is so fragile and depends on allocators. Even if we find a combination of settings and allocators that do not fail on CI/our local machines, we cannot be sure our consumers will not experience random heap corruptions and segfaults. 
May I suggest writing a lot of thorough tests (both Rust and integration through R)? If they all succeed on R-devel, I see no problem in requiring `4.2.0` across all platforms in order to be able to use graphics. At least for now. We can then merge version-guarded PR and perhaps later investigate the root cause of this issue (and perhaps gather some user feedback from different setups).

yutannihilation:
I agree currently it's fragile. I'll add such tests, but let me clarify first.

When we go across the boundary of FFI, the general rule is the following.

> Memory allocators provide paired allocation and deallocation routines. When you allocate a piece of memory with one allocator, you must free it with that allocator. (ref: <https://stackoverflow.com/a/54858033>)

The problem I missed first was the mismatch here; my previous code allocated `DevDesc` on Rust's side, but it's what gets deallocated on R's side in `GEdestroyDevDesc()`. Since Rust and R uses different allocators, the memory will be corrupted.

But, unlike other R's objects, there's no API to create a `DevDesc` on R's side. So, we have to find some measure to do the allocation on R's side.I thought it's `R_calloc_gc()`, but it seems there are still some mismatches. I think this is worrying. In theory, if we use R's API both to allocate and deallocate, it should be safe. But the current failure on R-release on Windows indicates something might be wrong with this assumption (of course, there are many possibilities that it's just I make some silly mistake). So, I hope I can find the root cause before merging this.

Anyway, I'll write tests and a sample package over the next days...

Ilia-Kosenkov:
Did you check other device implementations? What about this example (`tikzDevice`)?
https://github.com/daqana/tikzDevice/blob/master/src/tikzDevice.c#L177

yutannihilation:
Thanks, I didn't checked tikzDevice, but I studied several implementations. I think my code basically follow the same logic.

- postscript device: https://github.com/wch/r-source/blob/trunk/src/library/grDevices/src/devPS.c
- svglite package: https://github.com/r-lib/svglite/blob/main/src/devSVG.cpp
- ragg package: https://github.com/r-lib/ragg/blob/main/src/init_device.h
- devout package: https://github.com/coolbutuseless/devout/blob/master/src/rdevice.cpp

Ilia-Kosenkov:
What I did not realize is that your tests succeed on 32bit Windows but fail on 64bit. Did you disable these tests on 32 bit?

yutannihilation:
No, I did nothing special on 32bit. Curious, indeed...

Ilia-Kosenkov:
This could be an issue with the struct size: 32bit version has correct size while 64bit is (likely) larger, having incorrect layout. Did you try out `#[repr(C)]`? I had no luck with it locally, but you have introduced a lot of changes since I tried, perhaps it will help?

yutannihilation:
`DevDesc` comes from libR-sys so it's of course marked as `#[repr(C)]`.

https://github.com/extendr/libR-sys/blob/355cf0de0dd7f28cfa3933a70b29deb247eb2017/bindings/bindings-windows-x86_64-R4.1.rs#L7128

Among the things this pull request introduced, what's passed to R's side is only `DeviceDriver` (more precisely, a struct that implements this trait). But, it's passed as a pointer and never treated as a struct on R's side, so it should not matter, I guess.

yutannihilation:
This is the minimal reproducible example that crashes with R-release but doesn't with R-devel. this just allocates and deallocates the `DevDesc`.

``` rs
use extendr_api::prelude::*;
use libR_sys::{GEcreateDevDesc, GEdestroyDevDesc, R_calloc_gc, DevDesc};

fn create_and_destroy() {
    single_threaded(|| unsafe {
        let p_dev_desc = R_calloc_gc(1, std::mem::size_of::<DevDesc>() as _) as *mut DevDesc;
        let device = GEcreateDevDesc(p_dev_desc);

        GEdestroyDevDesc(device);
    });
}

#[test]
fn graphic_device_test() {
    test! {
        create_and_destroy();
    }
}

fn main() {
    create_and_destroy();
}
```

yutannihilation:
Update: this time, it seems the problem is not related to R/Rust boundary. I can reproduce the crash with the following C++ code. I thought `R_calloc_gc()` and `calloc()` are the same as long as the first `calloc()` succeeds, but it seems my assumption was wrong.

``` r
cpp11::cpp_source(
  code = '
#include "cpp11/integers.hpp"
#include <R_ext/GraphicsEngine.h>

[[cpp11::register]]
void create_and_destroy(bool use_R_API) {
  pDevDesc dd;
  
  if (use_R_API) {
    dd = (DevDesc*) R_calloc_gc(1, sizeof(DevDesc));
  } else {
    dd = (DevDesc*) calloc(1, sizeof(DevDesc));
  }
  pGEDevDesc device = GEcreateDevDesc(dd);
  
  GEdestroyDevDesc(device);
}
')

# This works
create_and_destroy(FALSE)

# This crashes
create_and_destroy(TRUE)
```

yutannihilation:
It seems `R_calloc_gc()` should be considered as "non-API call" as it's not mentioned in WRE. I need to switch it to `R_Calloc()`.

https://github.com/extendr/libR-sys/issues/92

Note that this doesn't make the situation worse or better. I confirmed the C++ code above still crashes with `R_Calloc()`...

yutannihilation:
After posting the above comment, I found WRE says

> Do not assume that memory allocated by Calloc/Realloc comes from the same pool as used by malloc: in particular do not use free or strdup with it. 

so this is not what we want. Ouch...

I'd say this isn't the end of the world, but it seems there's no API that helps us. I'll revert it back to `libc::calloc()` and fight with that failure on R-devel...

yutannihilation:
Now all checks pass. The mistake was that I used `mingw64` for R-devel, which is built with UCRT...

https://github.com/extendr/extendr/blob/96bcf7d888728cc8ddf3d83a679285c54dcc9deb/.github/workflows/test.yml#L121

Ilia-Kosenkov:
I am not sure I understand, I thought the issue was that R-release fails on Windows while R-devel succeeds. Can you give a short summary of what was the issue and how you solved it? I am sure it will be useful in the future. Perhaps as a separate informative issue?

yutannihilation:
I'll write, but the short version of the answer is the issue is the allocator mismatch. I tried to avoid it by using R's API, but it in turn caused [the unknown issue on R-release, which was irrelevant to Rust](https://github.com/extendr/extendr/pull/360#issuecomment-1021784499). Besides, I noticed `R_calloc_gc()` was not an "API", so I chose `libc::calloc()` again. Later I found the failure on R-devel was just the misconfiguration of the toolchain on CI.

Ilia-Kosenkov:
Oh, I see, so `libc::calloc()` solves the problem? In the link to `tikzDevice` they actually use `calloc()` to allocate memory so I was super-frustrated that `libc::calloc()` did not work in our case. 
Anyway, impressive job, but I need some time to go through the code in order to write a review :)

yutannihilation:
> so `libc::calloc()` solves the problem?

Yes, but, to be precise, it solves the problem only when the same "libc" (or C runtime) is used. Otherwise there still can be allocator mismatch. I don't understand the details here, but this depends on the PATHs on compilation (probably especially on which GCC is used), and we have to provide `${RTOOLS40_HOME}/ucrt64/bin`, the toolset used for compiling R-devel.

This misconfiguration can happen on our users as well, so I need to write some more documents on this, especially on the README of libR-sys. When compiled within an R session, the PATH is correctly set by R, so we don't need to worry too much, though...

Thanks for reviewing, I also need some time to write tests and documents, so it's not in a hurry!

Ilia-Kosenkov:
Some time ago on our Discord channel, I proposed to create Github Actions for setting up all things extendr-related. Perhaps having a [composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) can be beneficial -- that way we will have all the configurations in one place and take care of ucrt path issue, at least on CI.

yutannihilation:
Yeah, it should be useful.

andy-thomason:
Well done with all of this @yutannihilation.

It should be a lot easier to create graphics devices this way.

I can help with a 3D device if that is interesting. Ideally using WebGPU.

yutannihilation:
Thanks! WebGPU is actually my main motivation to develop this feature :)

extendr-api/src/graphics/mod.rs
```diff
@@ -547,7 +654,7 @@ impl Device {
     pub fn rectangle(&self, from: (f64, f64), to: (f64, f64), gc: &Context) {
         let from = gc.t(from);
         let to = gc.t(to);
-        unsafe { GELine(from.0, from.1, to.0, to.1, gc.context(), self.inner()) }
+        unsafe { GERect(from.0, from.1, to.0, to.1, gc.context(), self.inner()) }
```
yutannihilation, line 657: I just remembered I can use `Device` methods for testing. As a bonus, I found a small mistake here :)

Currently, I'm exposing the `DeviceDriver`'s methods as raw as the underlying R API (e.g. `fn rect(&mut self, x0: f64, y0: f64, x1: f64, y1: f64, gc: R_GE_gcontext, dd: DevDesc) {}` for this), but probably I should match the signatures.

yutannihilation:
The failures are caused by this, unrelated to this pull request.

https://github.com/extendr/rextendr/issues/172

Cargo.toml
```diff
@@ -7,4 +7,7 @@ members = [
 ]
 
 [patch.crates-io]
-#libR-sys = { git = "https://github.com/extendr/libR-sys" }
+# When uncomment this, do not forget to uncomment the same line in
```
Ilia-Kosenkov, line 10: 'Uncommenting'

extendr-api/src/graphics/device_descriptor.rs
```diff
@@ -0,0 +1,241 @@
+use super::{color::Color, FontFace, LineType};
+
+// R internals says:
```
Ilia-Kosenkov, line 3: Perhaps 'states' would be better? Or simply 'From R internals:'
If it is written in the source code, perhaps you can insert a link to R mirror on github?

extendr-api/src/graphics/device_descriptor.rs
```diff
@@ -0,0 +1,241 @@
...
+const HEIGH_INCH: f64 = 7.0;
+
+#[allow(dead_code)]
+pub(crate) enum CanHAdjOption {
+    NotSupported = 0,
+    FixedAdjustment = 1,
+    VariableAdjustment = 2,
+}
+
+pub enum GraphicDeviceCapabilityTransparency {
```
Ilia-Kosenkov, line 36: What about using abbreviations, e.g. `GDCTransparency`?

extendr-api/src/graphics/device_driver.rs
```diff
@@ -0,0 +1,827 @@
...
+            let data = ((*dd).deviceSpecific as *mut T).as_mut().unwrap();
+            // TODO
+            // data.releaseMask(ref_, *dd);
+        }
+
+        //
+        // ************* defining the wrapper functions ends here ****************
+        //
+
+        // `Box::new()` allocates memory on the heap and place `self` into it.
```
Ilia-Kosenkov, line 589: 'places'

extendr-api/src/graphics/device_driver.rs
```diff
@@ -0,0 +1,827 @@
...
+        // Then, an unsafe function `Box::into_raw()` converts it to a raw
+        // pointer. By doing this, Rust won't drop the object so that it will
+        // live after passing it to the R's side. In other words, extendr should
+        // take care of dropping in the `close()` wrapper.
```
Ilia-Kosenkov, line 593: I suggest 'By doing *so*' and 'it will *survive* after', and perhaps 'after *being passed* to the R's side'.

extendr-api/src/graphics/device_driver.rs
```diff
@@ -0,0 +1,827 @@
...
+                GraphicDeviceCapabilityCapture::Yes as _
+            } else {
+                GraphicDeviceCapabilityCapture::No as _
+            };
+
+            (*p_dev_desc).haveLocator = GraphicDeviceCapabilityLocator::Unset as _;
+
+            // NOTE: Unlike the features that will be added in  Graphics API
+            // version 15 (i.e. R 4.2), the features in API v13 & v14 (i.e. R
+            // 4.1) are not optional. We need to provice the placeholder
```
Ilia-Kosenkov, line 778: 'provide'

extendr-api/src/graphics/mod.rs
```diff
@@ -92,6 +177,52 @@ pub enum FontFace {
...
+
+impl LineJoin {
+    fn to_u32(&self) -> u32 {
+        match self {
+            Self::RoundJoin => 1,
+            Self::MitreJoin => 2,
+            Self::BevelJoin => 3,
+        }
+    }
+}
```
Ilia-Kosenkov, line 198: What about changing the names of constants to reduce repetition? I believe most of the time it will be accessed as `LineJoin::RoundJoin`, which might be more readable as `LineJoin::Round`.

extendr-api/src/graphics/mod.rs
```diff
@@ -92,6 +177,52 @@ pub enum FontFace {
...
+    fn to_i32(&self) -> i32 {
+        match self {
+            Self::PlainFont => 1,
+            Self::BoldFont => 2,
+            Self::ItalicFont => 3,
+            Self::BoldItalicFont => 4,
+            Self::SymbolFont => 5,
+        }
+    }
+}
```
Ilia-Kosenkov, line 224: Same applies here. `FontFace::Italic` sounds natural.

extendr-api/src/graphics/device_descriptor.rs
yutannihilation: It is.

https://github.com/wch/r-source/blob/trunk/doc/manual/R-ints.texi

But, I think I should link to the HTML version (https://cran.r-project.org/doc/manuals/r-devel/R-ints.html)

extendr-api/src/graphics/device_descriptor.rs
yutannihilation: I agree this is a bit too long, but I think `GDC` a bit too short. I need to invent a nicer name...

extendr-api/src/graphics/mod.rs
yutannihilation: Agreed.

yutannihilation:
Thanks, addressed comments.

extendr-api/src/graphics/mod.rs
```diff
@@ -544,10 +651,10 @@ impl Device {
     /// Draw a stroked/filled axis-aligned rectangle.
     /// gc.color() is the stroke color.
     /// gc.fill() is the fill color.
-    pub fn rectangle(&self, from: (f64, f64), to: (f64, f64), gc: &Context) {
+    pub fn rect(&self, from: (f64, f64), to: (f64, f64), gc: &Context) {
```
yutannihilation, line 654: I want to choose `rect()` here. I believe "rect" is more popular than "rectangle" in 2D drawing libraries.

* https://processing.org/reference/rect_.html
* https://developer.mozilla.org/en-US/docs/Web/SVG/Element/rect
* https://docs.rs/lyon/latest/lyon/math/fn.rect.html

extendr-api/tests/graphics_tests.rs
```diff
@@ -92,3 +96,236 @@ fn graphics_test() {
...
+
+        device.clip((1.1, 2.2), (3.3, 4.4), &gc);
+        device.circle((1.1, 2.2), 3.3, &gc);
+        device.line((1.1, 2.2), (3.3, 4.4), &gc);
+        device.rect((1.1, 2.2), (3.3, 4.4), &gc);
+
+        // // TODO: uncomment the following lines when https://github.com/extendr/extendr/issues/370 is solved.
+        // device.polyline([(0.0, 0.0), (0.0, 2.0)], &gc);
+        // device.polygon([(0.0, 0.0), (1.0, 2.0), (2.0, 0.0)], &gc);
+        // device.path([[(0.0, 0.0), (1.0, 0.0), (1.0, 1.0), (0.0, 1.0)], [(0.3, 0.0), (0.3, 0.3), (0.7, 0.3), (0.3, 0.7)]], true, &gc);
```
yutannihilation, line 302: For some reason, these ones don't work. I'll uncomment here when https://github.com/extendr/extendr/issues/370 is resolved.

yutannihilation:
Now my toy SVG device works to the extent I can ggplot! 🎉 

https://github.com/yutannihilation/extendrSVGdevice

image

yutannihilation:
I couldn't reproduce this error on my local. Even on the CI, it seems this doesn't always happen. Considering the graphic device has some side effects, it's understandable it's unstable when the tests concurrently. I added `--test-threads=1` for now. Hope this fixes the problem...

<https://github.com/extendr/extendr/runs/4990967826?check_suite_focus=true#step:8:1607>:
```
  running 2 tests
  TextMetric { ascent: 0.269, descent: 0.007500000000000001, width: 0.278 }
  TextMetric { ascent: 0.269, descent: 0.11, width: 0.278 }
  TextMetric { ascent: 0.367, descent: 0.007500000000000001, width: 0.278 }
  TextMetric { ascent: 0.269, descent: 0.007500000000000001, width: 0.278 }
  TextMetric { ascent: 0.269, descent: 0.11, width: 0.278 }
  error: test failed, to rerun pass '--test graphics_tests'
  
  Caused by:
    process didn't exit successfully: `D:\a\extendr\extendr\target\i686-pc-windows-gnu\debug\deps\graphics_tests-7e9b9a5cfacd9a36.exe --nocapture` (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
  Error: 'Test extendr-api \w tests-minimal for i686-pc-windows-gnu target': cargo failed with code -1073741819 (args: +nightly-msvc test --manifest-path extendr-api/Cargo.toml --features tests-minimal --target=i686-pc-windows-gnu -Zdoctest-xcompile -- --nocapture)
ci-cargo: D:\a\_temp\0b077ab7-841c-483f-bb13-5494e895555e.ps1:18
Line |
  18 |    ci-cargo +$toolchain test --manifest-path extendr-api/Cargo.toml -- …
     |    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | 'Test extendr-api \w tests-minimal for i686-pc-windows-gnu target': cargo failed with code -1073741819
     | (args: +nightly-msvc test --manifest-path extendr-api/Cargo.toml --features tests-minimal
     | --target=i686-pc-windows-gnu -Zdoctest-xcompile -- --nocapture)
```

yutannihilation:
It seems passed.

While there are still some homework, I think I've done what I should implement in this pull request.

I'll address the following tasks, except for the first one, after this pull request gets merged.

- [ ] update snapshot tests to fix the CI failures: extendr/rextendr#172 (this is irrelevant to the changes by this pull request)
- [ ] update libR-sys's README to setup the toolchain on R-devel: (no issue)
- [ ] create extendr-graphics crate: https://github.com/extendr/extendr/issues/361
- [ ] consider a better interface for capabilities: https://github.com/extendr/extendr/issues/367
- [ ] implement the methods added recently: https://github.com/extendr/extendr/issues/366
- [ ] consider if there can be a better interface for `path()`, `polyline()`, and `polygon()`: (no issue)
- [ ] fix segfault when controlling the graphic device on Rust code: https://github.com/extendr/extendr/issues/370 (this seems irrelevant to the changes by this pull request)

andy-thomason:
Very good! Feel free to merge if you feel comfortable.

I think that there are many things we can do in the graphics space, but getting this is a huge milestone.

PR #363, opened 2022-01-18
Add support for saving/loading R objects.

andy-thomason:
This PR adds the ability to save and load R objects to RDS.

```
        let v = Robj::from(1).save_to_bytes(PstreamFormat::AsciiFormat, 3, None)?;
        let s = std::str::from_utf8(&v).unwrap();
        assert!(s.starts_with("A\n"));
        assert!(s.contains("\nUTF-8\n"));
```


extendr-api/src/io/outstream.rs
```diff
@@ -0,0 +1,149 @@
...
+                OutPersistHookData: hook_data,
+            };
+            let mut os = Box::new(OutStream { r_state, writer });
+            os.r_state.data = &mut os.writer as *mut W as R_pstream_data_t;
+            os
+        }
+    }
+}
+
+pub trait Serialize {
```
yutannihilation, line 73: Can we choose a different name to distinguish from `serde::Serialize`? e.g. `SerializeRDS`

extendr-api/src/io/outstream.rs
```diff
@@ -0,0 +1,149 @@
...
+    fn save_to_file<P: AsRef<std::path::Path>>(
+        &self,
+        path: &P,
+        format: PstreamFormat,
+        version: i32,
+        hook: Option<Hook>,
+    ) -> Result<()>;
+    fn save_to_bytes(&self, format: PstreamFormat, version: i32, hook: Option<Hook>) -> Result<Vec<u8>>;
```
yutannihilation, line 81: How about something like `to_bytes()`? "save to" sounds a bit strange to me here (it might make sense if we pass a `&mut Vec<u8>` and "save" the data into it).

extendr-api/src/io/outstream.rs
```diff
@@ -0,0 +1,149 @@
...
+            };
+            let mut os = Box::new(OutStream { r_state, writer });
+            os.r_state.data = &mut os.writer as *mut W as R_pstream_data_t;
+            os
+        }
+    }
+}
+
+pub trait Serialize {
+    fn save_to_file<P: AsRef<std::path::Path>>(
```
yutannihilation, line 74: How about simply `save()`?

extendr-api/src/wrapper/externalptr.rs
```diff
@@ -24,6 +24,34 @@ pub struct ExternalPtr<T: Debug + 'static> {
...
+
+/// find_var() etc.
+impl<T: Debug + 'static> Rinternals for ExternalPtr<T> {}
+
+/// as_typed_slice_raw() etc.
+impl<T: Debug + 'static> Slices for ExternalPtr<T> {}
+
+/// dollar() etc.
+impl<T: Debug + 'static> Operators for ExternalPtr<T> {}
+
```
yutannihilation, line 54: I guess these changes are accidentally included? They look irrelevant to saving/loading R objects.

extendr-api/src/io/outstream.rs
andy-thomason: Oky doky.

extendr-api/src/io/outstream.rs
andy-thomason: And `to_file` also?

extendr-api/src/io/outstream.rs
andy-thomason: Sounds good.

extendr-api/src/wrapper/externalptr.rs
andy-thomason: I was working on the save hook, which is a challenge to implement at the moment.

This would have required an `ExternalPtr` state.

I could open another PR.

andy-thomason:
I think this is good for now.

I would like to revisit the hook interface in the future, but it is good to have this.

With this change, we can generate data in rust code and save it as RDS files or vice versa.

Combined with the Serde code, we can read/write JSON/YAML and a few other formats.

Ilia-Kosenkov (approved):
I am not really familiar with how this type of saving/loading works in R, but LGTM. As a side note, I think it could be nice to have some sort of an integration test of both save/load and serde in `{extendrtests}` to both test behavior and showcase capabilities of extendr.

andy-thomason:
Thanks @Ilia-Kosenkov 

I'm going to try to spend some time on documentation soon.

PR #371, opened 2022-02-04
Create extendr-graphics

### LINKED ISSUE #361: Move graphics-related things to a separate crate (extendr-graphics?)
c.f. https://github.com/extendr/extendr/pull/360#issue-1104523978

yutannihilation:
Close #361 

This pull request moves graphics-related code into a separate crate "extendr-graphics". While there are a lot room for improvement, I'd like to hold the changes minimal.


tests/extendrtests/src/rust/Cargo.toml
```diff
@@ -14,19 +14,24 @@ crate-type = ["staticlib"]
...
 extendr-api = "*"
+# Remove "path" when we publish "extendr-graphics" to crates.io
+extendr-graphics = { path = "../../../../../../../extendr/extendr-graphics" }
 
 [patch.crates-io]
 ## This is configured to work with RStudio features.
 ## Replace by absolute path to simplify testing.
 ## CI overrides this path.
-extendr-api = { path = "../../../../../extendr/extendr-api" }
+extendr-api = { path = "../../../../../../../extendr/extendr-api" }
```
yutannihilation, line 24: Honestly, I have no idea why the current path has worked on the CI. I've never succeeded with this on my local...

yutannihilation:
One thing I'm not certain if this is the right direction is that currently the error type is defined in extendr-api. While it's possible that extendr-graphics has its own `Error`, I think it anyway needs to be integrated in extendr-api's error in order to let `#[extendr]` handle the error as well (I might be wrong. I don't see the whole picture here).

https://github.com/extendr/extendr/blob/7a2eed990101c36304b0e78d70bd455061c66e59/extendr-api/src/error.rs#L79

Should I not separate the crate and use feature gate instead...?

yutannihilation:
After some thinking, I'm withdrawing this pull request because

* To separate graphics-related things into another crate, probably we also need to separate extendr-api more into several fine-grained crates (e.g. extendr-errors?), but it doesn't sound worth to me, at least at the moment (it might eventually make sense).
* One of the motivation of this is to reduce the CI time, but it seems this wouldn't improve it much (e.g. last check on this took 18m 49s, while the current master took 19m 48s).

I'll create another pull request to put the graphics-related code behind a feature gate.

PR #374, opened 2022-02-05
Use r-lib/actions@v2 for installing and caching the dependencies

yutannihilation:
* Use [`r-lib/actions/setup-r-dependencies@v2`]() for installing and caching the dependencies
* Remove RSPM as it's now automatically used internally in [`r-lib/actions/setup-r@v2`](https://github.com/r-lib/actions/tree/v2-branch/setup-r) (if we want to stop this, we can set `use-public-rspm` to `false`)
* Remove considerations on the case of R < 4.0 because it's been a while since R 4.0 was released.
* Set a version on `actions/upload-artifact`
* Use `actions/checkout` to checkout extendr/rextendr

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -80,6 +80,7 @@ jobs:
         uses: r-lib/actions/setup-r@v2
         with:
           r-version: ${{ matrix.config.r }}
+          use-public-rspm: true
       
       - name: Set up Pandoc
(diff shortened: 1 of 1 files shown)


Ilia-Kosenkov (approved):
LGTM, thanks!

yutannihilation:
Thanks!

PR #375, opened 2022-02-09
Improved TryFrom implementations

### LINKED ISSUE #369: Task: Compile a list of TryFrom conversions and test against it.
For example.

On Input:

Let B bet the set of basic types:

```
i8/u8-i128/u128 isize/usize
f32/f64
Rcomplex (from libR-sys).
```

Let S be the set of scalar types:

`Rint, Rfloat, Rint, Rcplx.`

Let W be the set of wrapper types including Robj.

`Robj, List, Integers etc.`

Let A be the set of slices:

`&[B], &[S], &[Rstr], &[Robj]`

Let O be the set of Options:

`Option<B>, Option<S>, Option<W>, Option<A>` and all other types.

We should have special behaviour for:

`&S, &W, &Rstr, &str - these can use TryFrom<&Robj>` and avoid cloning the parameter.

Option should always be None for NULL.

For NA, the set S has explicit methods.

In the code generator:

Primitive types B, references and Option<&X> should all use `TryFrom<&Robj>`.

We should encourage users to use references as these can be faster. Internally `&Robj` and `&SEXP`
are isomorphic.

We need code examples in rextendr for all of these.

The crate documentation needs to be updated.

andy-thomason:
This PR improves the TryFrom conversions used in input parameters.

* All `TryFrom` conversions now support `&Robj`
* All `TryFrom` conversions now support Option<> with NA and NULL filtering. 

Types supported:
```
    u8 u16 u32 u64 usize
    i8 i16 i32 i64 isize
    bool
    Rint Rfloat Rbool Rcplx
    f32 f64
    Vec::<Rint> Vec::<Rfloat> Vec::<Rbool> Vec::<Rcplx> Vec::<u8> Vec::<i32> Vec::<f64>
    &[Rint] &[Rfloat] &[Rbool] &[Rcplx] &[u8] &[i32] &[f64]
    &str String
```

closes #369

### DIFF
--- a/extendr-api/src/robj/tests.rs
+++ b/extendr-api/src/robj/tests.rs
@@ -1,3 +1,4 @@
+use crate::scalar::*;
 use crate::*;
 
 #[test]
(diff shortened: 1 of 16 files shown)


yutannihilation:
It seems the graphics tests are still unstable. Sorry for the trouble.

andy-thomason:
Hi @yutannihilation , no problems! I'm sure this is going to be resolved soon. I'm somewhat snowed under with work, but if you need help tracking this down, please ask.

yutannihilation:
Thanks! I think I can handle it. Probably the fix is simply adding `--test-threads=1`, as graphics devices inevitably have side effects.

CGMossa:
Hmm.. any particular reason why `i128` and `u128` was skipped for this one?

PR #376, opened 2022-02-12
Use Rtools42, stop cross-compiling

yutannihilation:
It seems now r-lib/actions provides Rtools42. Let's try this and see if we can live without cross-compiling...

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -119,7 +119,7 @@ jobs:
       - name: Configure Windows (R-devel)
         if: startsWith(runner.os, 'Windows') && matrix.config.r == 'devel'
         run: |
-          $rtools42_home = "C:/rtools42"
+          $rtools42_home = "C:\rtools42"
 
           # c.f. https://github.com/wch/r-source/blob/f1501504df8df1668a57d3a1b6f80167f24441d3/src/library/profile/Rprofile.windows#L70-L71
           echo "${rtools42_home}\x86_64-w64-mingw32.static.posix\bin"      | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append ;
@@ -144,6 +144,8 @@ jobs:
               "${rtools42_home}\x86_64-w64-mingw32.static.posix\lib\gcc\x86_64-w64-mingw32.static.posix\10.3.0\libgcc.a" `
               "${rtools42_home}\x86_64-w64-mingw32.static.posix\lib\gcc\x86_64-w64-mingw32.static.posix\10.3.0\libgcc_s.a"
           }
+
+          echo "BUILD_TARGETS=x86_64-pc-windows-gnu" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append ;
         env: 
           RUST_TOOLCHAIN: ${{ matrix.config.rust-version }}
 
@@ -405,7 +407,7 @@ jobs:
       - name: Configure Windows (R-devel)
         if: startsWith(runner.os, 'Windows') && matrix.config.r == 'devel'
         run: |
-          $rtools42_home = "C:/rtools42"
+          $rtools42_home = "C:\rtools42"
 
           echo "::group::Setting up x86_64"
           echo "C:\msys64\mingw64\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -144,7 +144,9 @@ jobs:
               "${rtools42_home}\x86_64-w64-mingw32.static.posix\lib\gcc\x86_64-w64-mingw32.static.posix\10.3.0\libgcc.a" `
               "${rtools42_home}\x86_64-w64-mingw32.static.posix\lib\gcc\x86_64-w64-mingw32.static.posix\10.3.0\libgcc_s.a"
           }
-
+          
+          # Add target
+          rustup target add x86_64-pc-windows-gnu
           echo "BUILD_TARGETS=x86_64-pc-windows-gnu" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append ;
         env: 
           RUST_TOOLCHAIN: ${{ matrix.config.rust-version }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -123,24 +123,24 @@ jobs:
           echo "${rtools42_home}\usr\bin"                                  | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append ;
           echo "$(Rscript.exe -e 'cat(normalizePath(R.home()))')\bin\x64"  | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append ;
 
-          # The default linker of rustc is "x86_64-w64-mingw32-gcc", however, there's no such name of executable in Rtools42
+          # rustc adds -lgcc_eh and -lgcc_s flag to the compiler, but there's no libgcc_eh or libgcc_a in Rtools42.
+          # So, we need to create an empty lib to please the compiler.
+          # For more details, please refer to https://github.com/r-windows/rtools-packages/blob/2407b23f1e0925bbb20a4162c963600105236318/mingw-w64-gcc/PKGBUILD#L313-L316
+          New-Item -Path libgcc_mock -Type Directory
+          New-Item -Path libgcc_mock\gcc.c -Type File
+          x86_64-w64-mingw32.static.posix-gcc.exe -c libgcc_mock\gcc.c -o libgcc_mock\gcc.o
+          x86_64-w64-mingw32.static.posix-ar.exe -r libgcc_mock\libgcc_eh.a libgcc_mock\gcc.o
+          x86_64-w64-mingw32.static.posix-ar.exe -r libgcc_mock\libgcc_s.a libgcc_mock\gcc.o
+
           New-Item -Path .cargo -ItemType Directory -Force
-          @'
+          @"
+          # The default linker of rustc is "x86_64-w64-mingw32-gcc", however, there's no such name of executable in Rtools42
           [target.x86_64-pc-windows-gnu]
           linker = "x86_64-w64-mingw32.static.posix-gcc.exe"
-          '@ | Out-File -FilePath .cargo/config.toml -Encoding utf8 -Append ;
 
-          # rustc adds -lgcc_eh and -lgcc_s flag to the compiler, but there's no libgcc_eh or libgcc_a in Rtools42.
-          # For more details, please refer to https://github.com/r-windows/rtools-packages/blob/2407b23f1e0925bbb20a4162c963600105236318/mingw-w64-gcc/PKGBUILD#L313-L316
-          cp `
-            "${rtools42_home}\x86_64-w64-mingw32.static.posix\lib\gcc\x86_64-w64-mingw32.static.posix\10.3.0\libgcc.a" `
-            "${rtools42_home}\x86_64-w64-mingw32.static.posix\lib\gcc\x86_64-w64-mingw32.static.posix\10.3.0\libgcc_eh.a"
-          
-          if ('${{ matrix.config.rust-version }}' -eq 'stable-gnu') {
-            cp `
-              "${rtools42_home}\x86_64-w64-mingw32.static.posix\lib\gcc\x86_64-w64-mingw32.static.posix\10.3.0\libgcc.a" `
-              "${rtools42_home}\x86_64-w64-mingw32.static.posix\lib\gcc\x86_64-w64-mingw32.static.posix\10.3.0\libgcc_s.a"
-          }
+          [env]
+          LIBRARY_PATH = "${PWD}\libgcc_mock"
+          "@ | Out-File -FilePath .cargo/config.toml -Encoding utf8 -Append ;
           
           # Add target
           rustup target add x86_64-pc-windows-gnu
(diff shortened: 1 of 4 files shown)


yutannihilation:
The test fails reliably, so something should be wrong. But I don't see such a problem on my local. I need some time to figure out what's happening.

yutannihilation:
Note to self:

It seems the binary uses a wrong CRT. `msvcrt.dll`  is of MSVCRT. `api-ms-win-crt-heap-l1-1-0.dll` is for UCRT, which is in `C:\Program Files (x86)\Windows Kits\10\Redist\ucrt\DLLs\x64`.

``` sh
$ objdump.exe -p /path/to/extendr/target/x86_64-pc-windows-gnu/debug/deps/graphics_tests-9c4d0e3a4e55a785.exe | grep -B20 -w calloc
        DLL Name: msvcrt.dll
        vma:  Hint/Ord Member-Name Bound-To
        2554d8     78  __dllonexit
        2554e6     81  __getmainargs
        2554f6     82  __initenv
        255502     83  __iob_func
        255510     90  __lconv_init
        255520     96  __set_app_type
        255532     98  __setusermatherr
        255546    116  _acmdln
        255550    123  _amsg_exit
        25555e    141  _cexit
        255568    203  _errno
        255572    253  _fmode
        25557c    257  _fpreset
        255588    332  _initterm
        255594    440  _lock
        25559c    613  _onexit
        2555a6    821  _unlock
        2555b0   1033  abort
        2555b8   1050  calloc
```

```
$ objdump.exe -p ~/AppData/Local/R/win-library/4.2/wgpugd/libs/x64/wgpugd.dll  | grep -B5 -w calloc
 00ba108c       00ba1584 00000000 00000000 00ba4df0 00ba231c

        DLL Name: api-ms-win-crt-heap-l1-1-0.dll
        vma:  Hint/Ord Member-Name Bound-To
        ba365a     23  _set_new_mode
        ba366a     24  calloc
```

yutannihilation:
Thanks, it helped a bit. If I dump the setting,

``` sh
rustc +nightly -Z unstable-options --target=x86_64-pc-windows-gnu --print target-spec-json
```

I found this line.

```
  "linker": "x86_64-w64-mingw32-gcc",
```

This name of executable exists on Rtools40, but doesn't in Rtools42. We probably need to tweak the "linker" setting.

### Rtools40

``` sh
$ find /ucrt64 -name x86_64-w64-mingw32-gcc.exe
/ucrt64/bin/x86_64-w64-mingw32-gcc.exe
```

### Rtools42

``` sh
$ find /x86_64-w64-mingw32.static.posix/ /usr/bin/ -name x86_64-w64-mingw32-gcc.exe

```

yutannihilation:
If I configure `.cargo/config.toml` as follows,

```
[target.x86_64-pc-windows-gnu]
linker = "x86_64-w64-mingw32.static.posix-gcc.exe"
```

I get this error, which we saw a year ago... https://github.com/extendr/rextendr/issues/19#issuecomment-761801778

```
  = note: C:\rtools42\x86_64-w64-mingw32.static.posix\bin/ld.exe: cannot find -lgcc_eh
          collect2.exe: error: ld returned 1 exit status
```

Actually, the problem looks the same as at that time.

### Rtools40

``` sh
$ find /ucrt64 -name 'libgcc*'
/ucrt64/lib/gcc/x86_64-w64-mingw32/10.3.0/libgcc.a
/ucrt64/lib/gcc/x86_64-w64-mingw32/10.3.0/libgcc_eh.a
```

### Rtools42

``` sh
$ find /x86_64-w64-mingw32.static.posix/ /usr/bin/ -name 'libgcc*'
/x86_64-w64-mingw32.static.posix/lib/gcc/x86_64-w64-mingw32.static.posix/10.3.0/libgcc.a
```

Ilia-Kosenkov:
:-D can you check what happens if you do `cp libgcc.a libgcc_eh.a` ?

Ilia-Kosenkov:
https://github.com/extendr/extendr/actions/runs/1847304993/workflow#L133

It is dangerous to `cd` like this, it is possible the next step does not return to 'work dir' and won't be able to find code.
Still I do not understand why the path cannot be found.

yutannihilation:
Apparently, libgcc_s should be a shared object (`.so`), not a static library (`.a`), so there should not be such thing as `libgcc_s.a`, but it seems adding this can let the compile success. I have no idea what's happening.

yutannihilation:
extendrtests still specifies `+stable-msvc`, so this fails.

```
   -- Install failure -------------------------------------------------------------
  
  * installing *source* package 'extendrtests' ...
  ** using staged installation
  ** libs
  rm -Rf extendrtests.dll ./rust/target/x86_64-pc-windows-gnu/debug/lextendrtest.a entrypoint.o
  gcc  -I"C:/R/include" -DNDEBUG     -I"c:/rtools42/x86_64-w64-mingw32.static.posix/include"     -O2 -Wall  -std=gnu99 -mfpmath=sse -msse2 -mstackrealign  -c entrypoint.c -o entrypoint.o
  cargo +stable-msvc build --target=x86_64-pc-windows-gnu --lib --manifest-path=./rust/Cargo.toml --target-dir ./rust/target
```

Ilia-Kosenkov:
If I execute 
```pwsh
rustc +nightly-gnu -Z unstable-options --print target-spec-json
```
part of the JSON output contains this
```json
 "late-link-args": {
    "gcc": [
      "-lmsvcrt",
      "-lmingwex",
      "-lmingw32",
      "-lgcc",
      "-lmsvcrt",
      "-luser32",
      "-lkernel32"
    ],
    "ld.lld": [
      "-lmsvcrt",
      "-lmingwex",
      "-lmingw32",
      "-lgcc",
      "-lmsvcrt",
      "-luser32",
      "-lkernel32"
    ]
  },
  "late-link-args-dynamic": {
    "gcc": [
      "-lgcc_s"
    ],
    "ld.lld": [
      "-lgcc_s"
    ]
  },
  "late-link-args-static": {
    "gcc": [
      "-lgcc_eh",
      "-l:libpthread.a"
    ],
    "ld.lld": [
      "-lgcc_eh",
      "-l:libpthread.a"
    ]
  }
```
perhaps these params can be used to configure linking and avoid these mistakes? Perhaps the solution always was to create custom target from scratch instead of trying to fix Rtools and resolve linking issues?

yutannihilation:
I'm not convinced with the idea of custom target because

* Simply, I'm not yet sure how to use a custom target
* I don't think I will be able to gain the knowledge enough to go "from scratch". For example, did you notice `-lmsvcrt` specified twice, and did you imagine it has a reason? ([ref](https://github.com/rust-lang/rust/blob/21e1cd9b95854bec31a4d05992561ef47c0c8b07/compiler/rustc_target/src/spec/windows_gnu_base.rs#L24-L39)) I expect there would be a lot of weird things like this, and I'm afraid we cannot handle them and maintain the workarounds.
* As I wrote above, it seems this won't be a problem when we use it in a R package, which I guess is the main usage of extendr.

andy-thomason:
This is getting pretty close. Only the GNU windows build failing.

yutannihilation:
> This is getting pretty close. Only the GNU windows build failing.

Thanks, but as this pull request only modifies GNU windows build, it means this is still 0% complete...

yutannihilation:
All R-devel builds with bindgen failed. It seems it generates the bindings for R 4.1 (wrong), but chooses R-devel when compiling. Why?

```
     Compiling extendr-engine v0.2.0 (D:\a\extendr\extendr\extendr-engine)
  error[E0609]: no field `defineGroup` on type `_DevDesc`
     --> extendr-api\src\graphics\device_driver.rs:871:31
      |
  871 |                 (*p_dev_desc).defineGroup = None;
      |                               ^^^^^^^^^^^ unknown field
      |
      = note: available fields are: `left`, `right`, `bottom`, `top`, `clipLeft` ... and 70 others
  
  error[E0609]: no field `useGroup` on type `_DevDesc`
```

mati865:
Not following closely but `-lmsvcrt` doesn't necessarily link MSVCRT.
It links default CRT configured in mingw-w64, the actual import libraries are `libmsvcrt-os.a` for MSVCRT and `libucrt.a` for UCRT.

> I don't think I will be able to gain the knowledge enough to go "from scratch". For example, did you notice -lmsvcrt specified twice, and did you imagine it has a reason? ([ref](https://github.com/rust-lang/rust/blob/21e1cd9b95854bec31a4d05992561ef47c0c8b07/compiler/rustc_target/src/spec/windows_gnu_base.rs#L24-L39)) I expect there would be a lot of weird things like this, and I'm afraid we cannot handle them and maintain the workarounds.

It's hybrid of static and import library so some linkers (like the default `ld.bfd`) require specifying it twice,
Note however this list is not considered stable. The libraries itself can change and some linkers (`ld.bfd` once more) require them to be in correct order.

> As a side note, it seems UCRT is not supported on the windows-gnu toolchain so maybe we should stick with msvc...?: 

The entire problem comes to mixing MSVCRT and UCRT runtimes within single binary.
I think it (almost?) always manifests in linking error nowadays but I haven't tested it much.

yutannihilation:
Thanks so much for the information! I don't fully understand yet, but it should be helpful.

> I think it (almost?) always manifests in linking error nowadays but I haven't tested it much.

I think we don't see linking errors here, unfortunately. In my understanding, the problem is that we are playing with two binaries compiled separately and communicate via the API...

.github/workflows/test.yml
```diff
@@ -119,10 +119,31 @@ jobs:
...
           echo "$(Rscript.exe -e 'cat(normalizePath(R.home()))')\bin\x64"  | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append ;
+
+          # The default linker of rustc is "x86_64-w64-mingw32-gcc", however, there's no such name of executable in Rtools42
+          New-Item -Path .cargo -ItemType Directory -Force
+          @'
+          [target.x86_64-pc-windows-gnu]
+          linker = "x86_64-w64-mingw32.static.posix-gcc.exe"
+          '@ | Out-File -FilePath .cargo/config.toml -Encoding utf8 -Append ;
+
+          # rustc adds -lgcc_eh flag to the compiler, but there's no libgcc_eh in Rtools42 (this might be a bug in MSYS2 itself?, but Rtools40
```
mati865, line 136: `libgcc_eh.a` is provided by GCC and is present in toolchain provided by MSYS2: https://packages.msys2.org/package/mingw-w64-x86_64-gcc (there is file list at the bottom).

.github/workflows/test.yml
```diff
@@ -119,10 +119,31 @@ jobs:
...
+          # has a workaround for this).
+          cp `
+            "${rtools42_home}\x86_64-w64-mingw32.static.posix\lib\gcc\x86_64-w64-mingw32.static.posix\10.3.0\libgcc.a" `
+            "${rtools42_home}\x86_64-w64-mingw32.static.posix\lib\gcc\x86_64-w64-mingw32.static.posix\10.3.0\libgcc_eh.a"
```
mati865, line 140: This is wrong.
There are 3 GCC libraries:
- `libgcc.a`
- `libgcc_eh.a`
- `libgcc_s.a`

`libgcc.a` contains compiler builtins among the other things but has nothing to do with unwinding.

`libgcc_eh.a` is static unwind library, while `libgcc_s.a` is import library for shared unwind library (on 64-bit it's `libgcc_s_seh-1.dll`).

.github/workflows/test.yml
yutannihilation: Yes, but we are tied to this toolchain where no `libgcc_eh.a` or `libgcc_s.a`

mati865: So it has nothing to do with MSYS2 but the toolchain you are using doesn't provide unwinding.

.github/workflows/test.yml
yutannihilation: Does this help?

> ```
> # 'gcc' when build with --disable-shared does not produce a standalone 'libgcc_eh.a'.
> # Instead, everything is put into 'libgcc.a'.
> # To please compilers that hardcode -lgcc_eh, such as 'rustc', 'libgcc.a' is duplicated as 'libgcc_eh.a'.
> ```
>(https://github.com/r-windows/rtools-packages/blob/2407b23f1e0925bbb20a4162c963600105236318/mingw-w64-gcc/PKGBUILD#L313-L316)

We confirmed `libgcc.a` and `libgcc_eh.a` are the same binaries. We don't figure out what `libgcc_s` is yet, so thanks for the details.

.github/workflows/test.yml
yutannihilation: Ah, yes. Sorry for blaming MSYS2 wrongly.

.github/workflows/test.yml
mati865: Oh, that's really unfortunate from GCC to that.

yutannihilation:
> All R-devel builds with bindgen failed.

Now I see what's happening. All R-devel builds **without** bindgen are failing because we don't have precomputed bindings for R 4.3-devel now. The tests with `--features graphics` don't use bindgen, so it try to fall back to the precomputed bindings, but it's not available. So, we need to

1. merge https://github.com/extendr/libR-sys/pull/102 and
2. use bindgen also in graphics tests

yutannihilation:
Failing the same error as https://github.com/extendr/rextendr/pull/187#issuecomment-1079964364, which cannot be reproduced on my local. Considering the runner with stable-msvc passes, I guess `cargo` looks up a wrong toolchain? (i.e. the pre-installed MSYS2).

andy-thomason:
This is a tricky issue!

Likely the library search path is missing/incorrect and may need an extra -L.

Note that this is supposed to b release and yet this has `-L./rust/target/x86_64-pc-windows-gnu/debug` instead of release.

```
gcc -shared -s -static-libgcc -o extendrtests.dll tmp.def entrypoint.o -L./rust/target/x86_64-pc-windows-gnu/debug -lextendrtests -lws2_32 -ladvapi32 -luserenv -lbcrypt -Lc:/rtools42/x86_64-w64-mingw32.static.posix/lib/x64 -Lc:/rtools42/x86_64-w64-mingw32.static.posix/lib -LC:/R/bin/x64 -lR
  Error: Error: R CMD check found ERRORs
  C:\rtools42\x86_64-w64-mingw32.static.posix\bin/ld.exe: cannot find -lextendrtests
  collect2.exe: error: ld returned 1 exit status
```

yutannihilation:
@andy-thomason 
Thanks.

> Note that this is supposed to b release and yet this has `-L./rust/target/x86_64-pc-windows-gnu/debug` instead of release.

I don't remember but I think we mix release and debug for different purposes (debug build for speed?), and I mistook it in the earlier commits here. So I believe this wasn't the blocker here.

The only problem was this, and we overcame it!
https://github.com/extendr/rextendr/pull/187#issuecomment-1084837443

yutannihilation:
@andy-thomason @Ilia-Kosenkov 
I'm merging this. This pull request is safe in the sense that this doesn't modify any of the Rust source code. If there's anything that needs to be explained, please let me know in #396!

PR #377, opened 2022-02-15
Fixed warnings that comes from calling

CGMossa:
`cargo doc`. URLs being invalid, and
renamed variants, etc.


Ilia-Kosenkov (approved):
Thanks for noticing! I re-ran CI checks because one of the runners failed.

yutannihilation:
Thanks! I'm merging this first because this might conflict with my pull request...

Issue #391, opened 2022-03-09
Rust tests fail with invalid memory reference - R tests succeed


sstadick:
Minimal failing example can be found here: https://github.com/sstadick/minimal_example_extendr/tree/sigenv_err

Running `cargo test -- --test-threads=1` in the `src/rust` directory produces an invalid memory error.
In R, after building with `rextendr::document()`, running tests with `devtools::test()` succeeds. 

It looks like the invalid memory error pops up in the function as soon as I try to convert a rust value into `Strings` (or really any other R type). If I pass back Rust types everything works just fine.

sstadick:
I fixed the link above, in case anyone got a 404.

Here is the specific test as well: https://github.com/sstadick/minimal_example_extendr/blob/d191917ecb5c87df8220361f94a3ad902d68642c/src/rust/src/lib.rs#L43

sstadick:
The above PR fixes this issue.

PR #392, opened 2022-03-10
[bug] Fix test macro

sstadick:
The `test!` macro depends on `extendr_engine` being in scope to use. This PR re-exports `estendr_engine` in the prelude so that `test!` will work without adding a silent `dev-dependency`.

This also documents how to test code in Rust.

Fixes: https://github.com/extendr/extendr/issues/391

### DIFF
--- a/extendr-api/src/lib.rs
+++ b/extendr-api/src/lib.rs
@@ -10,6 +10,8 @@
 //! See [Robj] for much of the content of this crate.
 //! [Robj] provides a safe wrapper for the R object type.
 //!
+//! ## Examples
+//!
 //! Use attributes and macros to export to R.
 //! ```ignore
 //! use extendr_api::prelude::*;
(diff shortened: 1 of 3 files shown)


andy-thomason:
Nice one.

We should make sure that macros use `$crate` as a prefix.

I've been working slowly to tidy up the exports and avoid the `pub use`s in the crate root.

The prelude is getting a bit out of hand, but it is convenient for casual use.

andy-thomason (approved):

andy-thomason:
We'll have to fix the windows link failures at some point.

yutannihilation:
Quick note: we should revert this change, otherwise the package will be rejected by CRAN. c.f. https://github.com/extendr/extendr/pull/128#discussion_r559245149

```
✔  checking compilation flags used ...
N  checking compiled code (16.8s)
   File 'string2path/libs/x64/string2path.dll':
     Found non-API calls to R: 'R_CStackLimit', 'R_CleanTempDir',
       'R_RunExitFinalizers', 'Rf_initialize_R', 'setup_Rmainloop'
   
   See 'Writing portable packages' in the 'Writing R Extensions' manual.
```

PR #394, opened 2022-03-21
Skip extendr-macros test on Windows

### LINKED ISSUE #372: doctest fails on Windows with "LINK : fatal error LNK1181: cannot open input file 'R.lib'" when proc_macro = true
(From https://github.com/extendr/extendr/pull/347#issuecomment-1030119516)

Here's the minimal reproducible example. Note that, if I change `proc_macro` to `false`, no error happens.

### `Cargo.toml`

```toml
[package]
name = "foo"
version = "0.1.0"
edition = "2021"

[lib]
proc_macro = true

[dependencies]

[dev-dependencies]
extendr-api = { path = "../extendr-api" }
```

### `src/lib.rs`

``` rs
/// ```
/// use extendr_api::prelude::*;
/// ```
fn foo() {}
```

### command

``` sh
cargo test --manifest-path foo/Cargo.toml --target=x86_64-pc-windows-gnu --doc -- --nocapture
```

yutannihilation:
Close #372

Please refer to the issue for the details, but in case it's too long to read to you, I quote myself here:

> The reason is simply `proc_macro` doesn't allow cross-compilation.
> 
> > `--crate-type=proc-macro`, `#[crate_type = "proc-macro"]` - (...snip...) The crates are always compiled with the same target that the compiler itself was built with. For example, if you are executing the compiler from Linux with an x86_64 CPU, the target will be `x86_64-unknown-linux-gnu` even if the crate is a dependency of another crate being built for a different target.  
>     (<https://doc.rust-lang.org/reference/linkage.html>)
> 
> So, I think this is a game over. I'd suggest documenting this limitation and waiting for R 4.2 to avoid cross-compilation at all.

### DIFF
--- a/extendr-macros/src/lib.rs
+++ b/extendr-macros/src/lib.rs
@@ -180,21 +180,24 @@ pub fn Rraw(item: TokenStream) -> TokenStream {
 /// In the below example, `foo_from_list` is an instance of the `Foo` struct, that has been converted
 /// from an R list:
 /// ```
-/// use extendr_api::prelude::*;
-/// use extendr_macros::TryFromRobj;
-/// # use extendr_api::test;
-/// # test!{
+/// #[cfg(not(target_os = "windows"))]
+/// {
+///     use extendr_api::prelude::*;
+///     use extendr_macros::TryFromRobj;
+///     # use extendr_api::test;
+///     # test!{
 ///
-/// #[derive(TryFromRobj, PartialEq, Debug)]
-/// struct Foo {
-///     a: u64,
-///     b: String
+///     #[derive(TryFromRobj, PartialEq, Debug)]
+///     struct Foo {
+///         a: u64,
+///         b: String
+///     }
+///     let native_foo = Foo { a: 5, b: "bar".into() };
+///     let foo_from_list: Foo = R!("list(a = 5, b = 'bar')")?.try_into()?;
+///     assert_eq!(native_foo, foo_from_list);
+///     # }
+///     # Ok::<(), extendr_api::Error>(())
 /// }
-/// let native_foo = Foo { a: 5, b: "bar".into() };
-/// let foo_from_list: Foo = R!("list(a = 5, b = 'bar')")?.try_into()?;
-/// assert_eq!(native_foo, foo_from_list);
-/// # }
-/// # Ok::<(), extendr_api::Error>(())
 /// ```
 #[proc_macro_derive(TryFromRobj)]
 pub fn derive_try_from_robj(item: TokenStream) -> TokenStream {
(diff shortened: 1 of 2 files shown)


yutannihilation:
It turned out these doc tests cannot compile without compiling extendr-api, so it seems there's no other way than disabling them, unfortunately.

```
    --> src/lib.rs:234:1
     |
  22 | Ok::<(), extendr_api::Error>(())
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `()`, found enum `std::result::Result`
     |
     = note: expected unit type `()`
                     found enum `std::result::Result<(), extendr_api::Error>`
  help: consider using a semicolon here
     |
  22 | Ok::<(), extendr_api::Error>(());
     |                                 +
  help: try adding a return type
     |
  3  | fn main() { #[allow(non_snake_case)] fn _doctest_main_src_lib_rs_215_0() -> std::result::Result<(), extendr_api::Error> {
     |                                                                          ++++++++++++++++++++++++++++++++++++++++++++++
  
  For more information about this error, try `rustc --explain E0308`.
    --> src/lib.rs:199:1
     |
  20 | Ok::<(), extendr_api::Error>(())
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `()`, found enum `std::result::Result`
     |
     = note: expected unit type `()`
                     found enum `std::result::Result<(), extendr_api::Error>`
  help: consider using a semicolon here
     |
  20 | Ok::<(), extendr_api::Error>(());
     |                                 +
  help: try adding a return type
     |
  3  | fn main() { #[allow(non_snake_case)] fn _doctest_main_src_lib_rs_182_0() -> std::result::Result<(), extendr_api::Error> {
     |                                                                          ++++++++++++++++++++++++++++++++++++++++++++++
  
```

yutannihilation:
@multimeric @Ilia-Kosenkov @andy-thomason 
I'm merging this to make the CI checks work again, though this isn't a very nice solution. Feel free to comment.

multimeric:
Rather than ignoring the doctest outright (which we could do), would either of these solutions help here: https://stackoverflow.com/a/50314936/2148718?

yutannihilation:
Thanks, but it doesn't work. See the comment above.

multimeric:
Hmm but either of these macros should actually hide the code from the compiler right? So I would have thought it would prevent the compiler trying to compile it at all.

yutannihilation:
I was hoping so, too.

yutannihilation:
To be precise, I guess you can write more explicit error handling to prevent extendr_api:: from appearing in the return type, but I don't want to complicate this pull request in order to recover the CI quickly.

multimeric:
Sounds reasonable to me.

Ilia-Kosenkov:
I am not a fan of this approach, as I already have mentioned in the original issue. As a temporary solution this should do just fine, but I'd prefer to have a better, more permanent solution at some point later.

yutannihilation:
I see. Let's file a new issue for it.

Issue #397, opened 2022-04-02
Memory corruption (?) when too many Robjs are created


bovee:
When I create a lot (hundreds of thousands) of unique `Robj`s, I sometimes see a large slowdown and duplicate objects appear in my output. This is fairly reproducible for me on Linux although if I run the same command multiple times in the console, it sometimes starts "working".

I put together a simplified test case here: https://github.com/bovee/extendr_alloc

yutannihilation:
Thanks, I confirmed it can be reproduced on the CI. No idea at the moment... It seems the numbers are not random (at least on Linux), so this can be some clue?

https://github.com/yutannihilation/extendr_alloc/runs/5803411848?check_suite_focus=true#step:8:45
https://github.com/yutannihilation/extendr_alloc/runs/5803411865?check_suite_focus=true#step:8:45

andy-thomason:
Creating unique Robj's is indeed expensive as we need to keep track of them all
so that they can be protected.

This is done by using a hash table of addresses and indices into a R List, which maintains them.
The alternative is the Rf_protect mechanism that does a linear search on every free through every object.

If you can collect them into a List (for example by from_values). The ownership gets transferred to the List and we don't need
to track individual Robjs. In the future, we might even be able to bypass the Robj creation entirely.

The List is conceptually a slice of `[&Robj]`. And `&Robj` is equivalent to a SEXP.

Avoid making `Vec<Robj>` or equivalent if you want decent performance.

andy-thomason:
There is a known problem with `elt()` on `List`s which return `&Robj` when they really can't.

I've put a quick fix in my latest PR, but it should go in on its own.

bovee:
I simplified the example a little bit so that it's only one long column and it appears that all of the "bad" entries are coming from later in the stream. I think that what's happening is allocVector (or R's allocator itself?) is returning the same SEXP twice and the write to the later one is overwriting the first one? It also appears to be repeatable from run to run for me too.

And sorry, I should have said that the slowdown appears coupled to the appearance of the reused allocations. I added some timing to the example and initially it takes ~50 to 300 ms (on my laptop) to allocate 100k objects (no errors), but by the millionth object or so it's taking 500 ms (~1 error per 100k) and by 2 million a second (2-3 errors per 100k). If I rerun the script immediately in the same environment, it only takes 20 to 200 ms per 100k and there are no errors at all.

andy-thomason:
[...]

PR #398, opened 2022-04-11
Extra Rf_protect and faster growth.

andy-thomason:
Closes issue #397 

The memory protection uses a hashmap to ensure that there is at least one instance of each live SEXP
in a list visible to R.

This issue was because during the growth of the array, the current object was not protected.

I have also increased the initial size of the list and the growth is now exponential. This improves
the performance of the allocation at some risk to memory  overflow.

### DIFF
--- a/extendr-api/src/ownership.rs
+++ b/extendr-api/src/ownership.rs
@@ -13,8 +13,8 @@ use std::collections::hash_map::{Entry, HashMap};
 use std::sync::Mutex;
 
 use libR_sys::{
-    R_NilValue, R_PreserveObject, R_ReleaseObject, R_xlen_t, Rf_allocVector, LENGTH,
-    SET_VECTOR_ELT, SEXP, VECSXP, VECTOR_ELT,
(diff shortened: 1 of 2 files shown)


extendr-api/src/ownership.rs
```diff
@@ -154,7 +158,7 @@ impl Ownership {
     // Garbage collect the tracking structures.
     unsafe fn garbage_collect(&mut self) {
         // println!("garbage_collect {} {}", self.cur_index, self.max_index);
-        let new_size = self.cur_index + EXTRA_PRESERVATION_SIZE;
+        let new_size = self.cur_index * 2 + EXTRA_PRESERVATION_SIZE;
```
Ilia-Kosenkov, line 161: Why was this chaned? It is always difficult to chose best initial size and growth pattern for growable collections. I'd be very careful with such changes unless we benchmark memory usage in a typical extend scenario. 
Using default values, the first growth will happen from 10k to 30k (if I am not mistaken), the second will jump from 30k to 70k. I am not sure it is justified.

PR #401, opened 2022-05-03
Separate the ALTREP constructor for vectors.

andy-thomason:
Before this change, vector constructors require many bounds on the iterator.

This is still true of ALTREP constructors, but the regular `from_values` is now
much simpler.

Explicit allocated vector.

```
        let vec = Doubles::from_values((0..3).map(|i| 2.0 - i as f64));
        assert_eq!(vec.is_altrep(), false);
        assert_eq!(r!(vec.clone()), r!([2.0, 1.0, 0.0]));
        assert_eq!(vec.elt(1), 1.0);
        let mut dest = [0.0.into(); 2];
        vec.get_region(1, &mut dest);
        assert_eq!(dest, [1.0, 0.0]);
```

Explicit ALTREP vector.

```
        let vec = Doubles::from_values_altrep((0..1000000000).map(|x| x as f64));
        assert_eq!(vec.is_altrep(), true);
        assert_eq!(vec.elt(12345678), 12345678.0);
        let mut dest = [0.0.into(); 2];
        vec.get_region(12345678, &mut dest);
        assert_eq!(dest, [12345678.0, 12345679.0]);
```

### DIFF
--- a/extendr-api/src/prelude.rs
+++ b/extendr-api/src/prelude.rs
@@ -59,7 +59,7 @@ pub use super::thread_safety::{
 
 pub use super::wrapper::{
     Complexes, Dataframe, Doubles, EnvIter, Environment, Expressions, ExternalPtr, FromList,
-    Function, Integers, IntoDataFrame, Language, List, ListIter, Logicals, Nullable, Pairlist,
+    Function, Integers, IntoDataFrameRow, Language, List, ListIter, Logicals, Nullable, Pairlist,
     Primitive, Promise, Raw, Rstr, Strings, Symbol,
 };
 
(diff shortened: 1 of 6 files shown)


andy-thomason:
If no one objects, I'm just going to merge this.

Ilia-Kosenkov:
If you want a thouroug review, I need some extra time. Otherwise, feel free to merge though I am concerned about failed CIs.

andy-thomason:
I'll look at the tests. Looks like the very large vector is allocated in memory, which is not good!

multimeric (changes requested):
Should really add a test for the base `X::from_values()` method.

andy-thomason:
Down to an ephemeral failure in test_altreal.

In theory everything is protected by single_threaded, if this persists. I'll take a deeper look.

andy-thomason:
The original `from_values` tests should be unchanged, but I've renamed some of them to `from_values_altrep`.

multimeric (approved):
Oh yes, I see the tests. They just didn't come up in the diff.

Ilia-Kosenkov (approved):

andy-thomason:
Oops, mea culpa.

> I think you managed to put your data frame derive commits into this branch before you merged, which isn't ideal.

PR #389, opened 2022-03-06
ADD: wrapper for locator

Stalkcomrade:

PR #408, opened 2022-05-25
Tweak on doc of `eventHelper()`

yutannihilation:
A small followup of #389


yutannihilation:
No, I just wanted to prevent these words from appearing in the doc as this doesn't look very helpful to users. Thanks for approving.

image

multimeric:
Ah right you changed it from `///` to `//`. Still looks fine!

multimeric:
Actually while you're thinking about that function, do you think you could rename it to `snake_case` ie `event_helper`? It should have been caught in code review for #389.

yutannihilation:
Ah, it seems so. Sorry, that was my fault. Probably I can fix it next week or so.

PR #417, opened 2022-06-05
Switch CI to cache v3 

### LINKED ISSUE #414: Update version of `actions/cache`
It seems we depend on [`actions/cache`](https://github.com/actions/cache) indirectly through [`setup-r-dependencies`](https://github.com/r-lib/actions/tree/v2/setup-r-dependencies) and we are currently using `v2` https://github.com/extendr/extendr/blob/7e41a09686066b11878e415eea8cb968c91399f2/.github/workflows/test.yml#L205-L216

I encountered an issue with this cache in my forked repo (e.g., [this](https://github.com/Ilia-Kosenkov/extendr/runs/6704483127?check_suite_focus=true#step:1:41)), so I found this issue https://github.com/actions/cache/issues/811.

From the discussion there I found out the latest version of `actions/cache` is [`3.0.3`](https://github.com/actions/cache/releases/tag/v3.0.3) (released quite recently), so perhaps it is time to switch to the newer version for us?

Ilia-Kosenkov:
Closes #414. See improvements in [`actions/cache@v3`](https://github.com/actions/cache#v3).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -206,14 +206,14 @@ jobs:
       - name: Install dependencies for extendrtests and rcmdcheck
         uses: r-lib/actions/setup-r-dependencies@v2
         with:
-          cache-version: 2
+          cache-version: 3
           working-directory: tests/extendrtests
           extra-packages: any::rcmdcheck
 
       - name: Install dependencies for rextendr
         uses: r-lib/actions/setup-r-dependencies@v2
         with:
-          cache-version: 2
+          cache-version: 3
           working-directory: tests/rextendr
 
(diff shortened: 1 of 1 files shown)


yutannihilation:
`cache-version` is not what you expect. It's just a kind of "revision," which we can bump when we want to clear the cache as the document says:

> cache-version - default 1. If you need to invalidate the existing cache pass any other number and a new cache will be used. Ignored if cache: false.
(ref: https://github.com/r-lib/actions/tree/v2/setup-r-dependencies#usage)

Sorry for commenting this late, I saw your issue but didn't read it enough...

Ilia-Kosenkov:
Wow, I misread the docs. I guess you are right and I did not pay enough attention. I reject this or, thanks for catching the problem

Issue #415, opened 2022-06-02
Docs build error: `` error[E0432]: unresolved imports `libR_sys::DevDesc`, `libR_sys::R_GE_gcontext` ``


multimeric:
See: https://docs.rs/crate/extendr-api/0.3.1/builds/567780

yutannihilation:
The build should be reproducible on local with Docker: https://github.com/rust-lang/docs.rs/blob/master/README.md#build-subcommand

Other information about their build environment can be found on the docs: https://docs.rs/about/builds#failures-and-rebuilds

yutannihilation:
I don't try to reproduce this yet, but I think I found the reason. docs.rs runs a Ubuntu 20.04,

https://github.com/rust-lang/crates-build-env/blob/8adee15db57330f64c2c49779210bbf0e646a1ca/linux/Dockerfile#L1

and the version of R is 3.6.

https://packages.ubuntu.com/focal/r-base-core

Why is this a problem? Well, we do have the precompiled bindings file for R 3.6 on libR-sys, but, since this was generated before we added support for graphics-related ones, this is unusable for `graphics` feature on extendr-api.

https://github.com/extendr/libR-sys/blob/master/bindings/bindings-linux-x86_64-R3.6.rs

So, I think we have several options:

1. Do nothing: as the problem is the build image is a bit too old, we can just wait until they switch to Ubuntu 22.04. But it's unclear when the day will come c.f. https://github.com/rust-lang/crates-build-env/issues/84
2. Remove `graphics` feature
3. Use `libR-sys/bindgen` feature
4. Remove or regenerate bindings for R 3.6 on libR-sys

multimeric:
And we can't submit compiled decks ourselves, from CI, can we?

yutannihilation:
I believe we can't. What do you mean?

yutannihilation:
Confirmed. The error is reproducible with R 3.6 while the same command passes with the latest version of R.

https://github.com/yutannihilation/extendr-test-R-3.6/runs/7376904757?check_suite_focus=true

multimeric:
> I believe we can't. What do you mean?

Oh, I just meant that maybe there's a way to submit pre-compiled docs without having the docs.rs build system compile them for us. But I can't find any such mechanism. I think your PR is a good solution.

yutannihilation:
Ah, I see. If it were possible, we should build it by ourselves...

Ilia-Kosenkov:
[...]

PR #426, opened 2022-07-17
Use bindgen on docs.rs

yutannihilation:
Currently, the build on docs.rs fails with the following error (ref: [build log](https://docs.rs/crate/extendr-api/0.3.1/builds/567780)). This is because the precomputed bindings for the version of R that docs.rs uses, R 3.6, doesn't contain the graphics-related objects because the file was generated long ago.

We have several options to avoid this error (see https://github.com/extendr/extendr/issues/415#issuecomment-1186461753), but I believe this is the best; we cannot expect what version of R docs.rs will use, but using bindgen should make the build success with arbitrary version.

```
[INFO] [stderr]  Documenting extendr-api v0.3.1 (/opt/rustwide/workdir)
[INFO] [stderr] error[E0432]: unresolved imports `libR_sys::DevDesc`, `libR_sys::R_GE_gcontext`
[INFO] [stderr]   --> src/graphics/mod.rs:83:20
[INFO] [stderr]    |
[INFO] [stderr] 83 | pub use libR_sys::{DevDesc, R_GE_gcontext};
[INFO] [stderr]    |                    ^^^^^^^  ^^^^^^^^^^^^^ no `R_GE_gcontext` in the root
[INFO] [stderr]    |                    |
[INFO] [stderr]    |                    no `DevDesc` in the root
[INFO] [stderr] 
[INFO] [stderr] error[E0412]: cannot find type `DevDesc` in this scope
[INFO] [stderr]   --> src/graphics/device_driver.rs:82:32
[INFO] [stderr]    |
[INFO] [stderr] 82 |     fn activate(&mut self, dd: DevDesc) {}
[INFO] [stderr]    |                                ^^^^^^^ not found in this scope
[INFO] [stderr] 
[INFO] [stderr] error[E0412]: cannot find type `R_GE_gcontext` in this scope
...snip...
```

### DIFF
--- a/extendr-api/Cargo.toml
+++ b/extendr-api/Cargo.toml
@@ -44,4 +44,4 @@ tests-graphics = ["tests-minimal", "graphics"]
 tests-all = ["tests", "graphics"]
 
 [package.metadata.docs.rs]
-features = ["graphics"]
+features = ["graphics", "libR-sys/use-bindgen"]


yutannihilation:
First of all, can we agree on this direction? As commented on #415, there are many possible solutions.

Ilia-Kosenkov (approved):
As we seemed to agree to use bindgen, let's merge this PR.

yutannihilation:
Thanks!

PR #434, opened 2022-09-23
explicitly drop box ptr

sorhawell:
compiling extendr with latest rust nightly I get warnings for every #[extendr] macro invocation
```
warning: unused return value of `Box::<T>::from_raw` that must be used
     --> src/rdataframe/mod.rs:98:1
      |
   98 | #[extendr]
      | ^^^^^^^^^^
      |
      = note: `#[warn(unused_must_use)]` on by default
      = note: call `drop(from_raw(ptr))` if you intend to drop the `Box`
      = note: this warning originates in the attribute macro `extendr` (in Nightly builds, run with -Z macro-backtrace for more info)
```

the warnings come from [this update](https://github.com/rust-lang/git2-rs/pull/860), where it is required to explicitly drop boxed pointers

another contributer to rust-lang solved it like [this](https://github.com/rust-lang/git2-rs/pull/860/files):

I have used the same pattern in this PR

### DIFF
--- a/extendr-api/src/wrapper/altrep.rs
+++ b/extendr-api/src/wrapper/altrep.rs
@@ -492,7 +492,7 @@ impl Altrep {
             unsafe extern "C" fn finalizer<StateType: 'static>(x: SEXP) {
                 let state = Altrep::get_state_mut::<StateType>(x);
                 let ptr = state as *mut StateType;
-                Box::from_raw(ptr);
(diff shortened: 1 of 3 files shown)


multimeric:
I think https://github.com/rust-lang/rust/pull/99270 is the original original source of this change. Anyway, looks good!

Ilia-Kosenkov (approved):
Ok, looks good!

PR #436, opened 2022-10-22
Add support for `#[default = "..."]` in methods as well

### LINKED ISSUE #435: #[default = "NULL"] doesn't work in `impl` block
The use of `#[default = "NULL"] ` in the following example, seems to be ignored. This works fine for individual functions, but not when the function is part of an `impl` block.

```rust
struct Person {
    pub name: String,
}

#[extendr]
impl Person {
    fn new(#[default = "NULL"] name: &str) -> Self {
        Self {
            name: name.to_string(),
        }
    }

    fn set_name(&mut self, name: &str) {
        self.name = name.to_string();
    }

    fn name(&self) -> &str {
        self.name.as_str()
    }
}
```

Ilia-Kosenkov:
Fixes #435.
Unify how functions' and methods' args are handled when wrappers are generated. Slightly rework how formal and actual arguments are generated from `Arg` metadata.


extendr-api/src/metadata.rs
```diff
@@ -43,6 +43,37 @@ pub struct Metadata {
     pub impls: Vec<Impl>,
 }
 
+struct RArg {
+    name: String,
+    default: Option<&'static str>,
+}
+
+impl RArg {
```
multimeric, line 51: I think it's a bit misleading to call this `RArg` when it's actually representing a *Rust* argument that will be converted into R.

extendr-api/src/metadata.rs
```diff
@@ -215,25 +230,16 @@ fn write_method_wrapper(
...
+
+    // Skip a leading "self" argument.
+    // This is supplied by the environment.
+    let n_skip = r_args
+        .first()
+        .map_or(0, |a| if a.is_self() { 1 } else { 0 });
 
-    let formal_args = formal_args.join(", ");
-    let actual_args = actual_args.join(", ");
+    let formal_args = r_args.iter().skip(n_skip).map(|a| a.to_formal_arg());
```
multimeric, line 242: Could we not express this more concisely by just `.filter()`ing this instead of using `.skip()`?

multimeric (approved):

extendr-api/src/metadata.rs
Ilia-Kosenkov: I originally tried to expand the functionality of `Arg`, but then realized it is used in many places. Here I want to introduce a struct to contain metadata of R arguments, mainly sanitized name and default value (if present). Hence the name. So it is not really *Rust* argument but *R* argument. 

Maybe we can improve on naming by utilizing the difference between *parameter* and *argument*. Here R parameters are names with optional default values that go into R function definition, and R arguments are actual arguments supplied to the `.Call()` instruction within that wrapper.

I am open to suggestions.

extendr-api/src/metadata.rs
Ilia-Kosenkov: I propose a middle ground in form of `.skip_while(|a| a.is_self())`. This communicates the idea that we want to drop the `self` argument found at the beginning of the argument list.

multimeric: Sounds good to me.

extendr-api/src/metadata.rs
multimeric: Okay I take your point, I think this is a reasonable name.

Issue #437, opened 2022-12-17
Switch back the Windows toolchain from GNU to MSVC once again


yutannihilation:
(See https://github.com/extendr/rextendr/issues/216 for more context)

## Overview

extendr switched the Windows toolchain from MSVC to GNU when R 4.2 was released. At that time, that was the best choice because we can get free from the headaches of cross-compilation. But, now we have two big news to consider:

1. the CRAN machines now have Rust installed ([ref](https://stat.ethz.ch/pipermail/r-package-devel/2022q4/008638.html))
2. R-univese now supports R-Rust packages without any additional setups (because [the GNU target is now pre-installed on the GHA runner thanks to jeroen](https://github.com/actions/runner-images/pull/6661))

So, while we wanted to avoid the cross-compilation, I think the benefit of using MSVC toolchain is over the pain.

## Feasibility

It was I that actually changed the toolchain on addressing the release of R 4.2. The details can be found at https://github.com/extendr/extendr/issues/396. The actual changes made to each repo are the following pull requests. If you see the diffs, you'll notice there's no changes on Rust code, mainly on the testing infrastructure and on the documents. So, I think it's not difficult to revert the toolchain.

* libR-sys: https://github.com/extendr/libR-sys/pull/103
* extendr: https://github.com/extendr/extendr/pull/376
* rextendr: https://github.com/extendr/rextendr/pull/187
* document about the toolchain setups: https://github.com/extendr/libR-sys/pull/94

### Does the GNU toolchain actually work?

I'll try releasing my string2path package with GNU toolchain on CRAN. Let's see how it goes.

### What were the difficulties of relying on the cross compilation?

The only thing I can remember is that the doc tests don't run on cross-compiling. Were there any big difficulties that we suffered from?

yutannihilation:
@Ilia-Kosenkov 
Please let me know if you have any concern!

multimeric:
Can you please explain the advantages of using the MSVC toolchain? Also, what was the cross compilation that was happening? What target architecture on what compiler architecture? Finally, will the doc tests work now, or will it stop them working?

yutannihilation:
> Can you please explain the advantages of using the MSVC toolchain?

Since it's installed on the CRAN machine and the GHA runners, package authors can release there without providing precompiled binaries like I do.

> what was the cross compilation that was happening? What target architecture on what compiler architecture?

[...]

PR #439, opened 2022-12-17
Do not specify the toolchain on Windows with R >= 4.2

yutannihilation:
Related to https://github.com/extendr/extendr/issues/437

* Add a runner to test MSVC toolchain on R >= 4.2
* Remove the toolchain specification (`+$(TOOLCHAIN)`) from `Makevars.win`
* (minor) Follow-up of https://github.com/extendr/rextendr/pull/218

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -199,9 +199,7 @@ jobs:
       - name: Obtain 'rextendr'
         uses: actions/checkout@v2
         with:
-          # TODO
-          repository: yutannihilation/rextendr
-          ref: poc/support-msvc-toolchain-again
+          repository: extendr/rextendr
           path: ./tests/rextendr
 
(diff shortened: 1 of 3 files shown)


Ilia-Kosenkov:
`msvc` toolchain (as expected) does not execute doc tests https://github.com/extendr/extendr/actions/runs/3720454158/jobs/6309937796

How about we introduce a `rust-nightly` toolchain option for Windows and run tests there with `-Zdoctest-xcompile` (see e.g. https://github.com/rust-lang/cargo/issues/7040). I guess we can have an optional config variable with additional flags to pass to test phase.

yutannihilation:
This should be ready for review now.

Ilia-Kosenkov (approved):
Let's see how it goes. So far, the config looks much cleaner than it was before, so I guess this migration is a net positive for extendr.

PR #443, opened 2023-01-05
[ndarray]  Add `TryFrom<&Robj>` implementation for a broad set of scalar types

Ilia-Kosenkov:
*UPD* : I decided to commit unrelated to `ndarray` changes in separate PRs:
- [x] #444 
- [x] #446
- [x] #454
- Implement `TryFrom<&Robj>` for `ArrayView1<T>`
- Implement `TryFrom<&Robj>` for `ArrayView2<T>`
- Improve error reporting if conversion fails to `ArrayView_<T>` fails
- Add a set of conversion tests
- Implement `Debug` for `Rcplx` since it was missing
- Implement `Debug` for `Complexes`
- Add `abs()` to `Rfloat`

`T` is `i32`, `f64`, `c64`, `Rint`, `Rfloat`, `Rcplx`, `Rbool`, `Rstr`


Ilia-Kosenkov:
Ok I inadvertently hijacked #430. I will close that if this passes all the checks and reviews.

Ilia-Kosenkov:
I am apparently messed up trait implementation since `#[extendr(use_try_from = true)]` fails to compile for ndarray.

Ilia-Kosenkov:
Things look good mut the PR is huge - a lot of features were unimplemeted, which I had to fix here. I can I principle split it into several PRs if it would be more convenient.

extendr-api/src/robj_ndarray.rs
```diff
@@ -19,31 +19,106 @@ where
...
+                            .map_err(|err| Error::NDArrayError(err));
+                    } else {
+                        return Err($error_fn(robj.clone()));
+                    }
+                }
+                return Err(Error::ExpectedMatrix(robj.clone()));
+            }
+        }
+
+        impl<'a> TryFrom<Robj> for ArrayView2<'a, $type> {
```
multimeric, line 80: As above, please delegate.

extendr-api/src/robj_ndarray.rs
```diff
@@ -19,31 +19,106 @@ where
...
+                        return Err($error_fn(robj.clone()));
                     }
                 }
-                return Err($error_str);
+                return Err(Error::ExpectedMatrix(robj.clone()));
             }
         }
     };
 }
+make_array_view_1!(Rbool, Error::ExpectedLogical);
```
multimeric, line 100: Could we leverage `ToVectorValue` here to make this a generic impl instead of using macros? ie `impl<'a, T> TryFrom<Robj> for ArrayView1<'a, T> where T: ToVectorValue`.

multimeric: I also note we don't have non-R types like `i64` and `f32`. This suggestion would solve that.

extendr-api/tests/ndarray_try_from_tests.rs
```diff
@@ -0,0 +1,347 @@
+#[cfg(feature = "ndarray")]
+mod ndarray_try_from_tests {
+    use extendr_api::prelude::*;
+
+    #[test]
+    fn integer_matrix_2d() {
```
multimeric, line 6: These tests all seem to have the same structure. Could you reduce redundancy using a `macro_rules!`?

tests/extendrtests/src/rust/src/ndarray.rs
```diff
@@ -0,0 +1,33 @@
...
+/// Calculate Euclidean distance matrix
+/// Test case adopted from https://github.com/mikemahoney218/examplerust/blob/23d21b1ced4e24b7a7c00dd36290114dc1bbd113/src/rust/src/lib.rs#L5
+/// @param a : Matrix of real values or `NULL`
+/// @export
+#[extendr(use_try_from = true)]
+fn euclidean_dist(a: Nullable<ArrayView2<Rfloat>>) -> Nullable<Doubles> {
+    if let NotNull(a) = a {
+        let nrow = a.nrows();
+
+        let result = (0..(nrow - 1))
```
multimeric, line 12: May just iter `a.rows()` for a clearer intent?

tests/extendrtests/tests/testthat/test-ndarray.R
```diff
@@ -0,0 +1,12 @@
+test_that("[ndarray] Euclidean distance", {
+    m <- matrix(1.0 * 1:6, nrow = 3, ncol = 2)
+    dist <- euclidean_dist(m)
+
+    expect_equal(dist, c(sqrt(2), 2 * sqrt(2), sqrt(2)))
```
multimeric, line 5: Could you compare to R's native `dist(x, method = "euclidean")`?

multimeric (changes requested):

extendr-api/src/error.rs
```diff
@@ -81,6 +81,9 @@ pub enum Error {
 
     ExpectedExternalPtrType(Robj, String),
     Other(String),
+
+    #[cfg(feature = "ndarray")]
+    NDArrayError(ndarray::ShapeError),
```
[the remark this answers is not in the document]
Ilia-Kosenkov: I found no other `Error` types in ndarray and assumed it is enough to wrap it. See [this](https://docs.rs/ndarray/latest/ndarray/struct.ShapeError.html).

CHANGELOG.md
```diff
@@ -4,7 +4,8 @@
 
 ### Added
 
-- `Sum` for scalars like `Rint`, `Rfloat` and `Rcplx`, which accept `Iterator<Item = &Rtype>` [[#454]](https://github.com/extendr/extendr/pull/454) 
+- [**ndarray**] `TryFrom<&Robj>` for `ArrayView1<T>` and `ArrayView2<T>`, where `T` is `i32`, `f64`, `c64`, `Rint`, `Rfloat`, `Rcplx`, `Rstr`, `Rbool` [[#443]](https://github.com/extendr/extendr/pull/443)
+- `Sum` for scalars like `Rint`, `Rfloat` and `Rcplx`, which accept `Iterator<Item = &Rtype>` [[#454]](https://github.com/extendr/extendr/pull/454)
```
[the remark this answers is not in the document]
Ilia-Kosenkov: I am not following, what is not needed?

extendr-api/src/robj_ndarray.rs
```diff
@@ -2,7 +2,7 @@
 use ndarray::prelude::*;
 use ndarray::{Data, ShapeBuilder};
 
-use crate::prelude::dim_symbol;
+use crate::prelude::{c64, dim_symbol, Rcplx, Rfloat, Rint};
 use crate::*;
 
 impl<'a, T> FromRobj<'a> for ArrayView1<'a, T>
```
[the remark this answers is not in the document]
Ilia-Kosenkov: I suggest to keep it until we enable `use_try_from=true` by default.

extendr-api/src/error.rs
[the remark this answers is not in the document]
multimeric: Fair. Should you maybe just call it `NdArrayShapeError` in case they later add new error types?

extendr-api/src/robj_ndarray.rs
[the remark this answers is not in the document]
multimeric: Sounds good.

CHANGELOG.md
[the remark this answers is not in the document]
multimeric: This line in the changelog about `Sum`. It was in the other PR.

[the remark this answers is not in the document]
Ilia-Kosenkov: I think it is still correct. You see in diff two lines because I added `ndarray` atop `Sum`. And `Sum` was indeed merged as a separate PR before, I think it is worth mentioning here.

extendr-api/src/robj_ndarray.rs
Ilia-Kosenkov: Ok so the idea behind this is that we have type-specific errors. I see no way of achieving the same using blanket implementation for all `T : ToVectorValue`. 
The case of `i64` and `f32` are different. `ArrayView` here is a view, zero-copy, and since R supports only 32bit integers and 64bit doubles, there are simply no other valid variants apart from `u32`.

extendr-api/src/robj_ndarray.rs
```diff
@@ -19,31 +19,106 @@ where
...
+                if let Some(v) = robj.as_typed_slice() {
+                    Ok(ArrayView1::<'a, $type>::from(v))
+                } else {
+                    Err($error_fn(robj.clone()))
+                }
+            }
+        }
+    };
+
+    ($type: ty, $error_fn: expr) => {
```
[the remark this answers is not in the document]
Ilia-Kosenkov: `TryFrom<&Robj>` is default implementation (since we want to consume refs) and I delegated `TryFrom<Robj>`

extendr-api/src/robj_ndarray.rs
Ilia-Kosenkov: Should be the same

extendr-api/src/robj_ndarray.rs
Ilia-Kosenkov: I allowed reinterpreting `integers` as `u32` and added support here.

extendr-api/tests/ndarray_try_from_tests.rs
Ilia-Kosenkov: Honestly, no idea how to do that, since every case is slightly different from another. Test cases are pretty much hand-picked. Maybe [this](https://docs.rs/rstest/latest/rstest/) can help, but I never used it & I'd prefer to handle it in a separate PR (to unblock progress for others).

tests/extendrtests/src/rust/src/ndarray.rs
Ilia-Kosenkov: I copy-pasted an example provided by one of the users (see github link) and updated it so that it does not use types other than extendr types (like `Rfloat`). In this case it uses current row number to execute inner loop, and I see no simple solution here.

tests/extendrtests/tests/testthat/test-ndarray.R
Ilia-Kosenkov: Good idea, I actually increased the size of the test matrix and randomized test input. Seems to work fine

CGMossa:
Oh.. Do we need to do something here?

.gitignore
```diff
@@ -33,3 +33,5 @@
 extendr.Rproj
 tests/extendrtests.Rcheck/
 tests/extendrtests_0.1.0.tar.gz
+
+**/.cargo/*
```
multimeric, line 37: Can you fix this? Remove empty line and add final newline

multimeric (approved):

PR #444, opened 2023-01-07
`Debug` trait for `Rcplx` and `Complexes`

Ilia-Kosenkov:
- Add proper `Debug` trait implementation for `Rcplx`
- Add proper `Debug` trait implementation for `Complexes`
- Add `Rfloat::abs()` function
- Fix `Rstr` `Debug` implementation
- Correct debug strings for `NA` values
- Make `Rany::Complexes` wrap `Complexes` instead of `Robj`
- Redirect `Debug` for `Rany::Complexes` to `Debug` implementation for `Complexes`
- Add tests to check `Debug` implementation for scalars and vector wrappers


extendr-api/src/scalar/rcplx_default.rs
```diff
@@ -41,11 +41,15 @@ impl CanBeNA for c64 {
 /// Rcplx has a special NA value, obtained from R headers via R_NaReal.
 ///
 /// Rcplx has the same footprint as R's complex value allowing us to use it in zero copy slices.
-#[derive(Clone, Copy, Debug, Default, PartialEq)]
+#[derive(Clone, Copy, Default, PartialEq)]
+#[repr(C)]
```
CGMossa, line 45: why this repr?

CGMossa (approved):

extendr-api/src/scalar/rcplx_default.rs
Ilia-Kosenkov: Hm, maybe it is excessive, but it would not hurt in this case. The idea is that we want to be able to map an array of R's complex structs to an array of rust structs. Basically, we need to ensure that `Rcmplx` is 2 x f64 in size, exactly, and that `Re` goes first. 
A direct quote from [docs](https://doc.rust-lang.org/reference/type-layout.html):
> The only data layout guarantees made by this representation are those required for soundness. They are:

>    The fields are properly aligned.
>    The fields do not overlap.
>    The alignment of the type is at least the maximum alignment of its fields.

Basically, Rust knows how a struct is stored, but it is not necessarily equivalent to how C stores it. So in order to avoid any memory-layout-related issues, we ask our complex type to be 'as in C'.

CGMossa: I just don't see the same annotation on much anything else. That's why I asked.
I believe using repr(C) unnecessarily may hurt actually. I don't have a reasonable explanation for this though. 
Closing.

Issue #313, opened 2021-10-29
refactor: prefer Option<T> over Nullable<T>  


mlondschien:
It would be nice to have a blanket implementation

```rust
impl<T> From<Option<T>> for Robj
where
    T: Into<Robj>,
{
    fn from(value: Option<T>) -> Self {
        if let Some(some) = value {
            some.into()
        } else {
            ().into()
        }
    }
}
```

This is not possible as this clashes with `ToVectorValue` implementations for `Option<...>`. These return the correct NA value for the type instead of `NULL`.

Related: #308.

andy-thomason:
Rust's implementations have to be precise which may relax in the future to allow
some overlap.

There is a blanket `Nullable<T>` which is less likely to clash with other implemementations.

As we have it `Option<T>` is for NAs and `Nullable<T>` is for NULLs.

mlondschien:
> Rust's implementations have to be precise which may relax in the future to allow some overlap.

I don't understand this. The code I posted above works, as long as we remove all implementations of `ToVectorValue` for `Option<...>`. However, then `vec![Some(12), None]` will be converted to R as `c(12L, NULL)` instead of `c(12L, NA)`.

andy-thomason:
Yes. This is true. The current interpretation of `Option` is NA rather than `NULL`.

Now we have `Rint` and `Rfloat` we might drop this as it is counter intuitive, we might expect
`Option` to be the same as `Nullable`.

andy-thomason:
Perhaps that now we have the scalar types `Rint` and `Rfloat` that handle NAs, we should reserve
`Option` for NULL instead of NA and drop Nullable. Would this make more sense?

clauswilke:
This sounds reasonable to me. But we probably need a complete implementation of scalar types first. Which ones are still missing? Only `Rbool`?

Ilia-Kosenkov:
`Rbool`, `Rcomplex` at least. Ideally, we should also make `Rbyte`.

andy-thomason:
We should revise ToVectorValue

Ilia-Kosenkov:
I agree, I spent a lot of time trying to understand how it works while working on Integers & Doubles.

PR #446, opened 2023-01-14
Improve `Nullable<T>` conversions & support

Ilia-Kosenkov:
- Add `TryFrom<RObj>` and its variations for `Nullable<T>`
- Add conversions to  and from `Options<T>`
- Add `map()` method that works similar to that of `Options<T>`, allowing acting on the state while propagating `NULL`
- Add a few doctests
- Add integration test
- Fix integration tests setup/config, update with newest versions of `{rextendr}` and `{roxygen2}`


extendr-api/src/wrapper/nullable.rs
```diff
@@ -77,3 +131,88 @@ where
         }
     }
 }
+
+impl<T> Into<Option<T>> for Nullable<T> {
```
multimeric, line 135: I'm worried that we're using `Optional<T>` to represent a `T` that may be `NA` elsewhere in extendr. Should we consider adding a counterpart to this `Nullable<T>` that's like `NAable<T>` so we can drop the general `Option` conversion?

extendr-api/src/wrapper/nullable.rs
```diff
@@ -77,3 +131,88 @@ where
...
+    fn into(self) -> Option<T> {
+        match self {
+            Nullable::NotNull(value) => Some(value),
+            _ => None,
+        }
+    }
+}
+
+impl<'a, T> Into<Option<&'a T>> for &'a Nullable<T> {
+    /// Convert a Nullable type into Option
```
multimeric, line 153: Clarify that this is for a reference

extendr-api/src/wrapper/nullable.rs
multimeric, line 135: Why are we using `Into` here rather than `From`?

tests/extendrtests/DESCRIPTION
```diff
@@ -1,23 +1,25 @@
...
 Authors@R: c(
     person("Andy", "Thomason", , "andy@andythomason.com", c("aut")),
     person("Claus O.", "Wilke", , "wilke@austin.utexas.edu", c("aut", "cre"),
-      comment = c(ORCID = "0000-0002-7470-9261"))
+      comment = c(ORCID = "0000-0002-7470-9261")),
+    person("Ilia", "Kosenkov", , "ilia.kosenkov@outlook.com", c("aut"), comment = c(ORCID = "0000-0001-5563-7840"))
   )
 Description:
-  This package provides tests for extendr functionality used from
+  Provides tests for extendr functionality used from
```
multimeric, line 11: These R changes seem unrelated to the PR?

extendr-api/src/wrapper/nullable.rs
```diff
@@ -77,3 +131,88 @@ where
...
+    /// ```
+    /// use extendr_api::prelude::*;
+    /// test! {
+    ///     assert_eq!(Into::<Option<&i32>>::into(&Nullable::Null), None);
+    ///     assert_eq!(Into::<Option<&i32>>::into(&Nullable::NotNull(42)), Some(&42));
+    /// }
+    /// ```
+    fn into(self) -> Option<&'a T> {
+        match self {
```
multimeric, line 162: There's some redundancy here. Should we either a) delegate some of these impls: https://stackoverflow.com/a/56567737/2148718 or b) solve it with a macro (https://github.com/auto-impl-rs/auto_impl/). 

Because you haven't considered `mut` references here either.

extendr-api/src/wrapper/nullable.rs
Ilia-Kosenkov: A valid point. However, due to the lack of docs, people just prefer to use known types, like `Option<i32>` not `Rint`. I am honestly OK with dropping options conversion at all, but it would be nice to have more methods to work with nullable type, like `map`, `and_then`, etc.

tests/extendrtests/DESCRIPTION
Ilia-Kosenkov: I cleaned a little bit integration test. I can of course take it out into a separate PR if you feel like it should be done.

extendr-api/src/wrapper/nullable.rs
Ilia-Kosenkov: No particular reason other than me not knowing the convention/best practices :)

extendr-api/src/wrapper/nullable.rs
Ilia-Kosenkov: Maybe it is an argument for dropping `Option` here altogether and instead borrowing some methods from `Option` itself?

tests/extendrtests/DESCRIPTION
Ilia-Kosenkov: Ok, I moved these bits to #447. As soon as it is merged, I'll drop some of the commits from this branch.

extendr-api/src/wrapper/nullable.rs
multimeric: I think conversion from `Nullable` to `Option` is fine and idiomatic Rust. I just think we shouldn't have an automatic conversion from `Robj` to `Option<T>` specifically. Rather we should have an intermediate `NAable`. But this is probably a discussion for elsewhere.

multimeric: Yep, always use `From` instead of `Into` if you can. There are some scenarios where you can't do that though.

tests/extendrtests/DESCRIPTION
multimeric: Once it's merged this part of the diff should disappear right?

extendr-api/src/wrapper/nullable.rs
multimeric: Interesting idea, but I don't think that would be any cleaner. I think that might actually result in more code and redundancy rather than less.

extendr-api/src/wrapper/nullable.rs
multimeric: Actually on second thoughts, it's not as confusing as it seems. When we have a vector, a given element can be `NA`, but a given element can't be `NULL`. So there's never a point at which there is confusion between `NA` and `NULL`. 

For example when we convert a `Robj` to `Optional<Vector<i32>>` the "optional" part is very clearly "is this null or not", because being `NULL` means there is no vector at all. If we convert `Robj` to `Vector<Optional<i32>>` the optional part is "is this particular element `NA`". So as long as this is documented, I think it's fine.

Do you agree?

Ilia-Kosenkov: Should we disallow conversion from `Option` to `Nullable` then? In that case, `Option` might represent `NA`-aware value.

multimeric: No, I don't think so. There shouldn't ever be overlap between `Option<T>` derived from `Nullable` and `Option<T>` derived from `NA`, so it's not confusing.

extendr-api/src/wrapper/nullable.rs
multimeric: I agree that it doesn't make much sense to take a `&mut Nullable<T>` because there's no actual mutable state there. But on the other hand, there's no reason we can't convert that into an `Option`. It's just a reference with my permissions than usual.

Ilia-Kosenkov: So, right now we have no constraints on `T` when implementing `Nulalble<T> <--> Option<T>`. In reality though, `Nullable<T>` can be only constructed from something that can be `TryFrom<RObj>` or `Into<Robj>` (or some variation). I'd say we could put some trait constraints on `T` everywhere (something that says `T` should be `Robj` - compatible), in which case I see no real reason to expose `mut` reference support, since there is no scenario where we allow such things.

multimeric: Fair enough, there is no practical use case for converting mutable references.

But on that note, I like your idea for locking down the `impl` for `Nullable` using trait bounds to enforce that it's always a `Nullable<T> where T: TryFrom<Robj>`. I like the clear message of putting the trait bounds on the actual `enum` itself, but because https://github.com/rust-lang/rust/issues/44491 is still open, I guess it's best to just put the bounds on the `impl`.

Ilia-Kosenkov: So if you convert from `Nullable<T>` to `Option<T>`, we should require `T : TryFrom<Robj>`, it is a scenario for accepting a `Nullable<T>`. However, if we convert from `Option<T>` to `Nullable<T>`, we require `Into<Robj>` since it is a scenario of returning a `Nullable<T>`. The majority of types implement both traits, but not e.g. `ndarray.`
So we should be able to go `Nullable<ArrayView>` -> `Option<ArrayView>`, and then take an `Option<ArrayBase>` and convert to `Nullable<ArrayBase>` to return it back as nullable.

CGMossa (approved):

multimeric:
Does this PR relate to https://github.com/extendr/extendr/issues/313 in any way?

Ilia-Kosenkov:
Not directly, but maybe it covers some of the cases listed there. The original issue is about `Option<T>`, which we reserve for `NA`.

PR #450, opened 2023-01-17
Document the robj_ndarray module

multimeric:
* Document the robj_ndarray module 
* Add conversion for owned arrays, with tests


multimeric:
Two questions for whoever reviews this:

1. Why is the WIndows build failing? Is that at all relating to my PR?
2. Why does it seem like the doctests aren't being run for this PR. I've added a number of examples that do run correctly and it would be nice to know that they're being checked.

extendr-api/src/robj_ndarray.rs
```diff
@@ -1,3 +1,57 @@
+//! Defines conversions between R objects and the [ndarray](https://docs.rs/ndarray/latest/ndarray/) crate, which offers native Rust array types and numerical computation routines.
+//!
+//! To enable these conversions, you must first enable the `ndarray` feature for extendr:
+//! ```toml
+//! [dependencies]
+//! extendr-api = { version = "0.3.1", features = ["ndarray"] }
+//! ```
+//!
+//! Specifically, extendr supports the following conversions:
+//! * [`Robj` → `ArrayView1`](FromRobj#impl-FromRobj<%27a>-for-ArrayView1<%27a%2C%20T>), for when you have an R vector that you want to analyse in Rust:
```
Ilia-Kosenkov, line 10: In #443 I am trying to add support for `TryFrom<Robj>`, which is the proper way to handle conversions. Should we document it instead of obsolete `FromRobj` or wait till I merge my PR?

multimeric: For the sake of simplicity, maybe we just leave my PR as is, and then we can update the link in your PR? Unless yours gets merged first, in which case I'll do the fix here.

Ilia-Kosenkov (approved):
Ok, LGTM

constantingoeldel:
Thanks for the documentation, this already helped me!

multimeric:
Going to hold off on merging this until we merge #460, so I can rewrite to use `rstest`.

multimeric:
Has now been updated to use `rstest`. Once the CI is running again, this is good to go.

Ilia-Kosenkov (changes requested):
How about we also mention in the docs that we can handle `ArrayView<Rint>` and friends?

Ilia-Kosenkov (approved):

Issue #248, opened 2021-08-01
clippy::not_unsafe_ptr_arg_deref error on extendr_module! macro


yutannihilation:
I got this when I run `cargo +nightly clippy` in `src/rust` of an R package using extendr.

```
error: this public function might dereference a raw pointer but is not marked `unsafe`
  --> src/lib.rs:13:1
   |
13 | / extendr_module! {
14 | |     mod foo;
15 | |     fn hello_world;
16 | | }
   | |_^
   |
   = note: `#[deny(clippy::not_unsafe_ptr_arg_deref)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#not_unsafe_ptr_arg_deref
   = note: this error originates in the macro `extendr_module` (in Nightly builds, run with -Z macro-backtrace for more info)
```

andy-thomason:
Interesting. Do you know which code caused it?

You can run `cargo inspect` if you install it which will expand the proc macros.

yutannihilation:
I couldn't figure out how to run clippy on expanded code, but I suspect this line:

https://github.com/extendr/extendr/blob/dc5d90106a969654f2b792b9f457e3df1d324641/extendr-macros/src/extendr_module.rs#L116

yutannihilation:
Confirmed moving `unsafe` to before `fn` will fix the warning. But I'm not sure if this is a false alarm or not.

``` patch
-        pub extern "C" fn #module_init_name(info: * mut extendr_api::DllInfo) {
+        pub unsafe extern "C" fn #module_init_name(info: * mut extendr_api::DllInfo) {
-            unsafe { extendr_api::register_call_methods(info, #module_metadata_name()) };
+            extendr_api::register_call_methods(info, #module_metadata_name());
```

clauswilke:
Is it because it takes a raw pointer as argument?

yutannihilation:
Yes, I think so. What I meant by "false alerm" is that it probably doesn't matter whether the function is marked as unsafe or not, because the unsafety isn't visible to the external caller of the function anyway.

andy-thomason:
Did we make this fix?

We could also use #[allow(...)] for warnings.

yutannihilation:
No. Let's add `#[allow()]`.

daniellga:
Hi, I am seeing these same warnigns on `extendr_module!` and on each function called in `extendr_module!`. Was I supposed to?

```
rustup show
[...]

PR #451, opened 2023-01-18
Missing `unsafe` on module wrappers

CGMossa:
Running `cargo clippy` on a freshly minted rust-r-package from `{rextendr}`-template, i.e.

```rust
use extendr_api::prelude::*;

/// Return string `"Hello world!"` to R.
/// @export
#[extendr]
fn hello_world() -> &'static str {
    "Hello world!"
}

// Macro to generate exports.
// This ensures exported functions are registered with R.
// See corresponding C code in `entrypoint.c`.
extendr_module! {
    mod EID_ABM;
    fn hello_world;
}
```

yields this error:
```
this public function might dereference a raw pointer but is not marked `unsafe`
for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#not_unsafe_ptr_arg_deref
`#[deny(clippy::not_unsafe_ptr_arg_deref)]` on by default
```
And if I look at the expanded code, I find places where

```rust
#[allow(non_snake_case, clippy::not_unsafe_ptr_arg_deref)]
```

Thus I've added it to the missing part here.

I need a review on this from someone who knows what's going on.

### DIFF
--- a/extendr-macros/src/extendr_module.rs
+++ b/extendr-macros/src/extendr_module.rs
@@ -92,7 +92,7 @@ pub fn extendr_module(item: TokenStream) -> TokenStream {
         }
 
         #[no_mangle]
-        #[allow(non_snake_case)]
+        #[allow(non_snake_case, clippy::not_unsafe_ptr_arg_deref)]
         pub extern "C" fn #wrap_make_module_wrappers(
             use_symbols_sexp: extendr_api::SEXP,
             package_name_sexp: extendr_api::SEXP,


multimeric:
Do you think you could actually add a CI test for this lint? e.g. using https://github.com/actions-rs/clippy-check and `cargo clippy -- -A clippy::all -W clippy::not_unsafe_ptr_arg_deref `? Might be a nice time to start progressively adding clippy lints.

CGMossa:
That's a good idea. But it is a whole thing of its own.. Please file a
different issue in this and maybe someone can attempt this (like me or
someone who is a CI Warrior)

multimeric:
But it will only test this single lint. It's like a unit test for your change.

CGMossa:
Is that a good idea to do only for one lint at a time? I don't think so. I
might be wrong.

yutannihilation:
> Do you think you could actually add a CI test for this lint?

We decided not to add, but feel free to file a new issue.

https://github.com/extendr/extendr/issues/54#issuecomment-757345207

I *think* we shouldn't because we might not have the bandwidth to fix the warnings, but please ignore me.

yutannihilation (approved):
In my understanding, the problem is that `SEXP` is a raw pointer with dereferencing, and clippy got smart enough to detect it (I guess at the time of https://github.com/extendr/extendr/issues/248, it wasn't). So, in short, as the problem is the same as https://github.com/extendr/extendr/issues/248, I think the same fix as https://github.com/extendr/extendr/issues/248 is enough.

https://docs.rs/libR-sys/latest/libR_sys/type.SEXP.html

CGMossa:
Alright, then let us merge this.

PR #453, opened 2023-01-19
Add mutable ndarray conversions

multimeric:
Lets us convert from a `&mut Robj` to `ArrayViewMut1<T>` and `ArrayViewMut2<T>`. This is neat because it means we can manipulate R arrays passed to Rust or allocated in Rust directly using the high level ndarray API without copying.

See discord discussion for more detail: https://discord.com/channels/751419551132155925/800474459630731335/1065613818409013339.

### DIFF
--- a/extendr-api/src/robj_ndarray.rs
+++ b/extendr-api/src/robj_ndarray.rs
@@ -19,6 +19,20 @@ where
     }
 }
 
+impl<'a, T> TryFrom<&'a mut Robj> for ArrayViewMut1<'a, T>
+where
+    Robj: AsTypedSlice<'a, T>,
+{
+    type Error = Error;
+    fn try_from(robj: &'a mut Robj) -> std::result::Result<Self, Error> {
+        if let Some(v) = robj.as_typed_slice_mut() {
+            Ok(ArrayViewMut1::<'a, T>::from(v))
+        } else {
+            Err(Error::TypeMismatch(robj.to_owned()))
+        }
+    }
+}
+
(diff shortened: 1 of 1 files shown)


Ilia-Kosenkov:
It would be nice to first merge other ndarray PRs. I for instance introduced a dedicated error type for failed conversion into array view. https://github.com/extendr/extendr/blob/89c5b471271e35884d73192ba13296676098366c/extendr-api/src/error.rs#L85-L86

multimeric:
I agree, although your PR is a draft so I assumed it wasn't finished?

Ilia-Kosenkov:
It's finished but I am moving parts of that PR into smaller PRs to simplify reviewing :) I need your approval on #446 (since you already started reviewing it). I will finish the rest over the weekend

multimeric:
Note to self: we need to strongly warn users against doing this on existing R objects that are user provided, because it will affect all variables that point to the same array: http://adv-r.had.co.nz/C-interface.html (under "Modifying inputs"). The use case I have in mind for this is creating the R array data within Rust, using `Robj::alloc_vector` or a wrapper constructor like `Strings::new()`, *then* wrapping the vector in an `ndarray` type, doing some mutable manipulations on that `ndarray`, and then returning the `Robj` to R.

CGMossa:
Uhm.. What's the status of this?

multimeric:
I still intend to finish it, but it's in the doldrums while I work out how to handle `MAYBE_SHARED` in extendr: https://github.com/extendr/libR-sys/issues/139.

PR #454, opened 2023-01-20
Implement `Sum` for Iterator of `&Rtype`

Ilia-Kosenkov:
Self-explanatory. We have `Sum` for `Iterator<Item = Rfloat>` but not for `Iterator<Item = &Rfloat>`. This PR adds missing implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


multimeric (approved):
lgtm. Thanks for updating the changelog, too.

PR #460, opened 2023-01-24
Add rstest

multimeric:
Plus demonstration tests

### DIFF
--- a/extendr-api/src/robj_ndarray.rs
+++ b/extendr-api/src/robj_ndarray.rs
@@ -94,138 +94,146 @@ where
     }
 }
 
-#[test]
-fn test_from_robj() {
-    test! {
-        assert_eq!(
-            <ArrayView1<f64>>::from_robj(&Robj::from(1.)),
-            Ok(ArrayView1::<f64>::from(&[1.][..]))
-        );
-        assert_eq!(
-            <ArrayView1<i32>>::from_robj(&Robj::from(1)),
-            Ok(ArrayView1::<i32>::from(&[1][..]))
-        );
-        assert_eq!(
-            <ArrayView1<Rbool>>::from_robj(&Robj::from(true)),
-            Ok(ArrayView1::<Rbool>::from(&[TRUE][..]))
-        );
+#[cfg(test)]
+mod test {
+    use super::*;
+    use rstest::rstest;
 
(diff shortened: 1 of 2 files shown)


multimeric:
This PR demonstrates the test parametrisation feature of `rstest`. The main advantage of this is cleaner code, less redundancy (ie I was able to remove boilerplate like `Robj::try_from` from each test case), plus it turns each of the copies of the test into its own test case which makes the reporting more fine grained.

multimeric:
I suggest looking at this diff with whitespace diff turned off, and using the split view.

extendr-api/src/robj_ndarray.rs
```diff
@@ -94,138 +94,146 @@ where
...
-        // assert_eq!(mx[[1, 1]], 6);
-        // assert_eq!(mx[[2, 1]], 7);
-        // assert_eq!(mx[[3, 1]], 8);
+#[cfg(test)]
+mod test {
+    use super::*;
+    use rstest::rstest;
+
+    #[rstest]
+    fn test_from_robj() {
```
Ilia-Kosenkov, line 103: Any idea of how to switch thist test to `rstest`?

multimeric: Not sure what you mean. This one was changed to `rstest`?

Ilia-Kosenkov: Sorry I did not explain myself. Can we use a fixture to provide test case data for this test as well, right now it is a lot of copy-paste.

CGMossa:
Now that I've had to write a test, I now see the point of having such a test-framework in Extendr. @Ilia-Kosenkov What do you think?

Unfortunately, I don't know much about this. I'd leave a review, but it would be a blanket approval, as I think it looks fine, and this `rstest`-crate looks fairly stable + alive.
"282,214 downloads per month"
"Used in [343 crates (338 directly)](https://lib.rs/crates/rstest/rev)"

That's good enough for me.

multimeric:
One issue I've noticed is that `rstest` forces all test cases for the same test function to have the same type, which can lead to some convoluted type signatures (see the Higher-Rank Trait Bounds in this PR for an example). On the other hand, it might encourage us to split our test functions into more sane units which might be a good thing.

Ilia-Kosenkov (approved):

PR #461, opened 2023-01-25
Document how to make modules available to R packages

JosiahParry:
This PR adds documentation to `extendr_module` to describe how to make module functions available to R 

Per https://discord.com/channels/751419551132155925/751419551132155928/1067806341491609772


multimeric:
The only thing I'm worried about here is recommending that they `use submodule` in the `#[extendr]` block rather than just asking them to `use submodule` in normal Rust and then defining the entire extendr module in one place. Would that not be simpler and cleaner?

Also, could you please pull from master please @JosiahParry?

sorhawell:
From discussion it appears to me that the consensus is now, that extendr-modules are an ok pattern.
https://discord.com/channels/751419551132155925/800474459630731335/1071355741950197770

The need for `use utils::get_utils_metadata;` might soon be obsolete with #469, but no harms done if not updated immediately when #469 merges.

I think this brings value as is.

multimeric (approved):

JosiahParry:
That should be fixed now.

JosiahParry:
The failed CI runs are due to upstream error in ubuntu 20.04 r-lib action.

CGMossa (approved):
I approve of this, but the metadata part needs to be removed. As soon as my PR is merged we can do that.

JosiahParry:
@CGMossa removed the note about get_{mod}_metadata

extendr-macros/src/lib.rs
```diff
@@ -102,6 +102,34 @@ pub fn extendr(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+/// ```
+///
+/// In order to include these functions in your R package you must include them in `lib.rs`.
+/// To do so you must include the `mod` and `use` keywords to make them available.
+/// Lastly, the module must be imported in the `extendr_module` macro.
+///
+///
+/// ```ignore
+/// mod utils;
+/// use utils::get_utils_metadata;
```
CGMossa, line 124: this needs to be removed as well

CGMossa:
I *think* one module has to act as the a "root" though. But I'm unsure...

multimeric (changes requested):
This is basically good to go, but please add a changelog entry under a "Unreleased" title, using the `[docs]` tag. This won't be in the next release but we need to get in the habit of updating the changelog.

PR #462, opened 2023-01-25
document external pointer

JosiahParry:
This PR adds a few lines of documentation for the external pointer per https://discord.com/channels/751419551132155925/1067823507712516146/1067824501598986421.

Adds language that might be a bit more helpful for R users to understand. Additionally makes a note that in order to dereference the objects must have the Clone trait or be a reference.

### DIFF
--- a/extendr-api/src/wrapper/externalptr.rs
+++ b/extendr-api/src/wrapper/externalptr.rs
@@ -7,8 +7,8 @@ use std::fmt::Debug;
 /// Your R code can use external pointers to reference objects in Rust memory.
 ///
 /// External pointers can be made avaialable as their underlying type by dereferencing. 
-// In order to deference, the object must either implement the `Copy` trait or be a reference—e.g. `&T`. 
-//
+/// In order to deference, the object must either implement the `Copy` trait or be a reference—e.g. `&T`. 
+///
 /// ```
 /// use extendr_api::prelude::*;
 /// test! {
@@ -19,6 +19,7 @@ use std::fmt::Debug;
 ///     assert_eq!(*extptr2, 1);
 /// }
 /// ```
+///
 #[derive(PartialEq, Clone)]
 pub struct ExternalPtr<T: Debug + 'static> {
     /// This is the contained Robj.
--- a/extendr-api/src/wrapper/externalptr.rs
+++ b/extendr-api/src/wrapper/externalptr.rs
@@ -2,12 +2,12 @@ use super::*;
 use std::any::Any;
 use std::fmt::Debug;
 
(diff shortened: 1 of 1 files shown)


JosiahParry:
The one error is from https://github.com/extendr/extendr/actions/runs/4007428814/jobs/6896132941 with a cryptic 
`Process completed with exit code 1` under section "Check code formatting". 

I suspect this is because a few lines only had `//` as opposed to `///`. New commit to fix.

JosiahParry:
CI errors are due to tests of graphic devices. This PR does not add or modify tests.

yutannihilation:
Yeah, the API version is bumped on R-devel. I'll take care of it, so you can ignore. (I have no idea what's happening on the Windows runner with R-release, but let's forget about it now...)

multimeric:
Might be worth mentioning that `ExternalPointer::new()` is the recommended way to create one.

Also isn't the "the object must either implement the `Copy` trait" thing just a function of it being a pointer? We put the original object in a box and then [consumed the box](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.into_raw), so Rust doesn't really know who owns the original object.

yutannihilation:
@JosiahParry 
The graphics device things should be fixed. It seems you have trailing spaces in these lines. Just in case you don't notice, instead of editing carefully by hand and praying, you can just run `cargo fmt`!

https://github.com/extendr/extendr/actions/runs/4017050519/jobs/6927322567

JosiahParry:
Added cargo fmt :)

extendr-api/src/wrapper/externalptr.rs
```diff
@@ -3,6 +3,11 @@ use std::any::Any;
 use std::fmt::Debug;
 
 /// Wrapper for creating R objects containing any Rust object.
+/// External pointers (`ExternalPtr`) are accessible from R as an object of class `externalptr`.
+/// Your R code can use external pointers to reference objects in Rust memory.
+///
+/// External pointers can be made avaialable as their underlying type by dereferencing.
```
multimeric, line 9: Clarify that this is in Rust.

extendr-api/src/wrapper/externalptr.rs
```diff
@@ -3,6 +3,11 @@ use std::any::Any;
...
+/// In order to deference, the object must either implement the `Copy` trait or be a reference—e.g. `&T`.
```
multimeric, line 10: I still don't get this. Doesn't `extptr.deref()` return a `&T`? So it doesn't matter if the `T` implements copy?

PR #467, opened 2023-01-29
Use Rf_mkCharLenCE() instead of Rf_mkCharLen()

yutannihilation:
Just happened to find this. This can be a problem only on Windows with R < 4.2 (`Rf_mkCharLen()` uses the native encoding), where UTF-8 is not R's default encoding, so this is almost just a nitpicking. 

For reference, cpp11 also uses `Rf_mkCharLenCE()`.

https://github.com/r-lib/cpp11/blob/c22dc9b611fe7d56ddfc57add2e01b1424869c0f/inst/include/cpp11/r_string.hpp#L19-L20

### DIFF
--- a/extendr-api/src/robj/into_robj.rs
+++ b/extendr-api/src/robj/into_robj.rs
@@ -6,7 +6,13 @@ pub(crate) fn str_to_character(s: &str) -> SEXP {
         if s.is_na() {
             R_NaString
         } else {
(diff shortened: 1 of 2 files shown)


Ilia-Kosenkov:
Do we need to test this explicitly? Perhaps use a non-ASCII string to see how it is handled?

yutannihilation:
I don't think so. We can test this only on R 4.1, which will retire soon.

Ilia-Kosenkov (approved):

PR #469, opened 2023-02-03
Fix `extendr_module` importing

CGMossa:
The issue is that these `get_<module>_metadata` are generated in their respective modules, but then they aren't realy imported in the place where they get recursively called.
Others experienced this, see [discord message](https://discord.com/channels/751419551132155925/751419551132155928/1067806341491609772)

### DIFF
--- a/extendr-api/tests/extendr_module_tests.rs
+++ b/extendr-api/tests/extendr_module_tests.rs
@@ -1,0 +1,30 @@
+//! Tests for [`extendr_module`] procedural macro.
+//!
+mod root {
+    use extendr_api::extendr;
+    use extendr_api::extendr_module;
+    use extendr_api::GetSexp;
+
+    mod nested_module {
+        use extendr_api::extendr;
+        use extendr_api::extendr_module;
+        use extendr_api::GetSexp;
+
+        #[extendr]
+        fn dummy() {}
+
+        extendr_module! {
+            mod nested_module;
+            fn dummy;
+        }
+    }
+
+    #[extendr]
+    fn hello_dummy() {}
+
+    extendr_module! {
+        mod top_level;
+        use nested_module;
+        fn hello_dummy;
+    }
+}
--- a/extendr-api/tests/extendr_module_tests.rs
+++ b/extendr-api/tests/extendr_module_tests.rs
@@ -1,12 +1,12 @@
 //! Tests for [`extendr_module`] procedural macro.
 //!
+use extendr_api::{extendr, extendr_module};
+
 mod root {
-    use extendr_api::extendr;
-    use extendr_api::extendr_module;
+    use super::*;
 
     mod nested_module {
-        use extendr_api::extendr;
-        use extendr_api::extendr_module;
+        use super::*;
 
         #[extendr]
         fn dummy() {}
@@ -23,6 +23,19 @@ mod root {
     extendr_module! {
         mod top_level;
         use nested_module;
+        use adjacent_module;
         fn hello_dummy;
+    }
+}
+
+mod adjacent_module {
+    use super::*;
+
+    #[extendr]
+    fn foo() {}
+
+    extendr_module! {
+        mod adjacent_module;
+        fn foo;
     }
 }
--- a/extendr-api/tests/extendr_module_tests.rs
+++ b/extendr-api/tests/extendr_module_tests.rs
@@ -3,12 +3,10 @@
 mod root {
     use extendr_api::extendr;
     use extendr_api::extendr_module;
-    use extendr_api::GetSexp;
 
     mod nested_module {
         use extendr_api::extendr;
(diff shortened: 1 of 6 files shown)


CGMossa:
I accidentally made a branch from a branch where I just was half done doing something.
Please accept the minor documentation edits. I was going to make a full PR of these things.

multimeric:
Can you write a test for this?

CGMossa:
I've no idea how to...

multimeric:
Maybe an [integration test](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html) that has multiple Rust modules in the same file, each with their own `extendr_module`?

CGMossa:
Writing a test revealed another issue.
Can we agree that this should compile?

``rust
mod top_level {
    use extendr_api::extendr_module;
    use extendr_api::GetSexp;

    // pub(in super) mod nested_module {
    pub mod nested_module {
        use extendr_api::extendr_module;
        use extendr_api::GetSexp;

        // pub(in super::super) fn dummy() {}
        pub fn dummy() {}

        extendr_module! {
            mod nested_module;
            fn dummy;
        }
    }

    extendr_module! {
        mod top_level;
        use nested_module;
    }
}
```

Where I'm getting issues...
I'll try to solve the need for adding `use extendr_api::GetSexp;`.

**Comment by CGMossa** [comment]
I get errors like

```
warning: build failed, waiting for other jobs to finish...
error[E0425]: cannot find function `meta__dummy2` in this scope
  --> extendr-api\tests\extendr_module_tests.rs:33:12
   |
33 |         fn dummy2;
   |            ^^^^^^ not found in this scope

error[E0425]: cannot find function `meta__dummy` in this scope
  --> extendr-api\tests\extendr_module_tests.rs:15:16
   |
15 |             fn dummy;
   |                ^^^^^ not found in this scope

\"[Finished running. Exit status: 0]\"
```

**Comment by CGMossa** [comment]
Perfect. I added a test and fixed another `extendr_module!` issue.

**Comment by multimeric** [comment]
Hmm it would be nice to be able to make assertions about the generated code. Can you call the `metadata` function in a test block and assert that it contains all the exported functions, or something?

**Comment by CGMossa** [comment]
I'll think about that. But no I don't think that this is easily possible.

I'm hoping this is a smal pr that we would get merged soon. And released. It is very annoying to users.

**Comment by multimeric** [comment]
I understand the desire to get it merged, but from my perspective as a reviewer I have no particular evidence that this works. Each module should define a `pub fn #module_metadata_name() -> extendr_api::metadata::Metadata`. Are you sure you can't just call that for each module to assert that it exists?

**Comment by CGMossa** [comment]
First, I did put the integration test that you wanted. And that one shows that this now works.
Second, the way this works, is you have a root module that calls its own meta data functions. Then based on the use statements it calls the other ExtendR modules' meta data.

So these functions are already being called.

I do valid your input and if I can make this more robust in the future, I'll make a PR.

**Comment by CGMossa** [comment]
This is a an issue with the current code-base that Andy had already warned us about in #404. Quite frankly it is up to people like me to clean this sort of thing up, but I've not done it before now.
So this addresses it part of the way, but generally everywhere we should be able to invoke the macros without
special importing statements.

**Comment by CGMossa** [comment]
Alright, I think we should merge this ASAP. I'm going to request rewiews now.

File: extendr-api/tests/extendr_module_tests.rs [hunk 1]
```diff
@@ -0,0 +1,28 @@
+//! Tests for [`extendr_module`] procedural macro.
+//!
+mod root {
```
Ilia-Kosenkov, line 3: Does this test guarantee that we solved the problem correctly? Can we verify the same behaviour in the integration tests, where we already use different modules (but not necessarily nested ones)?

CGMossa: I've expanded the test to include an adjacent module.
I don't know what else to do to prove that this is a solution.
The test fails on what was there before.
No other integration test have `extendr_module` for that matter.

Ilia-Kosenkov: I was suggesting putting this (or similar code) into `extendrtests` -- an R integration test package. There, if you can safely call one of the functions of an inner module, this guarantees we get correct metadata.

CGMossa: But I see you wrote R integration tests.

CGMossa:
Oh I see.
Anything in `crate/tests` are integration tests.
These things are not runtime things, but compile-time issues.

I'm not going to add a runtime test for the compile-time proc-macro in this
PR.

CGMossa:
Alright, so the `submodule` in `extendrtests` now no longer imports everything, thus it should have made an error just now. It hasn't. So there are now an **R** integration test as well.

Ilia-Kosenkov (approved):

PR #470, opened 2023-02-04
Added nonzero-types.

CGMossa:
There are non-zero types in [`std::num`](https://doc.rust-lang.org/std/num/index.html). I believe it would be nice if they were implemented as well.

This is a draft because I need some comments and reviews from ya'll before I can finalize this.


CGMossa:
The CI output doesn't make sense to me.
I'll remove the `i128` and `u128` first and see if that helps.

extendr-api/src/robj/tests.rs
```diff
@@ -100,13 +100,37 @@ fn test_try_from_robj() {
...
+        assert_eq!(<usize>::try_from(Robj::from(1)), Ok(1));
         assert_eq!(<i8>::try_from(Robj::from(1)), Ok(1));
         assert_eq!(<i16>::try_from(Robj::from(1)), Ok(1));
         assert_eq!(<i32>::try_from(Robj::from(1)), Ok(1));
         assert_eq!(<i64>::try_from(Robj::from(1)), Ok(1));
+        assert_eq!(<isize>::try_from(Robj::from(1)), Ok(1));
         assert_eq!(<f32>::try_from(Robj::from(1)), Ok(1.));
         assert_eq!(<f64>::try_from(Robj::from(1)), Ok(1.));
 
+        assert_eq!(<std::num::NonZeroU8>::try_from(Robj::from(1)).ok(), std::num::NonZeroU8::new(1));
```
multimeric, line 112: Could you refactor these into the new `rstest` cases?

extendr-api/src/robj/try_from_robj.rs
```diff
@@ -88,6 +88,30 @@ macro_rules! impl_try_from_scalar_real {
     };
 }
 
+macro_rules! impl_try_from_scalar_real_nonzero {
+    ($t:ty) => {
+        impl TryFrom<&Robj> for $t {
+            type Error = Error;
+
+            /// Convert a numeric object to a real value.
```
multimeric, line 96: More descriptive docs

multimeric (changes requested):

extendr-api/src/robj/tests.rs
CGMossa: I have made it use rstest, but right now I only see the need for one case. Is that good enough?

CGMossa:
@multimeric this one, I cannot request a review, and I did reply to messages.

multimeric:
Yep all good, I think because you already requested it you can't do so again.

extendr-api/src/robj/tests.rs
multimeric: No, I meant one case per assertion. You might have to do some magic with trait objects here.

CGMossa: Isn't that just obfuscating at this point? I'll give it a try though..

multimeric: It gives us more granular failures, it more accurately measures how many tests we have, and it lets you reduce some boilerplate (ideally).

CGMossa:
Alright. I've refactored the tests.
This has revealed to me that it is not possible to from from a `NonZeroType` to `Robj` as it stands in this PR.
This can be done in several ways, but until that's addressed, this PR shouldn't be merged IMO.

CGMossa:
Alright. I've dug into this and turns out `From<T> for Robj` is really only defined for `ToVectorValue`. It might seem like it is defined more broadly using `TryFrom<T> for Robj`, but here, some tricks of casting are involved. Strictly speaking, there aren't many one-to-one mappings.

Upon review, you might see there are some notices on lack of tests and conversions for `NonZeroIsize`. 
Turns out `isize` is neglected.
I will not address this in this PR; But if it is becomes relevant, under an issue of course, we can take a look.

extendr-api/src/robj/try_from_robj.rs
```diff
@@ -1,4 +1,4 @@
-//! Conversions to Robj
+//! Conversions to [`Robj`]
```
multimeric, line 1: Forgive my ignorance, but what do the square brackets do here?

multimeric (approved):

extendr-api/src/robj/try_from_robj.rs
```diff
@@ -88,6 +88,30 @@ macro_rules! impl_try_from_scalar_real {
     };
 }
 
+macro_rules! impl_try_from_scalar_real_nonzero {
+    ($t:ty) => {
+        impl TryFrom<&Robj> for $t {
+            type Error = Error;
+
+            /// Convert a numeric object without zeros, to a non-zero type.
```
multimeric, line 96:
```suggestion
            /// Convert a possibly non-zero R numeric object to a guaranteed non-zero Rust type.
```

extendr-api/src/robj/into_robj.rs
```diff
@@ -377,6 +377,27 @@ impl ToVectorValue for Option<bool> {
     }
 }
 
+macro_rules! impl_into_robj_nonzero {
+    ($type: ty) => {
+        impl From<$type> for Robj {
+            fn from(value: $type) -> Self {
```
multimeric, line 383:
```suggestion
            fn from(value: $type) -> Self {
                // A nonzero Rust type is automatically guaranteed  
                // to be a valid R type if the normal type can be converted to R
```

extendr-api/src/robj/tests.rs
```diff
@@ -92,21 +94,87 @@ fn test_from_robj() {
...
+#[case(1_i8)]
+#[case(1_i16)]
+#[case(1_i32)]
+#[case(1_i64)]
+// NOTE: `isize` is missing
+#[case(0_f32)]
+#[case(0_f64)]
+#[case(1_f32)]
+#[case(1_f64)]
+fn test_try_from_robj<T>(#[case] value: T)
```
multimeric, line 114:
```suggestion
fn test_round_trip<T>(#[case] value: T)
```

extendr-api/src/robj/tests.rs
```diff
@@ -92,21 +94,87 @@ fn test_from_robj() {
...
+where
+    Robj: From<T>,
+    T: TryFrom<Robj>,
+    T: ToVectorValue,
+    T: Copy,
+    T: std::fmt::Debug + PartialEq,
+    <T as std::convert::TryFrom<robj::Robj>>::Error: std::fmt::Debug,
+    <T as std::convert::TryFrom<robj::Robj>>::Error: PartialEq,
```
multimeric, line 122:
```suggestion
    <T as std::convert::TryFrom<robj::Robj>>::Error: std::fmt::Debug + PartialEq
```

extendr-api/src/robj/tests.rs
```diff
@@ -92,21 +94,87 @@ fn test_from_robj() {
     }
 }
 
-#[test]
-fn test_try_from_robj() {
+#[rstest]
```
multimeric, line 97: Can you put the tests inside a test module locked behind a config flag?

extendr-api/src/robj/tests.rs
```diff
@@ -92,21 +94,87 @@ fn test_from_robj() {
...
+// NOTE: `NonZeroU8` is missing
+#[case(std::num::NonZeroU16::new(1).unwrap())]
+#[case(std::num::NonZeroU32::new(1).unwrap())]
+#[case(std::num::NonZeroU64::new(1).unwrap())]
+#[case(std::num::NonZeroUsize::new(1).unwrap())]
+#[case(std::num::NonZeroI8::new(1).unwrap())]
+#[case(std::num::NonZeroI16::new(1).unwrap())]
+#[case(std::num::NonZeroI32::new(1).unwrap())]
+#[case(std::num::NonZeroI64::new(1).unwrap())]
```
multimeric, line 138: Consider just using `std::num::{Various Types}` to make this cleaner.

extendr-api/src/robj/tests.rs
```diff
@@ -92,21 +94,87 @@ fn test_from_robj() {
...
+// NOTE: `NonZeroIsize` is missing
+fn test_try_from_robj_nonzero<Value>(#[case] value: Value)
```
multimeric, line 140:
```suggestion
fn test_nonzero_round_trip_from_robj<Value>(#[case] value: Value)
```

extendr-api/src/robj/tests.rs
```diff
@@ -92,21 +94,87 @@ fn test_from_robj() {
...
+where
+    Robj: From<Value>,
+    Value: TryFrom<Robj>,
+    Value: Copy + std::fmt::Debug + PartialEq,
+    <Value as std::convert::TryFrom<robj::Robj>>::Error: std::fmt::Debug + PartialEq,
+{
+    test! {
+        assert_eq!(Value::try_from(dbg!(Robj::from(value))).ok(), Some(value));
```
multimeric, line 148: What's up with the `dbg!` here?

extendr-api/src/robj/tests.rs
```diff
@@ -92,21 +94,87 @@ fn test_from_robj() {
...
+#[case(std::num::NonZeroU8::new(1).unwrap())]
+#[case(std::num::NonZeroU8::new(1).unwrap())]
+#[case(std::num::NonZeroU16::new(1).unwrap())]
+#[case(std::num::NonZeroU32::new(1).unwrap())]
+#[case(std::num::NonZeroU64::new(1).unwrap())]
+#[case(std::num::NonZeroUsize::new(1).unwrap())]
+#[case(std::num::NonZeroI8::new(1).unwrap())]
+#[case(std::num::NonZeroI16::new(1).unwrap())]
+#[case(std::num::NonZeroI32::new(1).unwrap())]
+#[case(std::num::NonZeroI64::new(1).unwrap())]
```
multimeric, line 162: It's weird to not be using the values here, but I guess `rstest` doesn't support a turbofish invocation so maybe there is no better option.

multimeric (changes requested):
If possible, I would like to see this in a separate module from the normal `try_from_robj`, unless this would be particularly difficult. I think it would be easier to understand these conversions in isolation.

extendr-api/src/robj/tests.rs
CGMossa:
```suggestion
        assert_eq!(Value::try_from(Robj::from(value)).ok(), Some(value));
```

extendr-api/src/robj/tests.rs
CGMossa: These tests should already be behind a test-flag,
```rust

#[cfg(test)]
mod tests;
```

extendr-api/src/robj/try_from_robj.rs
CGMossa: I believe they make a link in the docs that leads to `Robj`. You can see if the link works
in vscode, as it will also become a Ctrl/Cmd+Click link there as well (with Rust Analyzer).

CGMossa:
@multimeric I should have now followed through with all the reviews.
Again, I apologise for missing the review you provided for this..

CGMossa:
Dear @Ilia-Kosenkov, I also believe this PR is ripe for merging. I'll fix the merge conflict, but I urge you to give it one last look and an approval.

PR #473, opened 2023-02-05
Special attr error handling

sorhawell:
POC

Use cases like this 
https://github.com/sorhawell/special_agent_attribute/blob/main/src/rust/src/lib.rs

### DIFF
--- a/extendr-api/src/robj/into_robj.rs
+++ b/extendr-api/src/robj/into_robj.rs
@@ -38,8 +38,14 @@ where
     T: Into<Robj>,
 {
     fn from(res: Result<T>) -> Self {
-        // Force a panic on error.
-        res.unwrap().into()
+        match res {
+            Ok(x) => x.into(),
+            Err(x) => x
+                .to_string()
+                .into_robj()
+                .set_attrib("extendr_err", true)
+                .unwrap(),
+        }
     }
 }
 
--- a/extendr-api/src/robj/into_robj.rs
+++ b/extendr-api/src/robj/into_robj.rs
@@ -70,6 +70,12 @@ impl From<Error> for Robj {
     }
 }
 
+impl From<Error> for String {
+    fn from(res: Error) -> Self {
+        res.to_string().into()
+    }
+}
+
 // // ... and Err does not have to implement Display
 // impl<T, E> From<std::result::Result<T, E>> for Robj
 // where
--- a/extendr-api/src/robj/into_robj.rs
+++ b/extendr-api/src/robj/into_robj.rs
@@ -33,11 +33,29 @@ impl From<()> for Robj {
 ///     assert_eq!(r!(my_func()), r!(1.0));
 /// }
 /// ```
-impl<T> From<Result<T>> for Robj
+// impl<T> From<Result<T>> for Robj
+// where
+//     T: Into<Robj>,
+// {
+//     fn from(res: Result<T>) -> Self {
+//         match res {
+//             Ok(x) => x.into(),
+//             Err(x) => x
+//                 .to_string()
+//                 .into_robj()
+//                 .set_attrib("extendr_err", true)
+//                 .unwrap(),
+//         }
+//     }
+// }
+
+// it does not have to be extendr result
+impl<T, E> From<std::result::Result<T, E>> for Robj
 where
     T: Into<Robj>,
+    E: std::fmt::Display,
 {
-    fn from(res: Result<T>) -> Self {
+    fn from(res: std::result::Result<T, E>) -> Self {
         match res {
             Ok(x) => x.into(),
             Err(x) => x
@@ -48,6 +66,20 @@ where
         }
     }
 }
+
+// // ... and Err does not have to implement Display
+// impl<T, E> From<std::result::Result<T, E>> for Robj
+// where
+//     T: Into<Robj>,
+//     E: Into<Robj>,
+// {
+//     fn from(res: std::result::Result<T, E>) -> Self {
+//         match res {
+//             Ok(x) => x.into(),
+//             Err(x) => x.into_robj().set_attrib("extendr_err", true).unwrap(),
+//         }
+//     }
+// }
 
 /// Convert an Robj reference into a borrowed Robj.
 impl From<&Robj> for Robj {
--- a/extendr-api/src/robj/into_robj.rs
+++ b/extendr-api/src/robj/into_robj.rs
@@ -50,20 +50,23 @@ impl From<()> for Robj {
 // }
 
 // it does not have to be extendr result
+
 impl<T, E> From<std::result::Result<T, E>> for Robj
 where
     T: Into<Robj>,
-    E: std::fmt::Display,
+    E: Into<Robj>,
 {
     fn from(res: std::result::Result<T, E>) -> Self {
         match res {
             Ok(x) => x.into(),
-            Err(x) => x
-                .to_string()
-                .into_robj()
-                .set_attrib("extendr_err", true)
-                .unwrap(),
+            Err(x) => x.into().set_attrib("extendr_err", true).unwrap(),
         }
+    }
+}
+
+impl From<Error> for Robj {
+    fn from(res: Error) -> Self {
+        res.to_string().into()
     }
 }
 


extendr-api/src/robj/into_robj.rs
```diff
@@ -33,16 +33,63 @@ impl From<()> for Robj {
...
     T: Into<Robj>,
+    E: Into<Robj>,
 {
-    fn from(res: Result<T>) -> Self {
-        // Force a panic on error.
-        res.unwrap().into()
+    fn from(res: std::result::Result<T, E>) -> Self {
+        match res {
+            Ok(x) => x.into(),
+            Err(x) => x.into().set_attrib("extendr_err", true).unwrap(),
```
multimeric, line 62: I actually think we should handle the attribute setting error too, I know it will be a tad annoying to implement.

multimeric:
So can you explain, for posterity, why using an attribute is the best way to handle this, rather than say returning a `errorCondition`? See: https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/conditions#:~:text=see%20details%20below.-,Details,-The%20condition%20system.

multimeric:
Is the idea that we will add some error handling on the R wrapper side to either throw the error or else return it?

sorhawell:
> So can you explain, for posterity, why using an attribute is the best way to handle this, rather than say returning a `errorCondition`? See: https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/conditions#:~:text=see%20details%20below.-,Details,-The%20condition%20system.

*"This is not the greatest song in the world, this is just an Attribute 🤣"*

Attribute is not the best way, it is just one way. So far this PR do not address (yet) how to switch between different error handling behaviors. I will try to add compiler feature, to let the user keep status quo as default or choose any of 3-4 different behavoirs.

I guess conditions are great if every error is a string and is highly likely is going to be raised immediately. Attribute-error supports non string errors and the further handling if not raising requires a less boilerplate code.

sorhawell:
> Is the idea that we will add some error handling on the R wrapper side to either throw the error or else return it?
Yes, that should be an option. Others might prefer to handle manually after the wrapper.

extendr-api/src/robj/into_robj.rs
sorhawell: If some Robj actually would fail on set_attrib() instead of panic! this code could another attribute error stating an "internal error". Maybe it is acceptable to panic on internal errors. I will try to look into when set_attrib could fail.

sorhawell:
@multimeric  many thanks for feedback!

eitsupi:
Perhaps we could do the error handling on the R side on the R file side that the (r)extendr generates?

multimeric:
> I guess conditions are great if every error is a string and is highly likely is going to be raised immediately. Attribute-error supports non string errors and the further handling if not raising requires a less boilerplate code.

Conditions can have other data though, the only requirement is that they *at least* have a string message. e.g. 

```r
> x = errorCondition(message="String expected", value = 9, class="ExpectedString")
> str(x)
List of 3
 $ message: chr "String expected"
 $ call   : NULL
 $ value  : num 9
 - attr(*, "class")= chr [1:3] "ExpectedString" "error" "condition"
```

This is great because it works natively with R's error handling. If we decide to throw this using `stop(x)`, the end user can explicitly catch this class of error using `tryCatch(call_rust_function(), ExpectedString=<handle the error>)`.

sorhawell:
@multimeric what is the most efficient way currently to creat  a condition with extendr?
I'm fiddling with something like this, but do I have to call R! ?
```R
R!("simpleError(message = 'extendr_Error', class = {{robj}})")
```

multimeric:
The most elegant syntax is probably `R!("errorCondition").call(pairlist!(message="extendr_Error", class=robj)).unwrap()`. For more info see: https://extendr.github.io/extendr/extendr_api/prelude/trait.Operators.html#method.call.

However this is definitely doing everything via the R APIs. The actual R function is dead simple:

```r
> errorCondition
function (message, ..., class = NULL, call = NULL)
structure(list(message = as.character(message), call = call,
    ...), class = c(class, "error", "condition"))
```

I wouldn't at all be surprised if `list!(message="extendr_Error").set_class([my_class, "error", "condition"]).unwrap()` was faster.

extendr-api/src/robj/into_robj.rs
sorhawell: I have tried to address set_attrib. I have been thinking about handling errors from set_attrib() list!() set_class(). If e.g. not possible to set attribute on an error or class on a condtion, then it is possible to make a new simpler Robj describing this as an internal error. However, if it is still not possible to create attributes or conditions (e.g. due to mem alloc), there is nothing left to do than panicking.

sorhawell: https://github.com/extendr/extendr/blob/aecff2890b4090c70fa80c54d870716d8f1e0537/extendr-api/src/robj/into_robj.rs#L91

multimeric:
Would be great to benchmark all of those options, too, in case we can eliminate one option or just provide meaningful docs for end users.

sorhawell:
> Would be great to benchmark all of those options, too, in case we can eliminate one option or just provide meaningful docs for end users.

True. Howver, this current feature-flag implementation is a bit annoying to benchmark, because each behavior needs to be compiled first. I will look into later with macro. Also this PR have not address yet how to raise errors via auto-generated wrappers yet. I shouldn't be any wrapper must check for conditions/attributes etc.

sorhawell:
> Would be great to benchmark all of those options, too, in case we can eliminate one option or just provide meaningful docs for end users.

I actually really like to see some new error handling options in the short run. It is unrealistic for me to rewrite the extendr macros to support multiple errror handling behaviors. So I performed a manual benchmark as a start.


[I have performed a handheld benchmark](https://github.com/sorhawell/special_agent_attribute/blob/main/R/run_bench.R) because each run requires recompilation. The story of results are is that all behaviors (panic, condition, attributem list) are really fast from 2-4 us per function call. That is up 500_000 calls/s ! 
returning ok-values in a list takes 4us and other methods about 2us. However for returning err values all behavoirs take 4us, and condition is slightly slower and attribute is slightly faster. Did not test panic as it will be very slow because printing to stdout cannot be disabled.

It is not enough to return the result type. It must also be handled on R side. The handling is slower because it is not in rust. call+unwrap_list takes ~7us, call+unwrap_attr takes ~6us and call+unwrap_condition ~9us.

In conclusion:
- all behaviors are close to equally fast.
- Returning list is only significant slower if returning ok values which are not handled in R. This happy-path use case is not relevant anyways, see last point.
- When returning err values attribute and condition have similar speed. After all  attributes and conditions are not nore lightweight than a list.
-  And lastly because a potential error must be checked for and handled all behaviors are similarly fast. Even though condition and attribute happy-path  is slightly faster, the checking makes them on par with list for both or and err values.

sorhawell:
I think this PR has some good selling points :) 
 - Unless the extendr user activates a feature, everything is exactly as normal.
 - The new error handling behaviors are actually not that slow and do not leak.
 
what do you think?

Ilia-Kosenkov:
While I like the concept of this PR, I am not entirely convinced it should be based on features. A feature is a compile-time constant, so if you choose one, you lose all other options, am I right? (otherwise you will have multiple trait implementations). Was there a discussion or comparison of feature-gates to attribute annotation, i.e. allowing different functions within the same crate to use different error handling?

sorhawell:
> Was there a discussion or comparison of feature-gates to attribute annotation, i.e. allowing different functions within the same crate to use different error handling?

I agree it is fair question, to wonder if there could be something better than feature gated behavior. I just don't know anyway of implementing that in a reasonable time frame and neither has there has been any progress on error handling for 1.5 years. Status quo is not good.

I like to invite someone to make a better PR in this month. If not, we allow a feature gated error handling until some time in future, where someone will write the perfect error handling system, that wins the consensus vote.

sorhawell:
oh I forgot to answer the question for current PR
If the user wants to panic, they should just use unwrap or panic explicitly. That is always possible.
Whether the user wants to use attribute-, condition- or list-encoding of a Result does not matter for speed or possibilities. Essentially a list, a attribute-set and a conditions are all kinda lists, there is no difference. I don't think a user would need to pick multiple error handling behaviors.

multimeric (commented):
This PR is very well done and documented. There are however some issues we need to think about. 

Firstly, I don't think features are the right mechanism for this, since they're designed to be additive by design. I wouldn't mind using arbitrary cargo config (e.g. [`--cfg 'result=condition'`](https://doc.rust-lang.org/rustc/command-line-arguments.html#--cfg-configure-the-compilation-environment)). I'm not 100% convinced I agree with Ilia that it should be a per-extendr-block option, because having a library whose error behaviour is inconsistent between functions would be quite strange.

Another point is that, because the `list` and `condition` option are so similar in behaviour right now, we should have `condition` be automatically thrown in R rather than returned. This is actually what I would want as a user, and it meaningfully distinguishes this mode from the `list` result.

Something else that I don't quite understand is how we are handling direct `panic!` calls in Rust. We handle the panic right, so will this PR cause even those errors to be safely handled?

sorhawell:
@multimeric good comments, I will try to improve the PR in that direction.

multimeric:
Hi @sorhawell, if you don't have time to finish this, I can fork it and proceed from there?

sorhawell:
@multimeric  Yes admittedly I will not continue on this PR before mid summer. Thank you.

>  I wouldn't mind using arbitrary cargo config (e.g. [--cfg 'result=condition'](https://doc.rust-lang.org/rustc/command-line-arguments.html#--cfg-configure-the-compilation-environment)).

I agree. This feature flagging is not the right way.

> we should have condition be automatically thrown in R rather than returned. This is actually what I would want as a user, and it meaningfully distinguishes this mode from the list result.

That makes sense to me. In r-polars it happens often now and arg goes back and forth between R an rust 2-3 times, so the list error makes sense. For most project a condition error being immediately thrown likely makes the most sense.

> Something else that I don't quite understand is how we are handling direct panic! calls in Rust. 

As I understand it, when panic! rust-extendr will unwind all owned mem allocations, write error msg to std err (not R terminal) , hence tricky to catch err, but ok performance wise and no mem bleeding. The baddie is the throw_r_error extendr-function, which leaves the rust context behind without unwinding/dropping objects.

> We handle the panic right, so will this PR cause even those errors to be safely handled?

This PR does not change the option to panic! from rust, but it adds an alternative to panic! . This PR change the default conversion of Result<T,E> into Robj. Instead of .unwrap() it returns the result as a list, condition  or attribute. Users can still use unwrap / panic! as much as they see fit.

multimeric (commented):
I forgot to submit this review earlier.

sorhawell:
I have picked this up this PR again because I have a little time this week.

@multimeric I think now compiler flag is neither the right way as setting behaviours not using features is also discouraged. It also makes it very clungy to set compiler flag for extendr-api only in a Cargo.toml.

Instead I believe the right way is to let incompatible features have precedence, such that if multiple are set simultanously just one of them takes effect. [This is officially recommended](https://doc.rust-lang.org/cargo/reference/features-examples.html#feature-precedence):


todo! before ready again:
- [ ] update readme
- [ ] update features
- [ ] update docs

eitsupi:
@extendr Could you look at this again?

.vscode/settings.json
eitsupi, line 1: I think this file should be removed.

extendr-api/Cargo.toml
```diff
@@ -32,7 +32,12 @@ serde = { version = "1.0", features = ["derive"], optional = true }
 rstest = "0.16.0"
 
 [features]
-default = []
+
+# Features to modify behaviour of returning Result<T,E> from extendr to R, instead of unwrap-throw_r_error().
+# Add any new feature below here and likely implement in src/robj/into_robj.rs
```
Ilia-Kosenkov, line 37: Not convinced `Cargo.toml` is the place for such comment

```suggestion
# Add new features below
```

extendr-api/src/robj/into_robj.rs
```diff
@@ -17,18 +17,30 @@ pub(crate) fn str_to_character(s: &str) -> SEXP {
...
     }
 }
 
-/// Convert a `Result` to an `Robj`. This is used to allow
-/// functions to use the `?` operator and return [`Result<T>`].
+/// Convert a Result to an Robj.
 ///
 /// Panics if there is an error.
+///
+/// To use the ?-operator the extendr-function must return an extendr_api::result::Result<T> type or in general the `std::result::Result<T,E>` enum.
```
Ilia-Kosenkov, line 32:
```suggestion
/// To use the ?-operator, an extendr-function must return either extendr_api::result::Result<T> or `std::result::Result<T,E>`.
```

extendr-api/src/robj/into_robj.rs
```diff
@@ -39,16 +51,157 @@ impl From<()> for Robj {
...
         res.unwrap().into()
     }
 }
 
+/// Convert a Result to an Robj either an Ok value or the Err value wrapped in a
+/// error condition. This is used to allow functions to use the ? operator
+/// and return [Result<T>] without panicking on an Err. T must impl IntoRobj.
+///
+/// Returns ok-value as is. Return err wrapped in a R error condition. The err is placed in
+/// $value. The condition messeage is simply 'extendr_arr'
```
Ilia-Kosenkov, line 70:
```suggestion
/// Returns Ok value as is. Returns Err wrapped in an R error condition. The Err is placed in
/// $value field of the condition, and its message is set to 'extendr_err'
```

extendr-api/src/robj/into_robj.rs
```diff
@@ -39,16 +51,157 @@ impl From<()> for Robj {
...
+///         r!(my_ok_f()),
+///         R!("x = list(ok=123.123, err=NULL)
+///             class(x)='extendr_result'
+///             x"
+///         ).unwrap()
+///     );
+/// }
+///
+/// ```
+#[cfg(feature = "result_list")] //write this feature and not any result_feature with presecence
```
Ilia-Kosenkov, line 154: Same here, please clarify the comment

extendr-api/src/robj/into_robj.rs
```diff
@@ -39,16 +51,157 @@ impl From<()> for Robj {
...
+//     T: Into<Robj>,
+//     E: Into<Robj>,
+// {
+//     fn from(res: std::result::Result<T, E>) -> Self {
+//         match res {
+//             Ok(x) => x.into(),
+//             Err(x) => x.into_robj().set_attrib("extendr_err", true).unwrap(),
+//         }
+//     }
+// }
```
Ilia-Kosenkov, line 203: Is this code needed here? If you do not plan to uncomment it, it should be removed.

extendr-api/src/robj/into_robj.rs
```diff
@@ -570,38 +740,28 @@ macro_rules! impl_from_as_iterator {
...
+impl_from_iter! {[T; 12]}
+impl_from_iter! {[T; 13]}
+impl_from_iter! {[T; 14]}
+impl_from_iter! {[T; 15]}
+impl_from_iter! {[T; 16]}
+impl_from_iter! {[T; 17]}
+impl_from_iter! {[T; 18]}
+impl_from_iter! {[T; 19]}
+impl_from_iter! {Vec<T>}
+impl_from_iter! {&Vec<T>}
```
Ilia-Kosenkov, line 764: How is this relevant to error handling?

Ilia-Kosenkov (commented):
Some documentation/comments editing here and there.

Ilia-Kosenkov:
Don't forget to run `cargo fmt` before pushing

sorhawell:
> Don't forget to run `cargo fmt` before pushing

I will give the pr some more work next week.

extendr-api/src/robj/into_robj.rs
sorhawell: Follow-up: 
If using feature `result_condition` then the error is tagged as @multimeric  suggested 
https://github.com/sorhawell/extendr/blob/4a7a4dd623a0446ee2ef2d8c6684f6b006dbc323/extendr-api/src/robj/into_robj.rs#L103

if using feature `result_list` then the `list(ok=?,err=?)` is tagged with class `extendr_result`
https://github.com/sorhawell/extendr/blob/4a7a4dd623a0446ee2ef2d8c6684f6b006dbc323/extendr-api/src/robj/into_robj.rs#L148

... and default with neither feature will just panic! on Err

extendr-api/src/robj/into_robj.rs
```diff
@@ -39,16 +39,157 @@ impl From<()> for Robj {
...
+where
+    T: Into<Robj>,
+    E: Into<Robj>,
+{
+    fn from(res: std::result::Result<T, E>) -> Self {
+        match res {
+            Ok(x) => x.into(),
+            Err(x) => {
+                let robj: Robj = x.into();
+                List::from_names_and_values(
```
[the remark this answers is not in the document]
sorhawell: agreed, corrected

extendr-api/src/robj/into_robj.rs
```diff
@@ -39,16 +39,157 @@ impl From<()> for Robj {
...
+/// ```
+#[cfg(feature = "result_list")]
+impl<T, E> From<std::result::Result<T, E>> for Robj
+where
+    T: Into<Robj>,
+    E: Into<Robj>,
+{
+    fn from(res: std::result::Result<T, E>) -> Self {
+        match res {
+            Ok(x) => List::from_names_and_values(&["ok", "err"], &[x.into(), NULL.into()]),
```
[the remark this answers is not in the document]
sorhawell: agreed, corrected

extendr-api/src/robj/into_robj.rs
sorhawell: I think the diff viewer is off here. I will try to add a comment to hack it back into order

sorhawell: fixed https://github.com/extendr/extendr/pull/473/commits/9923d2464eb2d32636826d1d51de32e3b856ccd7

extendr-api/src/robj/into_robj.rs
sorhawell: removed

extendr-api/src/robj/into_robj.rs
sorhawell: comment dropped

The comment was an instruction for how to add future encoding features. I don't think it was necessary.

extendr-api/src/robj/into_robj.rs
```diff
@@ -39,16 +51,157 @@ impl From<()> for Robj {
...
+///         r!(my_err_f()),
+///         R!(
+/// "structure(list(message = 'extendr_err',
+/// value = 42.0), class = c('extendr_error', 'error', 'condition'))"
+///         ).unwrap()
+///     );
+/// }
+///
+/// ```
+#[cfg(all(feature = "result_condition", not(feature = "result_list")))] //write this feature, right any feature with precedence
```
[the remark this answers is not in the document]
sorhawell: comment dropped 

it was about how to add more encoding features in future. Not necessary I think

sorhawell:
@Ilia-Kosenkov  all current comments / request has been addressed.

Ilia-Kosenkov:
Would you also handle wrapper generation afterward?

sorhawell:
Currently the default is to panic! on an Err, so no wrapper is needed there.
The user would be expected if opting-in to feature `result_list` or `result_condition` to unwrap errors explicitly on R-side.

I have not seen yet the code that produces wrappers. So not entirely sure how it works.

In future there could be something like `#[extendr(unwrap_err = 'true')]` and/or a feature flag like `never_unwrap` which toogles result-unwrapping on/off.

sorhawell:
User 1: Use extendr to develop a package. This user likely does not have the naked rust wrapper as public function and don't mind unwrapping the Result. Potentially the definition of returning a `Result` from rust to R could literally mean to do just that. Not to raise an error.

User 1a
In fact sometimes the user would like to work with the Result *as is* before unwrapping. Perhaps the user would like to pass the R-side-`Result` back to a rust function that takes a R-side-`Result` as input. For polars this needs to happen quite alot, because it implements a nested syntax. If all `Result` was automatic thrown then all internal rust functions would need to be wrapped in tryCatch everywhere to restore the `Result` once again on R side. It is really a pleasure to write code on R side with no error throwing functions at all, just clear `Results` everywhere.

user 1b
Really like throwing catching errors on R side and want to do just that.

User 2: Beginner who wants to do to do some  inline compilation of rust functions in R session. Probably always prefer 
any error immediately thrown.

sorhawell:
I think we agree on it should be an optional behavior and I don't disagree throwing should be the default as long as it is easy to revert.


I speculate that we can modify the extendr proc macro macro to remember these choices for each function / impl.

Since [`make_r_wrappers` is defined in extendr](https://github.com/extendr/extendr/blob/836b0cc0ec2ce6ae35c106c2ac4338426e46be35/extendr-api/src/metadata.rs#L335) and not rextendr the generated extendr-wrappers.R file is always generated accordingly to rust side code

sorhawell:
```rust
#[derive(Debug, PartialEq)]
pub struct Func {
    pub doc: &'static str,
    pub rust_name: &'static str,
    pub mod_name: &'static str,
    pub r_name: &'static str,
    pub args: Vec<Arg>,
    pub return_type: &'static str,
    pub func_ptr: *const u8,
    pub hidden: bool,
    pub auto_unwrap_function: bool // add this field to Func
}
```

then`make_r_wrappers` will deal with that accordingly. How to auto-unwrap is also dependent on choice of feature `result_list` `result_condition`

Ilia-Kosenkov (approved):
I forgot about this one. Let's deal with R-side wrappers elsewhere. LGTM. Can I merge this as is or do you need to fix anything?

Issue #425, opened 2022-07-17
Switch from actions-rs to dtolnay/rust-toolchain?


yutannihilation:
c.f https://www.reddit.com/r/rust/comments/vyx4oj/actionsrs_organization_became_unmaintained/

andy-thomason:
Seems like a sensible idea.

yutannihilation:
It seems we'll have no choice but to switch.

https://github.com/extendr/extendr/issues/452

CGMossa:
@yutannihilation Seems like a sensible solution. If you need support on this @yutannihilation let me know. Unfortunately I don't have the CI-skills to carry this on my own.

yutannihilation:
Thanks, this should be easy. I will create pull requests.

yutannihilation:
Hmm, it turned out it's not very easy. I'll investigate the failures...

yutannihilation:
It seems I hit this problem.

https://github.com/KyleMayes/clang-sys/issues/154

yutannihilation:
Done!

PR #477, opened 2023-02-06
Use dtolnay/rust-toolchain

yutannihilation:
Address https://github.com/extendr/extendr/issues/425.

To specify the toolchain version, `dtolnay/rust-toolchain` accepts two ways: (1) after `@` (e.g. `dtolnay/rust-toolchain@stable`) or (2) `toolchain` component. We need to use the latter way because we want to provide the version from the config matrix. Because of this design decision, ` dtolnay/rust-toolchain` doesn't have the version of the action itself, so we use `@master`. [The README](https://github.com/dtolnay/rust-toolchain#inputs) says:

> When passing an explicit toolchain as an input instead of @rev, you'll want to use "dtolnay/rust-toolchain@master" as the revision of the action. 

One more difference is that there's no `default` input. `dtolnay/rust-toolchain` always set the chosen toolchain as default ([code](https://github.com/dtolnay/rust-toolchain/blob/9cd00a88a73addc8617065438eff914dd08d0955/action.yml#L75-L76)).

I also include the change of `actions/checkout` to address the Node 12 deprecation warning on v2 (for more details, please see [the official announcement](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/) and [the issue filed on the original repo](https://github.com/actions/checkout/issues/959))

### DIFF
--- a/extendr-api/src/na.rs
+++ b/extendr-api/src/na.rs
@@ -8,7 +8,7 @@ use std::alloc::{self, Layout};
 lazy_static! {
     static ref EXTENDR_NA_STRING: &'static str = unsafe {
         // Layout::array() can fail when the size exceeds `isize::MAX`, but we
-        // only need 2 hwre, so it's safe to unwrap().
+        // only need 2 here, so it's safe to unwrap().
         let layout = Layout::array::<u8>(2).unwrap();
 
         // We allocate and never free it because we need this pointer to be
--- a/extendr-api/src/na.rs
+++ b/extendr-api/src/na.rs
@@ -7,8 +7,14 @@ use std::alloc::{self, Layout};
 // ourselves.
 lazy_static! {
     static ref EXTENDR_NA_STRING: &'static str = unsafe {
+        // Layout::array() can fail when the size exceeds `isize::MAX`, but we
+        // only need 2 hwre, so it's safe to unwrap().
         let layout = Layout::array::<u8>(2).unwrap();
+
+        // We allocate and never free it because we need this pointer to be
+        // alive until the program ends.
         let ptr = alloc::alloc(layout);
+
         let v: &mut [u8] = std::slice::from_raw_parts_mut(ptr, 2);
         v[0] = b'N';
         v[1] = b'A';
@@ -72,6 +78,6 @@ impl CanBeNA for &str {
     }
 
     fn na() -> Self {
-        *EXTENDR_NA_STRING
+        &EXTENDR_NA_STRING
     }
 }
--- a/extendr-api/src/na.rs
+++ b/extendr-api/src/na.rs
@@ -1,4 +1,22 @@
+use lazy_static::lazy_static;
 use libR_sys::{R_IsNA, R_NaReal};
+use std::alloc::{self, Layout};
+
+// To make sure this "NA" is allocated at a different place than any other "NA"
+// strings (so that it can be used as a sentinel value), we allocate it by
+// ourselves.
+lazy_static! {
+    static ref EXTENDR_NA_STRING: &'static str = unsafe {
+        let layout = Layout::array::<u8>(2).unwrap();
+        let ptr = alloc::alloc(layout);
+        let v: &mut [u8] = std::slice::from_raw_parts_mut(ptr, 2);
+        v[0] = b'N';
+        v[1] = b'A';
+
+        std::str::from_utf8_unchecked(v)
+    };
+}
+
 /// Return true if this primitive is NA.
 pub trait CanBeNA {
     fn is_na(&self) -> bool;
@@ -54,6 +72,6 @@ impl CanBeNA for &str {
     }
 
     fn na() -> Self {
-        unsafe { std::str::from_utf8_unchecked(&[b'N', b'A']) }
+        *EXTENDR_NA_STRING
     }
 }
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-20.04
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@stable
         with:
           toolchain: stable
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -12,7 +12,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: stable
       - name: Docs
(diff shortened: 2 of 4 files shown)


yutannihilation:
I have no idea why this fails reliably on some OSes, and why it says `(left != right)` where the left and the right look the same to my eyes.

```
  thread 'tests::test_na_str' panicked at 'assertion failed: `(left != right)`
    left: `0x10e6bb023`,
   right: `0x10e6bb023`', extendr-api/src/lib.rs:889:9
```

This line:

https://github.com/extendr/extendr/blob/974cddc4ee5a4c3ab10e8bf85772b74c0a182f32/extendr-api/src/lib.rs#L889

yutannihilation:
> why it says `(left != right)` where the left and the right look the same to my eyes.

It seems I was confused. I thought it was `assert_eq!`, not `assert_ne!`. Yet, I have no idea what's happening here.

yutannihilation:
Looking back when this assertion was introduced, it seems the intention was this. 

> a special null string with value "NA" but guaranteed to have a different address from the regular "NA" string.

https://github.com/extendr/extendr/pull/116

But, I'm now wondering if this is really "guaranteed." Isn't it possible that the compiler is smart enough to notice that the binary array of `[b'N', b'A']` is the same one as the string "NA"?

https://github.com/extendr/extendr/blob/974cddc4ee5a4c3ab10e8bf85772b74c0a182f32/extendr-api/src/na.rs#L57

(But, even if this is true, I have no idea why it fails only when we use the dtolnay/rust-toolchain action)

yutannihilation:
Sure, let's see what happens.

Ilia-Kosenkov:
Interesting, this one seems to be random 
>    test robj::tests::test_to_robj ... ok
  thread 'ownership::test::basic_test' panicked at 'assertion failed: `(left == right)`
    left: `1`,
   right: `0`', extendr-api/src/ownership.rs:220:17
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
  test ownership::test::basic_test ... FAILED
  test robj::tests::test_try_from_robj has been running for over 60 seconds
  test tests::big_r_macro_works has been running for over 60 seconds

A wild guess -- timeouts come from synchronization.

yutannihilation:
Yeah, the error happens here. This probably indicates there's some object that is in use, and probably it was the case, considering the test runs infinitely until I manually canceled.

https://github.com/extendr/extendr/blob/974cddc4ee5a4c3ab10e8bf85772b74c0a182f32/extendr-api/src/ownership.rs#L217-L224

 This should not happen, but I don't think this is related to this change.

yutannihilation:
Ah, but I think this is a wrong assumption.

``` rust
if elt == R_NilValue { 
    assert_eq!(self.ref_count(elt), 0); 
```

In `Ownership.preservation`, `R_NilValue` can mean

1. empty
2. manually added value

and in case 2, the reference count is not 0.

This can happen as `protect()` doesn't have any logic to prevent to be applied on `R_NilValue`.

https://github.com/extendr/extendr/blob/974cddc4ee5a4c3ab10e8bf85772b74c0a182f32/extendr-api/src/ownership.rs#L71-L107

Anyway, let's ignore this for now.

multimeric:
Do you need any help here? I want to get this merged asap.

yutannihilation:
Thanks, I really need someone's help to figure out what's happening, because I'm stuck here. Are there any reason to get this merged in a hurry? I'm afraid it takes some time for us to get confident enough to merge this.

multimeric:
>  Are there any reason to get this merged in a hurry?

Ah, well just because the CI is mostly broken at the moment, mostly caused by the old `setup-r` action. e.g. https://github.com/extendr/extendr/actions/runs/4104087907/jobs/7079108917

yutannihilation:
Hmm, then, this pull request doesn't help as this contains no change related to `setup-r`. What do you mean by "old"?

multimeric:
Ah my bad, I thought it was failing to set up the Rust toolchain but it was failing to set up R. Ignore me.

yutannihilation:
Ah, got it. No worries. I hope time will solve it...

yutannihilation:
Confirmed `CARGO_INCREMENTAL` actually matters.

diff: https://github.com/extendr/extendr/compare/master...yutannihilation:tmp
result: https://github.com/yutannihilation/extendr/actions/runs/4129557719/jobs/7135285809

This envvar is one of the subtle differences between `dtolnay/rust-toolchain` and `actions-rs/toolchain` ([code](https://github.com/dtolnay/rust-toolchain/blob/9cd00a88a73addc8617065438eff914dd08d0955/action.yml#L86-L89)). The above attempt shows only the cases with `dtolnay/rust-toolchain`, but `actions-rs/toolchain` also fails if the incremental compilation is disabled ([ref](https://github.com/yutannihilation/extendr/actions/runs/4125640513)).

While I haven't succeeded to reproduce this minimally so far (here's [my attempt](https://github.com/yutannihilation/compare-na-with-na-by-ptr)), my take is that the current implementation of `CanBeNA` for `&str` is broken. The intention of the following lines is that `&[b'N', b'A']` is on a different address from the regular "NA" string, but I'm afraid this is not guaranteed.

https://github.com/extendr/extendr/blob/974cddc4ee5a4c3ab10e8bf85772b74c0a182f32/extendr-api/src/na.rs#L50-L59

So, what should we do? I have no idea. I'd like to remove `CanBeNA` for `&str` at all, but probably this is because I'm biased (I generally don't like the idea of handling missing values on Rust types). We can anyway merge this with these failing tests disabled and think about the issue later.

yutannihilation:
The failure on Windows with R-devel seems irrelevant to this change. Maybe this is what we have to deal with on rextendr?

https://github.com/extendr/extendr/actions/runs/4209266707/jobs/7306115505

extendr-api/src/na.rs
```diff
@@ -1,4 +1,28 @@
...
 use libR_sys::{R_IsNA, R_NaReal};
+use std::alloc::{self, Layout};
+
+// To make sure this "NA" is allocated at a different place than any other "NA"
+// strings (so that it can be used as a sentinel value), we allocate it by
+// ourselves.
+lazy_static! {
+    static ref EXTENDR_NA_STRING: &'static str = unsafe {
+        // Layout::array() can fail when the size exceeds `isize::MAX`, but we
+        // only need 2 hwre, so it's safe to unwrap().
```
Ilia-Kosenkov, line 11:
```suggestion
        // only need 2 here, so it's safe to unwrap().
```

Ilia-Kosenkov (changes requested):
Looks good! There is a tiny typo in the comments.

extendr-api/src/na.rs
yutannihilation: Aw, thanks for catching...

Ilia-Kosenkov (approved):

yutannihilation:
Thanks. I'm merging this and file a new issue for `pretty_rel_path()` failure.

PR #488, opened 2023-02-24
Add `use_rng` option to `extendr`

CGMossa:
Restarted #476 PR as a rebase went really badly.


extendr-api/Cargo.toml
```diff
@@ -15,12 +15,12 @@ license = "MIT"
 repository = "https://github.com/extendr/extendr"
 
 [dependencies]
-libR-sys = "0.3.0"
-extendr-macros = { path = "../extendr-macros", version = "0.3.1" }
-extendr-engine = { path = "../extendr-engine", version = "0.3.1" }
-ndarray = { version = "0.15.3", optional = true }
+libR-sys = { workspace = true }
```
multimeric, line 18: Can you explain what you're doing here with the internal dependencies?

extendr-api/Cargo.toml
```diff
@@ -15,12 +15,12 @@ license = "MIT"
...
+extendr-macros = { path = "../extendr-macros" }
+extendr-engine = { path = "../extendr-engine" }
 lazy_static = "1.4"
 paste = "1.0.5"
+ndarray = { version = "0.15.3", optional = true }
```
multimeric, line 23: Why an ndarray change?

multimeric:
I think this PR is good, but we maybe need to think about whether it needs to live in the `extendr` macro or not. It seems like RNG has a "scope" (I don't really understand why, but I'm sure you do). So maybe `with_rng!{}` could be a macro that the user can put inside their extendr function? I'm a bit wary of overloading the `extendr` macro with things that don't relate to the wrapped function itself or the way in which it is converted to R.

extendr-api/Cargo.toml
CGMossa: Accidental. I think I've cleaned it up now.

extendr-api/Cargo.toml
CGMossa: I need to use `libR-sys` in the `extendr-macro`. There is no need to create version confusion, so instead I make it a workspace dependency and use that all around when needed. This is how that is done.

CGMossa:
I think it should live in `#[extendr]`. 

Honestly, it should be a feature in ExtendR from day one;

In principle, I think we could also provide a macro that provides this, but panics would result in lack of `PutRNGState` call to back to R. Which would be very bad. So I'd be doing the same capturing that is done now for extendr-binded-functions for this macro. Unless the `try!`-block feature is stable one day, it could be a mixture of two.

CGMossa:
`#[extendr]` is a little overloaded right now. It will loose an argument soon in `use_try_from`.
But I still vote that the rng-thing stays in a prominent place.

multimeric:
Oh okay, are you saying that you need a finalizer that runs in case of errors? And that we can't separate this from the function-level error handler because [`try-blocks`](https://doc.rust-lang.org/unstable-book/language-features/try-blocks.html) is still unstable? I agree that supporting the `?` operator is imperative. The only way to do it as I suggest would be to pull in a macro from another crate: https://docs.rs/tryvial/latest/tryvial/macro.try_block.html. I guess editing the `extendr` macro is the best solution at the moment, but I'd like to see what @Ilia-Kosenkov thinks.

CGMossa:
I don't know what I'm saying, sorry I'm still under the weather.
The problem is that no matter what happens on the Rust-side, the `PutRngState` has to be called.
The way we are doing it right now by capturing panics, makes it fit with this. 

Making a macro-block would have to handle panics as well; But a first step would be just to support errors, i.e.
put rng state back even when an error is encountered. And that I don't even know how to do as a macro.
Besides the macro would be verbose, and I would personally hate to work within a macro;

Ilia-Kosenkov:
I need more information on this. Can we enable this globally? What's the overhead for calling it for each method (an equivalent of having `use_rng = true` all the time)?

CGMossa:
This shouldn't be enabled globally.
I think there are substantial performance costs, that we wouldn't be nice to spur on users.
Rcpp does it this way:
```cpp
// [[Rcpp::export(rng = false)]]
double myFunction(double input) {
// ...code that never uses the
// R random number generation...
}
```
[source](https://cran.r-project.org/web/packages/Rcpp/vignettes/Rcpp-attributes.pdf)

Frankly, I think it should be opt-in, and not opt-out, but any agreement we make here is fair by me.

Ilia-Kosenkov:
Well, we could go the Rcpp way for sure (but I agree with opt-in).

multimeric:
Isn't the way that Mossa proposed exactly the same as Rcpp?

CGMossa:
@multimeric the review on this one is also waiting to be resolved.
I think this is as done as can be.

multimeric:
I would kind of like some reassurance that this correctly handles a panic, but I can see that would be difficult. At least the macro compiles correctly.

multimeric (approved):

Ilia-Kosenkov:
This one has a lot of merge conflicts, @CGMossa can you resolve it & then merge?

CGMossa:
This is a little less trivial to merge now.

CGMossa:
Apologies. I've merged, and changed a lot of stuff, and now I would appreciate another review on this, just to make sure that it isn't completely insane.

CGMossa:
Ehh.. I checked it several times and did several merges... Can you tell me what you see that is wrong please?

CGMossa:
```
kevin@pop-os:~/GitHub/extendr$ git merge origin/master
Already up to date.
```

extendr-api/tests/rng_tests.rs
```diff
@@ -0,0 +1,40 @@
...
+}
+
+#[test]
+fn test_extendr_rng() {
+    test! {
+        R!(r#"set.seed(20230205)"#).unwrap();
+        let x = generate_big_random_vec();
+        R!(r#"set.seed(20230205)"#).unwrap();
+        let y = generate_big_random_vec();
+        assert_eq!(y, x);
```
Ilia-Kosenkov, line 33: Just to be safe, does `assert_eq!` correctly compare iterable types like `Vec<T>`?

extendr-api/tests/rng_tests.rs
```diff
@@ -0,0 +1,40 @@
...
+        // if the rng state isn't retrieved, then distribution is
+        // 0 always
+        assert!(
+            !x.iter().all(|x|x.round() as isize == 0)
```
Ilia-Kosenkov, line 37: Hm, if I understand correctly, a failed rng setup will result in exactly `0f64` values, no? Like, bitwise zero? What is the point of doing `.round()` and `as isize`?

How 'bout
```suggestion
            !x.iter().all(|&x| x == 0f64)
```

extendr-api/tests/rng_tests.rs
CGMossa: Sure:
```rust
>> assert_eq!(vec![13,4,3], vec![22]);
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `[13, 4, 3]`,
 right: `[22]`', src/lib.rs:6:1
stack backtrace:
   0: rust_begin_unwind
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/panicking.rs:67:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
   4: std::panicking::try
   5: run_user_code_1
   6: evcxr::runtime::Runtime::run_loop
   7: evcxr::runtime::runtime_hook
   8: evcxr::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
>> assert_eq!(vec![13,4,3], vec![13,4,3]);
```

extendr-api/tests/rng_tests.rs
CGMossa: Well 0 is not really.. I mean it is an abstract construction; so I thought, give me a `isize`, because that's really a `i32`.. But that was back then when I was young and naive. 

We can go with your version; I just want this merged.

Ilia-Kosenkov (approved):
Beautiful! Let's wait for CI checks

PR #492, opened 2023-02-26
Update libR-sys to 0.4.0

yutannihilation:
I didn't expect this, but bumping the version of libR-sys makes extendr fail because `[patch.crates-io]` doesn't work. It doesn't work because

* the version of libR-sys on Git is v0.4.0, but
* the version of libR-sys extendr specifies is v0.3.0

```
  warning: Patch `libR-sys v0.4.0 (https://github.com/extendr/libR-sys#9d21fc23)` was not used in the crate graph.
  Check that the patched package version and available features are compatible
  with the dependency requirements. If the patch has a different version from
  what is locked in the Cargo.lock file, run `cargo update` to use the new
  version. This may also occur with an optional dependency that is not enabled.
```

https://github.com/extendr/extendr/actions/runs/4266437196/jobs/7437481677#step:11:22

so we have to update these lines.

### DIFF
--- a/extendr-api/Cargo.toml
+++ b/extendr-api/Cargo.toml
@@ -15,7 +15,7 @@ license = "MIT"
 repository = "https://github.com/extendr/extendr"
 
 [dependencies]
-libR-sys = "0.3.0"
+libR-sys = "0.4.0"
 extendr-macros = { path = "../extendr-macros", version = "0.3.1" }
 extendr-engine = { path = "../extendr-engine", version = "0.3.1" }
 ndarray = { version = "0.15.3", optional = true }
--- a/extendr-engine/Cargo.toml
+++ b/extendr-engine/Cargo.toml
@@ -15,7 +15,7 @@ license = "MIT"
 repository = "https://github.com/extendr/extendr"
 
 [dependencies]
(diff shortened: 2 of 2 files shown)


multimeric (approved):

CGMossa (approved):

yutannihilation:
We can also comment out `[patch.crates-io]`. I feel "When uncommenting this" indicates that this should be commented out in usual, but I'm not immediately sure which is better. Let's not do it in this pull request.

https://github.com/extendr/extendr/blob/6b0a6c7e75da87600485117a68457b78ee350799/Cargo.toml#L9-L13

CGMossa:
We shouldn't version a dependency that is git-linked. But that doesn't matter right now.

multimeric:
I'm afraid I don't understand. Are we patching the crate version of libRsys with the git master version? Why?

CGMossa:
That's not possible;
See
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies
> Note: crates.io does not allow packages to be published with path
dependencies (path dev-dependencies are ignored). See the Multiple
locations section for a fallback alternative.

PR #491, opened 2023-02-25
Prepare metadata for release

multimeric:
* Bump versions for release
* Add myself as author

Depends on https://github.com/extendr/libR-sys/issues/115

PR #493, opened 2023-02-26
Remove outdated MAINTAINERS_GUIDE.md

yutannihilation:
See https://github.com/extendr/extendr/pull/491#issuecomment-1445367234


multimeric:
Hmm, could you not just add `--no-git-push` to each `cargo ws` command, and then instruct the reader to make a pull request? The other stuff still seems okay?

yutannihilation:
Yeah, they are okay, but I feel the guide no one reads is meaningless.

CGMossa (approved):
Let's delete it.
Then there is room for @yutannihilation and us to write what we actually do when we release.
Sometimes tabula rasa is the right management approach. Even if it invites more churn.

Issue #494, opened 2023-02-26
Enable all main features on `docs.rs`


multimeric:
By default I believe docs.rs uses the default features (?), meaning that `ndarray` etc are not shown on docs.rs.  We can fairly easily fix this: https://docs.rs/about/metadata.

multimeric:
Closed by #495.

PR #495, opened 2023-02-26
add features-gated API to docs.rs

CGMossa:
Based on this #494.

Using `all-features` on workspace doesn't work, because `libR-sys` has a feature that modifies the build process (with asking for path to generated bindings), rather than being additive.

I've added the currently to features to the metadata-field that was already there.

I've also fixed warnings. I think I fixed these somewhere else, but that PR is blocked because there is a parsing error somewhere that we need to deal with.

### DIFF
--- a/extendr-api/Cargo.toml
+++ b/extendr-api/Cargo.toml
@@ -48,4 +48,4 @@ tests-graphics = ["tests-minimal", "graphics"]
 tests-all = ["tests", "graphics"]
 
 [package.metadata.docs.rs]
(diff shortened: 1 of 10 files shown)


multimeric (approved):

multimeric:
So this is good to go?

CGMossa:
Yes. I'm *always* incredibly hesitant to merge anything. Will merge this now.

multimeric:
Great, thanks for this. I was happy to merge, just wanted to make sure you were happy with it.

PR #497, opened 2023-02-27
FOR TESTING ONLY: Testing pr extendr/libr-sys#141

CGMossa:
extendr/libr-sys#141


CGMossa:
Alright; I'm re-running everything, but it already worked out, and I've addressed comments from you @multimeric.

Issue #500, opened 2023-03-04
Conversion trait implementation issue related to `ndarray`


Ilia-Kosenkov:
There is an issue with `TryFrom` implemented for `ArrayView1`,  but not for `ArrayView2`.

``` r
rextendr::rust_source(
    code = "#[extendr(use_try_from = true)] fn this_works(_x : ArrayView2<Rint>) {}",
    extendr_deps = list(
        `extendr-api` = list(
            git = "https://github.com/extendr/extendr",
            features = array("ndarray")
        )
    )
)
#> ℹ build directory: 'C:/Users/.../AppData/Local/Temp/RtmpAXOM9P/file3f586e29777a'
#> ✔ Writing 'C:/Users/.../AppData/Local/Temp/RtmpAXOM9P/file3f586e29777a/target/extendr_wrappers.R'

rextendr::rust_source(
    code = "#[extendr(use_try_from = true)] fn this_doesnt(_x : ArrayView1<Rint>) {}",
    extendr_deps = list(
        `extendr-api` = list(
            git = "https://github.com/extendr/extendr",
            features = array("ndarray")
        )
    )
)
#> ℹ build directory:
#> 'C:/Users/.../AppData/Local/Temp/RtmpAXOM9P/file3f586e29777a'
#> Error in `invoke_cargo()`:
#> ! Rust code could not be compiled successfully. Aborting.
#> ✖ error[E0277]: the trait bound `ArrayBase<ViewRepr<&extendr_api::prelude::Rint>, Dim<[usize; 1]>>: From<Robj>` is not satisfied
#>  --> src\lib.rs:2:1
#>   |
#> 2 | #[extendr(use_try_from = true)] fn this_doesnt(_x : ArrayView1<Rint>) {}
#>   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<Robj>` is not implemented for `ArrayBase<ViewRepr<&extendr_api::prelude::Rint>, Dim<[usize; 1]>>`
#>   |
#>   = help: the following other types implement trait `From<T>`:
#>             <ArrayBase<CowRepr<'a, A>, D> as From<&'a ArrayBase<S, D>>>
#>             <ArrayBase<CowRepr<'a, A>, D> as From<ArrayBase<OwnedRepr<A>, D>>>
#>             <ArrayBase<CowRepr<'a, A>, D> as From<ArrayBase<ViewRepr<&'a A>, D>>>
[...]

PR #501, opened 2023-03-04
BUG FIX: generate both `TryFrom` for `Robj` and `&Robj` for ndarray view 1.

CGMossa:
Will try to address #500

### DIFF
--- a/extendr-api/src/robj_ndarray.rs
+++ b/extendr-api/src/robj_ndarray.rs
@@ -81,7 +81,7 @@ where
 
 macro_rules! make_array_view_1 {
     ($type: ty, $error_fn: expr) => {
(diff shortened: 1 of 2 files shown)


CGMossa:
I think the bug in #500 occurred because none of the tests try to create the ndarray view from an `Robj`, but `&Robj`.
Thus I added a test for that. Feel free to code-golf the test.

Ilia-Kosenkov (approved):

CGMossa:
For some reason, I had convinced myself that @multimeric wrote this code. But that's not the case.
Thus I'll just merge now that we have one approval.

CGMossa:
Closes #500

Issue #478, opened 2023-02-10
Replace `ndarray` macro use with generics


multimeric:
We currently have a macro-based conversion between R vector and ArrayView (`make_array_view_1!`). I think the only reason we do that instead of using generics is so we can throw the correct error type if the conversion fails. This can be resolved by adding the error as an associated type to the AsTypedSlice trait, allowing the use of generics.

CGMossa:
I'm looking at this now. For this to work, you'd need to split 
```rust

#[derive(Debug, PartialEq)]
pub enum Error {
    Panic(Robj),
    NotFound(Robj),
    EvalError(Robj),
    ParseError(Robj),
    NamesLengthMismatch(Robj),
    [...]
```
into other small error-types, and inject them in `make_typed_slice!` macro?

Because the associated type doesn't allow to specify an enum variant via type.

PR #502, opened 2023-03-05
Moved from macros to generic-impls for type conversions

CGMossa:
Addresses #478. 

I would like a review of this; Is this worth it? Is `AsTypedSlice` the right place to put in an error?

### DIFF
--- a/extendr-api/src/robj/mod.rs
+++ b/extendr-api/src/robj/mod.rs
@@ -789,7 +789,6 @@ macro_rules! make_typed_slice {
     }
 }
 
-
 make_typed_slice!(Rbool, ExpectedLogical, INTEGER, LGLSXP);
 make_typed_slice!(i32, ExpectedInteger, INTEGER, INTSXP);
 make_typed_slice!(u32, ExpectedInteger, INTEGER, INTSXP);
--- a/extendr-api/src/robj/mod.rs
+++ b/extendr-api/src/robj/mod.rs
@@ -732,6 +732,7 @@ pub trait AsTypedSlice<'a, T>
 where
     Self: 'a,
 {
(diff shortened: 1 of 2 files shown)


multimeric:
This looks good to me. I think it's nice to move the error to the typed slice trait because it should be reusable to other conversions that use it.

CGMossa:
I don't known.. Then we need to do some renaming of things I think.
I need more instructions to polish this I think, as I'm not convinced myself..

extendr-api/src/robj/mod.rs
```diff
@@ -776,22 +781,26 @@ macro_rules! make_typed_slice {
                     _ => None
                 }
             }
+
+            fn error(robj: Robj) -> Self::Error {
```
multimeric, line 785: I think there is an argument that the `as_typed_slice*` functions should return a `Result` rather than an `Option`. Because then the trait becomes much simpler.

multimeric (commented):
I think a nice "proof" that this works would be to use this error type, and simplify the code in:
* `TryFrom<&Robj> for Vec<>`
* `TryFrom<Robj> for RMatrix<>`
* `TryFrom<Robj> for RColumn<>`

CGMossa:
Why do we even need the error to be an associated type? We have one major error-type for the whole thing? We can just insert it. We never, ever change the error type anyways.

multimeric:
Yeah, you're right. It doesn't need to be an associated type. I'm not sure why I thought it did.

CGMossa:
I've decided to overhaul this by starting from the "bottom" So this PR might never get merged, or maybe needs to be repurposed entirely. Let us wait on #509, #508, #505.

PR #505, opened 2023-03-05
Add const-generic impl for `Robj`

CGMossa:
By way of work on #502, I've hit on a lot of things that I think we should address.
This work has grown a lot, and instead of making a huge PR with all of, I'll try to make smaller ones here..

### DIFF
--- a/extendr-api/src/robj/into_robj.rs
+++ b/extendr-api/src/robj/into_robj.rs
@@ -587,26 +587,16 @@ macro_rules! impl_from_as_iterator {
 //     }
 // }
 
-// Template constants are still unstable in rust.
-impl_from_iter! {[T; 1]}
-impl_from_iter! {[T; 2]}
-impl_from_iter! {[T; 3]}
-impl_from_iter! {[T; 4]}
-impl_from_iter! {[T; 5]}
-impl_from_iter! {[T; 6]}
-impl_from_iter! {[T; 7]}
-impl_from_iter! {[T; 8]}
-impl_from_iter! {[T; 9]}
-impl_from_iter! {[T; 10]}
-impl_from_iter! {[T; 11]}
-impl_from_iter! {[T; 12]}
-impl_from_iter! {[T; 13]}
-impl_from_iter! {[T; 14]}
-impl_from_iter! {[T; 15]}
-impl_from_iter! {[T; 16]}
-impl_from_iter! {[T; 17]}
-impl_from_iter! {[T; 18]}
-impl_from_iter! {[T; 19]}
+impl<'a, T, const N: usize> From<[T; N]> for Robj
+where
+    Self: 'a,
+    T: ToVectorValue,
+{
+    fn from(val: [T; N]) -> Self {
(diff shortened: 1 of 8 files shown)


extendr-api/src/robj/into_robj.rs
```diff
@@ -587,26 +587,16 @@ macro_rules! impl_from_as_iterator {
...
+where
+    Self: 'a,
+    T: ToVectorValue,
+{
+    fn from(val: [T; N]) -> Self {
+        val.into_iter().collect_robj()
+    }
+}
+
 impl_from_iter! {Vec<T>}
```
multimeric, line 600: Could you do another generic impl for vectors, so we can delete `impl_from_iter!` entirely?

multimeric (approved):

CGMossa:
Done. Please merge when you're satisfied with it @multimeric.

extendr-api/src/robj/into_robj.rs
```diff
@@ -587,28 +570,37 @@ macro_rules! impl_from_as_iterator {
...
+        value.into_iter().collect_robj()
+    }
+}
+
+impl<T> From<&Vec<T>> for Robj
+where
+    T: ToVectorValue + Clone,
+{
+    fn from(value: &Vec<T>) -> Self {
+        value.into_iter().cloned().collect_robj()
```
multimeric, line 601: Do we not have an impl for borrow vector values? ie can't you remove `cloned()` here?

multimeric (approved):
Looks good! Always like to see a PR that removes more code than it adds.

extendr-api/src/robj/into_robj.rs
multimeric: Yes I get that you have references, I'm just saying that you can directly convert references to Robj:

```rust
>> test!{
   let vector = vec![1, 2, 3].into_iter().collect_robj();
   call!("print", pairlist!(vector));
}
[[1]]
[1] 1 2 3

()
```

CGMossa: Okay. I've made it compile. But I do not trust that this works at all. I will see if there are tests, but it is simply impossible for me to see how `ToVectorValue` is defined for ref-types in a non-clone way.

CGMossa: Fine, I believe it works:

```rust
    #[test]
    fn test_to_robj_from_references() {
        let vs = vec![0_f32, 2.4, 2.3, 222.2];
        test! { Robj::from(&vs); }
    }
```
This compiled and ran fine..

CGMossa:
```rust
    #[test]
    fn test_to_robj_from_references() {
        let vs = vec![0_f32, 2.4, 2.3, 222.2];
        test! { let vs_back = Robj::from(&vs); println!("{vs_back:?}"); }
    }
```
`[0.0, 2.4000000953674316, 2.299999952316284, 222.1999969482422]`

multimeric (approved):

PR #506, opened 2023-03-05
Add `CannotBeNA`-trait to be able to distinguish (maybe) between types

CGMossa:
I'd admit this is a little naive.
If we are to make generic impls instead of macro-based ones, I think we need
the tools to distinguish between `CanBeNA` and `CannotBeNA`. Rust doesn't
support negative bounds, but hopefully this distinction can be useful.
For now, this is a DRAFT, and I hope the use-case appears soon.

### DIFF
--- a/extendr-api/src/na.rs
+++ b/extendr-api/src/na.rs
@@ -23,9 +23,11 @@ lazy_static! {
     };
 }
 
-/// Return true if this primitive is NA.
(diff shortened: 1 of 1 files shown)


CGMossa:
It might make sense to add this, and we did discuss it, but no-one is approving, and I don't see how it can be used **right now**, so closing.

PR #508, opened 2023-03-05
`is_na` method works only if underlying type is a scalar

CGMossa:
This is about clarity.

`Robj` has a public method `is_na`, which only returns true if it is `NA` **scalar**.
Meaning if the underlying r-type is of length not equal `1`, the `NA` is not ignored.

I've renamed it for clarity.

I hope you'll agree with me that this is better.

### DIFF
--- a/extendr-api/src/na.rs
+++ b/extendr-api/src/na.rs
@@ -23,9 +23,11 @@ lazy_static! {
     };
 }
 
-/// Return true if this primitive is NA.
+/// Trait signifying if a given type may take the value `NA` in the R sense.
 pub trait CanBeNA {
+    /// Returns `true` if it is `NA`.
     fn is_na(&self) -> bool;
+    /// Returns the sentinel `NA` value for this type.
     fn na() -> Self;
 }
 
--- a/extendr-api/src/robj/mod.rs
+++ b/extendr-api/src/robj/mod.rs
@@ -322,7 +322,7 @@ pub trait Types: GetSexp {
 impl Types for Robj {}
 
 impl Robj {
-    /// Is this object is an NA scalar?
+    /// Is this object is an `NA` scalar?
     /// Works for character, integer and numeric types.
     /// ```
     /// use extendr_api::prelude::*;
@@ -333,7 +333,7 @@ impl Robj {
     /// assert_eq!(r!(NA_STRING).is_na(), true);
     /// }
     /// ```
-    pub fn is_na(&self) -> bool {
+    pub fn is_na_scalar(&self) -> bool {
         if self.len() != 1 {
             false
         } else {
--- a/extendr-api/src/robj/from_robj.rs
+++ b/extendr-api/src/robj/from_robj.rs
@@ -78,7 +78,7 @@ impl<'a> FromRobj<'a> for bool {
 
 impl<'a> FromRobj<'a> for &'a str {
     fn from_robj(robj: &'a Robj) -> std::result::Result<Self, &'static str> {
-        if robj.is_na() {
+        if robj.is_na_scalar() {
             Err("Input must not be NA.")
         } else if let Some(s) = robj.as_str() {
             Ok(s)
@@ -90,7 +90,7 @@ impl<'a> FromRobj<'a> for &'a str {
 
 impl<'a> FromRobj<'a> for String {
     fn from_robj(robj: &'a Robj) -> std::result::Result<Self, &'static str> {
-        if robj.is_na() {
+        if robj.is_na_scalar() {
             Err("Input must not be NA.")
         } else if let Some(s) = robj.as_str() {
             Ok(s.to_string())
@@ -122,7 +122,7 @@ impl<'a> FromRobj<'a> for Vec<f64> {
 
 impl<'a> FromRobj<'a> for Vec<String> {
     fn from_robj(robj: &'a Robj) -> std::result::Result<Self, &'static str> {
-        if robj.is_na() {
+        if robj.is_na_scalar() {
             Err("Input must be a character vector. Got 'NA'.")
         } else if let Some(v) = robj.as_string_vector() {
             let str_vec = v.to_vec();
@@ -187,7 +187,7 @@ impl<'a> FromRobj<'a> for HashMap<&str, Robj> {
 // NA-sensitive integer input handling
 impl<'a> FromRobj<'a> for Option<i32> {
     fn from_robj(robj: &'a Robj) -> std::result::Result<Self, &'static str> {
-        if robj.is_na() {
+        if robj.is_na_scalar() {
             Ok(None)
         } else if let Some(val) = robj.as_integer() {
             Ok(Some(val))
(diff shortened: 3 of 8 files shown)


extendr-api/src/na.rs
```diff
@@ -23,9 +23,11 @@ lazy_static! {
     };
 }
 
-/// Return true if this primitive is NA.
+/// Trait signifying if a given type may take the value `NA` in the R sense.
```
multimeric, line 26: This isn't clear. Does this trait mean it can be an NA scalar, or it can be a vector that contains NA?

extendr-api/src/robj/mod.rs
```diff
@@ -322,18 +322,18 @@ pub trait Types: GetSexp {
 impl Types for Robj {}
 
 impl Robj {
-    /// Is this object is an NA scalar?
+    /// Is this object is an `NA` scalar?
```
multimeric, line 325: I would elaborate that this function still operates on vectors, it's just that it checks for length-1 vectors. ie it checks for scalars in the R sense, but not the rust sense.

multimeric (changes requested):

extendr-api/src/robj/mod.rs
CGMossa: Alright; Given an `Robj` that is a vector `c(NA, NA)` it will say `false` though. Thus I need it to be crystal clear on that. Especially since it is used everywhere in the Rust conversion code.

multimeric: You haven't explained this in doc comment though?

extendr-api/src/na.rs
multimeric: I'm still not satisfied here. "take" is overloaded, and "in the R sense" is not precise enough.

multimeric (changes requested):
Both traits need to be precisely defined in terms of the dimensionality of the type. Is it precisely length 1, or can it be longer?

extendr-api/src/na.rs
JosiahParry: What about 

> Trait signifying if a given type may be represented as an `NA` value in R

extendr-api/src/robj/mod.rs
JosiahParry: Perhaps compare this to the vectorized `|` and `&` vs scalar comparison of `||` and `&&`. 

> `is_na_scalar()` is intended to be used on individual elements of a vector. It is _not_ vectorized and will return `false` if the vector has a length greater than 1.

CGMossa:
Thanks for the input. It is very close to what I wrote latest. But you're
somehow still seeing the code that was there when Michael reviewed it. Do
you mind writing your own review? Your comments are as valuable to us as
our own is. And maybe you'll see what I wrote next.

extendr-api/src/robj/mod.rs
```diff
@@ -322,18 +322,23 @@ pub trait Types: GetSexp {
 impl Types for Robj {}
 
 impl Robj {
-    /// Is this object is an NA scalar?
-    /// Works for character, integer and numeric types.
+    /// Returns `true` if the underlying value is a scalar `NA` value,
+    ///  otherwise `false`.
+    ///
+    /// NOTE: If the underlying value is not a single element vector then return `false`,
```
JosiahParry, line 328: Does `NOTE` get rendered specially? Otherwise I would omit it.

CGMossa, line 328:
```suggestion
    /// If the underlying value is not a single element vector then return `false`,
```

extendr-api/src/na.rs
```diff
@@ -23,9 +23,16 @@ lazy_static! {
...
 }
 
-/// Return true if this primitive is `NA`.
+/// Trait signifying if a given scalar type may be represented as an `NA` vlaue
+/// in an R vector.
+///
+/// The `is_na_scalar` method is intended to be used on individual
+/// elements of a vector. It is not vectorized and will return `false` if the
+/// vector has a length greater than 1.
 pub trait CanBeNA {
```
JosiahParry, line 32: Do we want to add the `Scalar` trait bound here?

extendr-api/src/na.rs
```diff
@@ -23,9 +23,16 @@ lazy_static! {
...
+    /// Returns `true` if it is `NA`.
     fn is_na(&self) -> bool;
```
JosiahParry, line 34: I see the `is_na_scalar()` method associated with the `Robj`  struct but not the `CanBeNA` trait. Should the Robj impl `CanBeNA` and not have its own method?

JosiahParry:
The current behavior is like `rlang::is_na()` which is in a questioning lifecycle. 

> `is_na()` is a scalar predicate and always returns a scalar boolean, TRUE or FALSE. If its input is not scalar, it returns FALSE.

I think this is confusing! The rlang alternative for a vectorized one is `are_na()`. Which...when you think about it is a good linguistic choice. 

I think the NA behavior needs to be broken down and clarified a bit more. We we have two types of NA checking that we need to be able to do. 

1. Scalars which is where `CanBeNA` trait comes into play
2. On non-scalar R objects

- Rbool, Rint, Rstr, Rfloat, Rcplx return a scalar `bool` with `is_na()`. 
- Logicals, Doubles, Integers, Strings should return a `Vec<bool>` perhaps with `are_na()` method? 
- List should return a `Vec<bool>` as well. `is.na(l)` only returns `TRUE` when the element is a scalar NA value.
- RArray / Matrix should probably return an array of the dimension with `bool` type?? Idk about this one
- `DataFrame` should either behave like a list or only have the scalar method

I'd like to get this to be nice and clean! What is the best way to have these NA options? Should there be two functions `is_na()` and `is_na_scalar()` or maybe `is_na()` and `are_na()`? Are these made available via a single trait? What does this look like from an implementation perspective? 


Reprex for some R behavior context:

```r
library(rlang)
l <- list(x = 1, b = c(NA, NA), c = NULL) 

is.na(l)
#>     x     b     c 
#> FALSE FALSE FALSE
is_na(l)
#> [1] FALSE
are_na(l)
#> Error in `are_na()`:
#> ! `x` must be an atomic vector, not a list.
#> Backtrace:
#>     ▆
#>  1. └─rlang::are_na(x = l)
#>  2.   └─rlang:::stop_input_type(x, "an atomic vector")
#>  3.     └─rlang::abort(message, ..., call = call, arg = arg)

f <- c(0.1, NA, 2.4)

is.na(f)
#> [1] FALSE  TRUE FALSE
is_na(f)
#> [1] FALSE
are_na(f)
#> [1] FALSE  TRUE FALSE
```

<sup>Created on 2023-10-03 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

CGMossa:
This last response is much better. I'd love it if you made it an ISSUE, and
then we should make a new PR, with a new conversation, really digging in
and dealing with these semantic pitfalls.

JosiahParry:
See https://github.com/extendr/extendr/issues/633

Issue #498, opened 2023-03-01
Conversion trait architecture


multimeric:
Although everyone is agreed that the conversion traits we use to convert to and from Robj should be fallible (ie return a `Result`), there has been some disagreement over whether we should use vanilla `TryFrom` as the conversion trait as we do now when `#[extendr(use_try_from = true)]` is set, or if we should simply change `FromRobj` to have a fallible method.

From discussions mostly on discord, there seem to be 3 main advantages of using a custom trait:
1. We get to allocate one 1 blanket implementation to the conversion trait, whereas `TryFrom` wouldn't allow this
2. A custom trait allows us to use custom associated types, although I can't think of a use case for this in extendr (input welcome)
3. A custom trait would allow us to make the trait generic, with a "marker type" as described in https://github.com/extendr/extendr/issues/440#issuecomment-1448873888. This may help resolve the issue described in that thread, whereby third party crates would have trouble "extendrizing" existing types.

As I see it, the benefits of using `TryFrom` instead are:
* Simple, familiar trait
* Some of the implementation has already been done on this trait. It wouldn't be hard to change this, however.

This is an important issue because it impacts the core design of extendr. In particular https://github.com/extendr/extendr/issues/336, which is generally agreed to be a requirement for our 1.0 release.

Any feedback is welcome on this topic.

multimeric:
@yutannihilation and @andy-thomason I'd love some input if you have time. This is been discussed a bit on discord and it seems like moving to a custom, faillible trait everywhere is the best option, but I want everyone's input before I implement this.

yutannihilation:
Thanks for inviting me. But, please ignore me on deciding the direction on this. I think I'm in a different direction and I don't want to disturb you. I think we should first research carefully on how other frameworks provide the ergonomics around conversion, but, considering Bevy chooses this way, it should be reasonable.

multimeric:
One downside I've identified of having a conversion trait is that it's difficult to write different implementations for e.g. an owned value, a reference, a mutable reference etc. With TryFrom you just say `TryFrom<&Robj>` etc whereas with `FromRobj` some thought might be needed to work this out. One solution might be to define `FromRobj` as `from_robj(robj: Borrow<Robj>) -> Result<Self>` which is implemented for free for `&Robj`, `Robj`, `Box<Robj>` etc.

multimeric:
@andy-thomason 

> Making a Robj is expensive, however, as we need to protect the data from the GC and so using &Robj
[...]

PR #509, opened 2023-03-05
Demacrofied `AsTypedSlice`

CGMossa:
The plan thus far is:

- Figure out and flatten `AsTypedSlice`.
- Impl. FromRobj using `AsTypedSlice`.
- Remove the methods on Robj that doesn't really do anything special 
  and use only FromRobj and TryFrom for conversions.
- Try to have a CanBeNA(T: CanBeNA) and implement it for those that can contain it.
- Remove all `Error = &'static str`, and replace it with our `Error`. No associated types, nothing.
- Add error to AsTypedSlice;

This is step one here; It is done due to #502 and #498. 
I hope you haven't worked on this @multimeric too much.. As I've dug into it as well, and would appreciate some reviews on this.

### DIFF
--- a/extendr-api/src/robj/mod.rs
+++ b/extendr-api/src/robj/mod.rs
@@ -734,64 +734,131 @@ where
 {
     fn as_typed_slice(&self) -> Option<&'a [T]>
     where
-        Self: 'a,
-    {
-        None
-    }
+        Self: 'a;
 
     fn as_typed_slice_mut(&mut self) -> Option<&'a mut [T]>
     where
-        Self: 'a,
-    {
-        None
+        Self: 'a;
+}
+
+trait SExpAccessor {
+    const XP: u32;
+    type AccessorType;
+    fn acessor(x: SEXP) -> *mut Self::AccessorType;
+}
+
+impl SExpAccessor for Rbool {
+    const XP: u32 = LGLSXP;
+    type AccessorType = raw::c_int;
+    fn acessor(x: SEXP) -> *mut Self::AccessorType {
+        unsafe { INTEGER(x) }
+    }
+}
+impl SExpAccessor for i32 {
+    const XP: u32 = INTSXP;
+    type AccessorType = raw::c_int;
+    fn acessor(x: SEXP) -> *mut Self::AccessorType {
+        unsafe { INTEGER(x) }
+    }
+}
+impl SExpAccessor for u32 {
+    const XP: u32 = INTSXP;
+    type AccessorType = raw::c_int;
+    fn acessor(x: SEXP) -> *mut Self::AccessorType {
+        unsafe { INTEGER(x) }
+    }
+}
+impl SExpAccessor for Rint {
+    const XP: u32 = INTSXP;
+    type AccessorType = raw::c_int;
+    fn acessor(x: SEXP) -> *mut Self::AccessorType {
+        unsafe { INTEGER(x) }
+    }
+}
+impl SExpAccessor for f64 {
+    const XP: u32 = REALSXP;
+    type AccessorType = Self;
+    fn acessor(x: SEXP) -> *mut Self::AccessorType {
+        unsafe { REAL(x) }
+    }
+}
+impl SExpAccessor for Rfloat {
+    const XP: u32 = REALSXP;
+    type AccessorType = f64;
+    fn acessor(x: SEXP) -> *mut Self::AccessorType {
+        unsafe { REAL(x) }
+    }
+}
+impl SExpAccessor for u8 {
+    const XP: u32 = RAWSXP;
+    type AccessorType = Rbyte;
+    fn acessor(x: SEXP) -> *mut Self::AccessorType {
+        unsafe { RAW(x) }
+    }
+}
+impl SExpAccessor for Robj {
+    const XP: u32 = VECSXP;
+    type AccessorType = Rbyte;
+    fn acessor(x: SEXP) -> *mut Self::AccessorType {
+        unsafe { VECTOR_PTR(x) }
+    }
+}
+impl SExpAccessor for Rstr {
+    const XP: u32 = STRSXP;
+    type AccessorType = SEXP;
+    fn acessor(x: SEXP) -> *mut Self::AccessorType {
+        unsafe { STRING_PTR(x) }
+    }
+}
+impl SExpAccessor for c64 {
+    const XP: u32 = CPLXSXP;
+    type AccessorType = Rcomplex;
+    fn acessor(x: SEXP) -> *mut Self::AccessorType {
+        unsafe { COMPLEX(x) }
+    }
+}
+impl SExpAccessor for Rcplx {
+    const XP: u32 = CPLXSXP;
+    type AccessorType = Rcomplex;
+    fn acessor(x: SEXP) -> *mut Self::AccessorType {
+        unsafe { COMPLEX(x) }
+    }
+}
+impl SExpAccessor for Rcomplex {
+    const XP: u32 = CPLXSXP;
+    type AccessorType = Rcomplex;
+    fn acessor(x: SEXP) -> *mut Self::AccessorType {
+        unsafe { COMPLEX(x) }
     }
 }
 
-macro_rules! make_typed_slice {
-    ($type: ty, $fn: tt, $($sexp: tt),* ) => {
-        impl<'a> AsTypedSlice<'a, $type> for Robj
-        where
-            Self : 'a,
-        {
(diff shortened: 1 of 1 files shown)


CGMossa:
Also I hope extendr/libr-sys#145 solves itself..

CGMossa:
As per extensive discussion over at extendr/libr-sys#144, I've removed `VECTOR_PTR`. It was causing clippy warnings on some platforms, but not others. The bindgen is depends on the platform and the used clang version; It is just a mess.

extendr-api/src/robj/mod.rs
```diff
@@ -734,65 +734,125 @@ where
...
 
     fn as_typed_slice_mut(&mut self) -> Option<&'a mut [T]>
     where
-        Self: 'a,
-    {
-        None
+        Self: 'a;
+}
+
+trait SExpAccessor {
```
multimeric, line 744: Needs docs. Also try to think of a more precise name for this. `SexpAccessibleAs?` Also, can this be generic rather than using an associated type?

extendr-api/src/robj/mod.rs
```diff
@@ -734,65 +734,125 @@ where
...
+    const XP: u32;
+    type AccessorType;
+    fn acessor(x: SEXP) -> *mut Self::AccessorType;
```
multimeric, line 747: It's unusual for a method to be a noun. Is there an appropriate verb you can use here? `access()`?

extendr-api/src/robj/mod.rs
```diff
@@ -734,65 +734,125 @@ where
...
-                        }
-                    }
-                    _ => None
-                }
+impl<'a, T> AsTypedSlice<'a, T> for Robj
+where
+    Self: 'a,
+    T: SExpAccessor,
+{
+    fn as_typed_slice(&self) -> Option<&'a [T]> {
```
multimeric, line 833: Since we're rewriting this, I again think we should consider if this should return a `Result`. One good reason is so we can add an error message.

multimeric (changes requested):

extendr-api/src/robj/mod.rs
CGMossa: First, I need to get all these small changes and PRs through. Then I'm doing that. I did an MVP and it made sense, and you could do it, but to de-macrofi things as well, a little more care has to be dealt.
But it is on the agenda. I just have to survive this hellish loop of compilation issues.

extendr-api/src/robj/mod.rs
CGMossa: Unfortunately, it has to be an associated type.

CGMossa: I don't know what to describe it as..

extendr-api/src/robj/mod.rs
CGMossa: Renamed.

extendr-api/src/robj/mod.rs
CGMossa: To begin with, I just wrote the trait necessary to distinguish between the types and in order to make the non-macro impl work. I didn't intend on diving into what this is and what it does.
But I've gained some understanding: Basically it unsafely-casts a pointer from an opaque pointer, to a rust-typed pointer.
I've renamed things to reflect that. 
A vector or not will be hidden behind a pointer regardless; In fact everything is hidden behind this type anyways, so there isn't really a need to write that. I've put in a reference to `Rtype` which shows that there are `XP` (C-ints that are ids for r-types) for everything. Incidentally, I've renamed `XP` to `R_TYPE_ID`.
Hope that's good enough for this PR. I don't mind exploring other things now that they come up, but I'm not sure this should always happen in PRs. This time I think it was fine.

extendr-api/src/lib.rs
```diff
@@ -425,6 +425,8 @@ pub unsafe fn register_call_methods(info: *mut libR_sys::DllInfo, metadata: Meta
 }
 
 /// Type of R objects used by [Robj::rtype].
+///
+/// [Robj::rtype]: crate::robj::Types::rtype
```
multimeric, line 429: What is this saying?

extendr-api/src/robj/mod.rs
```diff
@@ -734,65 +733,143 @@ where
...
-        None
+        Self: 'a;
+}
+
+/// Trait for providing a rust typed pointer to an r type.
+/// R data are hidden behind [`SEXP`] opaque pointers with an identifier
+/// for the underlying r storage type. Thus this trait is only meaningful for
+/// [`SEXP`]-wrappers.
+trait SexpAs {
+    /// Identifier for the R object modes / types represented as a value.
```
multimeric, line 748:
```suggestion
    /// Marker for the R object modes / types represented as a value.
```

I tend to expect an identifier is a string, although this is of course not always true

extendr-api/src/robj/mod.rs
```diff
@@ -734,65 +733,143 @@ where
...
+    /// See [`Rtype`] for reference.
+    const R_TYPE_ID: u32;
+    /// Corresponding rust type to `XP`.
```
multimeric, line 751: What is `XP`?

extendr-api/src/robj/mod.rs
```diff
@@ -734,65 +733,143 @@ where
...
-    {
-        None
+        Self: 'a;
+}
+
+/// Trait for providing a rust typed pointer to an r type.
+/// R data are hidden behind [`SEXP`] opaque pointers with an identifier
+/// for the underlying r storage type. Thus this trait is only meaningful for
+/// [`SEXP`]-wrappers.
+trait SexpAs {
```
multimeric, line 747: Could this be `SexpAsPtr` for clarity?

extendr-api/src/robj/mod.rs
```diff
@@ -734,65 +733,143 @@ where
...
+    /// See [`Rtype`] for reference.
+    const R_TYPE_ID: u32;
+    /// Corresponding rust type to `XP`.
+    type Type;
+    /// Returns a typed pointer to the underlying value
+    fn as_(x: SEXP) -> *mut Self::Type;
+}
+
+impl SexpAs for Rbool {
+    /// Corresponds to `logical` in R
```
multimeric, line 758: I think these comments should either be comments (`//`) or impl-level rustdocs (above the `impl`)

extendr-api/src/robj/mod.rs
```diff
@@ -734,65 +733,143 @@ where
...
 
     fn as_typed_slice_mut(&mut self) -> Option<&'a mut [T]>
     where
-        Self: 'a,
-    {
-        None
+        Self: 'a;
+}
+
+/// Trait for providing a rust typed pointer to an r type.
```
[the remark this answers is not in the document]
multimeric: Oh, I see, it's a pointer to a vector. This should be here in the doc

multimeric (dismissed):
Looks much better!

extendr-api/src/robj/mod.rs
CGMossa: `SEXP` is already a pointer. So yes.. maybe that would work better :)

extendr-api/src/robj/mod.rs
CGMossa: Why? The'll be annotated on the assoc. type in the docs properly. 
I truly dislike `//` usage for top-level items...

extendr-api/src/robj/mod.rs
multimeric: I'm not married to this idea. You can resolve if you don't like it.

extendr-api/src/robj/mod.rs
multimeric: Oh, I thought you intended this comment to be at struct level. If it's supposed to be just for the `R_TYPE_ID` then this is fine.

CGMossa (commented):
I don't know what's going on.

Ilia-Kosenkov:
I am not enitrely sure what is the benefit here -- you just spelled out a lot of boilerplate. I like the idea that we expose data pointer in a separate trait and then construct typed slice using blanket implementation, but perhaps `SexpAsPtr` can be macrofied? :-D 

Also, I'd mark trait or just `as_` as `unsafe` to explicitly communicate that this is dangerous.

CGMossa:
@Ilia-Kosenkov I'm unable to merge this, even after going through a comprehensive review process. Would you mind going through it one more time, please read summary here https://github.com/extendr/extendr/pull/509#issuecomment-1699558926 and maybe we can merge this?

PR #512, opened 2023-03-07
Minor warning clean-up

CGMossa:
Warnings were occurring due to:
- `VECTOR_PTR` being deprecated and thus invoking never-type `!`, so it is removed
- Various places where `unreachable` is called, but now we can just use `!` instead.

This is meant to be a quick PR that relaxes some of the nuisance that comes up..

### DIFF
--- a/extendr-api/src/thread_safety.rs
+++ b/extendr-api/src/thread_safety.rs
@@ -77,23 +77,17 @@ where
 {
     match std::panic::catch_unwind(f) {
         Ok(res) => res,
-        Err(_) => {
-            unsafe {
-                libR_sys::Rf_error(err_str.as_ptr() as *const std::os::raw::c_char);
-            }
-            unreachable!("handle_panic unreachable")
-        }
+        Err(_) => unsafe { libR_sys::Rf_error(err_str.as_ptr() as *const std::os::raw::c_char) },
     }
 }
 
 static mut R_ERROR_BUF: Option<std::ffi::CString> = None;
 
-pub fn throw_r_error<S: AsRef<str>>(s: S) {
+pub fn throw_r_error<S: AsRef<str>>(s: S) -> ! {
     let s = s.as_ref();
     unsafe {
         R_ERROR_BUF = Some(std::ffi::CString::new(s).unwrap());
         libR_sys::Rf_error(R_ERROR_BUF.as_ref().unwrap().as_ptr());
-        unreachable!("Rf_error does not return");
     };
 }
 
--- a/extendr-api/src/error.rs
+++ b/extendr-api/src/error.rs
@@ -11,7 +11,6 @@ pub fn unwrap_or_throw<T>(r: std::result::Result<T, &'static str>) -> T {
     match r {
         Err(e) => {
             throw_r_error(e.to_string());
-            unreachable!("");
         }
         Ok(v) => v,
     }
@@ -22,7 +21,6 @@ pub fn unwrap_or_throw_error<T>(r: std::result::Result<T, Error>) -> T {
     match r {
         Err(e) => {
             throw_r_error(e.to_string());
-            unreachable!("");
         }
         Ok(v) => v,
     }
--- a/extendr-api/src/robj/mod.rs
+++ b/extendr-api/src/robj/mod.rs
@@ -787,7 +787,6 @@ make_typed_slice!(Rint, INTEGER, INTSXP);
 make_typed_slice!(f64, REAL, REALSXP);
 make_typed_slice!(Rfloat, REAL, REALSXP);
 make_typed_slice!(u8, RAW, RAWSXP);
-make_typed_slice!(Robj, VECTOR_PTR, VECSXP);
 make_typed_slice!(Rstr, STRING_PTR, STRSXP);
 make_typed_slice!(c64, COMPLEX, CPLXSXP);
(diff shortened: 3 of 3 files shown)


CGMossa:
R-devel changed things, so this is why this is failing.
If anyone has ideas for this, we can try it out? Otherwise, we can close it for now.

CGMossa:
See an overview of `Rf_Error`s to see why all `devel` versions return `()` instead of `!`. 
https://github.com/extendr/libR-sys/issues/144#issuecomment-1455256822

extendr-api/src/thread_safety.rs
```diff
@@ -77,23 +77,17 @@ where
...
-            }
-            unreachable!("handle_panic unreachable")
-        }
+        Err(_) => unsafe { libR_sys::Rf_error(err_str.as_ptr() as *const std::os::raw::c_char) },
     }
 }
 
 static mut R_ERROR_BUF: Option<std::ffi::CString> = None;
 
-pub fn throw_r_error<S: AsRef<str>>(s: S) {
```
multimeric, line 91: Isn't `never` unstable? https://doc.rust-lang.org/std/primitive.never.html

CGMossa: Yes, but this is allowed, see this https://doc.rust-lang.org/reference/types/never.html

multimeric: Okay, so it's considered a "diverging function" which is what we called these functions before `never` was introduced and subsequently un-introduced: https://doc.rust-lang.org/rust-by-example/fn/diverging.html.

multimeric (approved):

yutannihilation:
Hmm, no idea, sorry. I need to figure out what causes this difference first.

https://github.com/extendr/libR-sys/issues/144#issuecomment-1455256822

CGMossa:
My number one suspect is which clang version is being used to generate the bindings.
If I look at the bindings:
```
bindings\bindings-windows-x86_64-R4.2.rs:
  340  extern "C" {
  341:     pub fn Rf_error(arg1: *const ::std::os::raw::c_char, ...) -> !;
  342  }
  ```
  Alright, my PC is on this platform, so when I generate bindings using the `--features use-bindgen`, I produce a file named `mybindings\bindings-windows-x86_64-R4.2.rs` and this file contains:
  ```
extern "C" {
    pub fn Rf_error(arg1: *const ::std::os::raw::c_char, ...);
}
```

yutannihilation:
> My number one suspect is which clang version is being used to generate the bindings.

What's curious is that even on a same platform it generates different signature. Windows might be affected by the change on Rtools, so let's focus on Linux for example.

```
bindings/bindings-linux-x86_64-R3.6.rs:    pub fn Rf_error(arg1: *const ::std::os::raw::c_char, ...);
bindings/bindings-linux-x86_64-R4.0.rs:    pub fn Rf_error(arg1: *const ::std::os::raw::c_char, ...);
bindings/bindings-linux-x86_64-R4.1.rs:    pub fn Rf_error(arg1: *const ::std::os::raw::c_char, ...) -> !;
bindings/bindings-linux-x86_64-R4.2.rs:    pub fn Rf_error(arg1: *const ::std::os::raw::c_char, ...) -> !;
bindings/bindings-linux-x86_64-R4.3-devel.rs:    pub fn Rf_error(arg1: *const ::std::os::raw::c_char, ...);
```

The bindings for R 3.6 and R 4.0 were generated with an older version before I bumped the version of bindgen, so it should matter (actually, [bindgen started to support no return attributes](https://github.com/rust-lang/rust-bindgen/pull/2259) only half a year ago). But, it's weird that the signature of R 4.3 is different than the ones of R 4.1 and R 4.2. All of these three versions keep updated by our CI.

I suspect this might be related to the update of C++ standards that R 4.2 supports. As I'm not a C/C++ person, I'm not very confident, of course.

yutannihilation:
This might matter, maybe?

https://github.com/r-devel/r-svn/commit/1831a08c4848475a4554a80d4352d1016a548b98

yutannihilation:
It seems bindgen cannot handle `_Noreturn`?

``` c
__attribute__((noreturn)) void foo1();
_Noreturn                 void foo2();
```

``` console
❯ bindgen tmp.h
/* automatically generated by rust-bindgen 0.64.0 */

extern "C" {
    pub fn foo1() -> !;
}
extern "C" {
    pub fn foo2();
}
```

yutannihilation:
```
❯ bindgen tmp.h --enable-function-attribute-detection
/* automatically generated by rust-bindgen 0.64.0 */

extern "C" {
    pub fn foo1() -> !;
}
extern "C" {
    pub fn foo2() -> !;
}
```

yutannihilation:
Using `.enable_function_attribute_detection()` fixes some of the signatures,

https://github.com/extendr/libR-sys/compare/master...yutannihilation:tmp1

but it's still less than when I manually define `NORET`.

https://github.com/extendr/libR-sys/compare/master...yutannihilation:tmp2

No idea which is better (and even what the first option does)...

yutannihilation:
I now feel defining `NORET` by ourselves because okay as we only use it for generating the bindings anyway.

https://github.com/extendr/libR-sys/pull/148

CGMossa:
This can now be merged, as extendr/libR-sys#148 has been merged and this PR passes CI.

PR #513, opened 2023-03-08
Add private items to developer docs

CGMossa:
This PR adds private items to the generated docs on our website.
Plus there is a feature now called `full-support`, that includes on source-code-additive features,
such that we can have that for docs-generation; 
Before, `"libR-sys/use-bindgen"` was included. This feature is really not additive, and while it is vital for compilation,
it isn't supposed to be included.

Reasons for inclusion of private items:

* Lately on discord users have references private items in ExtendR, that they might want to use or
learn from. These should be discoverable atleast on our website.
* I use the website docs to learn about ExtendR myself, as there are aspects of it that is not known to me.
Plus I intend to document *all items* including the private ones for the betterment of this project.
Having somewhere where this is shown and could be readily accessed is a good.

I hope that convinces you. Thanks!

### DIFF
--- a/extendr-api/Cargo.toml
+++ b/extendr-api/Cargo.toml
@@ -34,6 +34,8 @@ rstest = "0.16.0"
 [features]
 default = []
 
+full_support = ["graphics", "either", "ndarray", "num-complex", "serde"]
+
 # libc is needed to allocate a DevDesc (c.f., https://bugs.r-project.org/show_bug.cgi?id=18292)
 graphics = ["libc"]
 
--- a/extendr-api/Cargo.toml
+++ b/extendr-api/Cargo.toml
@@ -50,5 +50,5 @@ tests-graphics = ["tests-minimal", "graphics"]
 tests-all = ["tests", "graphics"]
 
 [package.metadata.docs.rs]
-features = ["graphics", "either", "ndarray", "num-complex", "serde", "libR-sys/use-bindgen"]
+features = ["graphics", "either", "ndarray", "num-complex", "serde"]
 
--- a/extendr-api/Cargo.toml
+++ b/extendr-api/Cargo.toml
@@ -34,7 +34,7 @@ rstest = "0.16.0"
 [features]
 default = []
 
-full_support = ["graphics", "either", "ndarray", "num-complex", "serde"]
+full-support = ["graphics", "either", "ndarray", "num-complex", "serde"]
 
 # libc is needed to allocate a DevDesc (c.f., https://bugs.r-project.org/show_bug.cgi?id=18292)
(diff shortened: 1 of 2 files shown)


extendr-api/Cargo.toml
```diff
@@ -34,6 +34,8 @@ rstest = "0.16.0"
 [features]
 default = []
 
+full-support = ["graphics", "either", "ndarray", "num-complex", "serde"]
```
multimeric, line 37: Choice of words matters here. I think we need to be 
```suggestion
# This dummy feature enables all features that increase the functionality of 
# extendr, via conversions or R features. Features that change behaviour
# but do not add functionality (such as `libR-sys/use-bindgen`) are excluded
full-functionality = ["graphics", "either", "ndarray", "num-complex", "serde"]
```

extendr-api/Cargo.toml
```diff
@@ -50,5 +52,4 @@ tests-graphics = ["tests-minimal", "graphics"]
 tests-all = ["tests", "graphics"]
 
 [package.metadata.docs.rs]
-features = ["graphics", "either", "ndarray", "num-complex", "serde", "libR-sys/use-bindgen"]
-
+features = ["graphics", "either", "ndarray", "num-complex", "serde"]
```
multimeric, line 55: We need bindgen for the docs to build. Recall #426.
```suggestion
features = ["full-functionality", "libR-sys/use-bindgen"]
```

multimeric (changes requested):

extendr-api/Cargo.toml
CGMossa: Call it `all-extentions`?

extendr-api/Cargo.toml
```diff
@@ -34,6 +34,11 @@ rstest = "0.16.0"
 [features]
 default = []
 
+# This dummy feature enables all features that increase the functionality of 
+# extendr, via conversions or R features. Features that change behaviour
+# but do not add functionality (such as `libR-sys/use-bindgen`) are excluded
+full-functionality = ["graphics", "either", "ndarray", "num-complex", "serde"]
```
Ilia-Kosenkov, line 40: I think this should be `full-support`, no? Since you reference `--features full-support` in CI script

multimeric: I asked Mossa to rename it to `full-functionality`. I just missed updating it in the CI script.

CGMossa:
This PR was started due to a suspicion of a bug, but it adds a clear "additive" feature plus private part of extendr is visible in our documentation, which are two small things, that can still be regarded as a net-positive.

multimeric (approved):

PR #515, opened 2023-03-10
Revert "Minor warning clean-up"

CGMossa:
Reverts extendr/extendr#512

### DIFF
--- a/extendr-api/src/error.rs
+++ b/extendr-api/src/error.rs
@@ -11,6 +11,7 @@ pub fn unwrap_or_throw<T>(r: std::result::Result<T, &'static str>) -> T {
     match r {
         Err(e) => {
             throw_r_error(e.to_string());
+            unreachable!("");
         }
         Ok(v) => v,
     }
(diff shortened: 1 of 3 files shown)


CGMossa:
I merged a PR that passed CI completely in the PR, but on the branch once merged that is, it no longer passes it. I may know why, but now I need to revert what was merged.

multimeric:
I can't parse that sentence, but anyway if you can find an easy fix I don't mind if we skip the reversion and go straight for the fix.

CGMossa:
I don't understand what is wrong here. I don't understand why reverting the PR doesn't pass CI. I'm completely lost now.

yutannihilation:
I think this setup is wrong. `Sys.setenv()` is effective on the same session, so this doesn't work on the next step.

https://github.com/extendr/extendr/blob/b5a5350541cd2d219a8732a017732e052bc1cc1a/.github/workflows/test.yml#L257-L274

yutannihilation:
Confirmed. 

https://github.com/extendr/extendr/pull/517

CGMossa:
Since @yutannihilation fixef this in #515, closing this.

I don't know if we should be testing with rextendr either.

PR #519, opened 2023-03-11
[Breaking changes]: Optional features overhaul

Ilia-Kosenkov:
This PR reworks how some of the optional features are exposed:
- Modules `optional` and `either` get their docs in order
- Module `robj_nadrray` is renamed to `ndarray` and placed under `optional`
- `ndarray` crate is no longer re-exported

### DIFF
--- a/extendr-api/src/optional/either.rs
+++ b/extendr-api/src/optional/either.rs
@@ -1,3 +1,36 @@
+/*!
+Enables support for [`either`](https://docs.rs/either/latest/either/) crate, allowing accepting and returning `Either<L,R>` values if both `L` and `R` are convertible to/from `Robj`.
+
+`either` crate support is currently available in the dev version of `extendr-api` and requires enabling `either` feature:
+```toml
+[dependencies]
+extendr-api = { git = "https://github.com/extendr/extendr" , features = ["either"] }
+```
+
+Accepting `Either<L,R>` as an input requires enabling `extendr` option `use_try_from = TRUE`:
+
+```rust
+use extendr_api::prelude::*;
+use either::Either;
+
+#[extendr(use_try_from = TRUE)]
+fn accept_numeric(input : Either<Integers, Doubles>) {}
+```
+
+Here is an example of `either` usage -- a type-aware sum:
+```rust
+use extendr_api::prelude::*;
+use either::Either::{self, Left, Right};
+
+#[extendr(use_try_from = TRUE)]
+fn type_aware_sum(input : Either<Integers, Doubles>) -> Either<Rint, Rfloat> {
+    match input {
+        Left(ints) => Left(ints.iter().sum::<Rint>()),
+        Right(dbls) => Right(dbls.iter().sum::<Rfloat>()),
+    }
+}
+```
+*/
 use crate::{Error, Robj};
 use either::Either::{self, Left, Right};
 
--- a/extendr-api/src/optional/either.rs
+++ b/extendr-api/src/optional/either.rs
@@ -1,5 +1,5 @@
 /*!
-Enables support for [`either`](https://docs.rs/either/latest/either/) crate, allowing accepting and returning `Either<L,R>` values if both `L` and `R` are convertible to/from `Robj`.
+Enables support for the [`either`](https://docs.rs/either/latest/either/) crate, to allow accepting and returning `Either<L, R>` values if both `L` and `R` are convertible to/from `Robj`.
 
 `either` crate support is currently available in the dev version of `extendr-api` and requires enabling `either` feature:
 ```toml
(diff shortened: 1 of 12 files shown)


Ilia-Kosenkov:
So if this PR is green-lit, I'll make the necessary changes to `{rextendr}`. Right now it fails integration tests because of the breaking changes in import mechanics.

multimeric:
* Optional as a module seems sensible now, but imagine a world where we refactor all of `extendr` into optional modules, and only the core `Robj` is default. Does an optional module still make sense? Possibly `either` and `ndarray` are more uniquely defined by the fact that they use external crates?
* Agreed with the `robj_ndarray` change
* Why don't you like re-exporting `ndarray`? It seems like it avoids a certain class of errors where the user's "local" `ndarray` is a different, incompatible version. [My very first issue with Rust back in 2021](https://stackoverflow.com/questions/68443825/how-do-i-use-ndarray-statscorrelationext-on-a-polarspreludedataframe) involved this type of error!

multimeric:
Yep, I'm pretty convinced re-exporting is a good practise. This page explains it in more detail: https://www.lurklurk.org/effective-rust/re-export.html

Ilia-Kosenkov:
1. Integrations with third party crates, I suspect, will not be that numerous. Maybe optional should be renamed into something different. In general it simplifies managing features since all features with dependencies (minus Serde for now) are in one module that you include and reference unconditionally instead of running around searching for all features. For instance, if we want to enable re-export, I will do it in one place exactly.

2. -

3. Noted. I have no strong opinion and since it is a regular practice, I have no objections. This will actually simplify a lot of code for me in `{rextrndr}` if we guarantee to always re-export things.

multimeric:
> Integrations with third party crates, I suspect, will not be that numerous. Maybe optional should be renamed into something different. In general it simplifies managing features since all features with dependencies (minus Serde for now) are in one module that you include and reference unconditionally instead of running around searching for all features. For instance, if we want to enable re-export, I will do it in one place exactly.

Right, but you will always have a self-contained module for `either`, I'm just questioning why it needs to be another level down inside the `optional` submodule.

Ilia-Kosenkov:
Well, the goal is too organize project into smaller modules rather than having huge 1k-lines monstrosities when you never ever can find what you want unless you do a full text search. Here we have either module dealing only with either and maybe some doc tests. Next to it is ndarray dealing only with ndarray, some doc tests and some unit tests. That's it. If tomorrow I will be working on either and you - on ndarray, our changes will likely never overlap, which definitely makes project more maintainable.

If we csn achieve the same file structure without the modules, I am all ears.

multimeric:
No, I still like `extendr` and `ndarray` modules! I just think an `optional` module is not needed in the long run.

Ilia-Kosenkov:
Ok, I hear your point. One concern I have (if I understand it correctly) that reexported crate (say, `ndarray`) is accessible as `extendr_api::ndarray` (judging by the example in the article you linked above). So if you include *our own* module `ndarray` in the top namespace, how will it distinguish one from another? In the current structure, all things `ndarray` related are under `extendr_api::optional::ndarray`, which makes sense to me.

Anyway, let's try out with reexport & keeping `optional` module for now, we can later change things.

extendr-api/src/optional/either.rs
```diff
@@ -1,5 +1,36 @@
+/*!
+Enables support for [`either`](https://docs.rs/either/latest/either/) crate, allowing accepting and returning `Either<L,R>` values if both `L` and `R` are convertible to/from `Robj`.
```
multimeric, line 2:
```suggestion
Enables support for the [`either`](https://docs.rs/either/latest/either/) crate, to allow accepting and returning `Either<L, R>` values if both `L` and `R` are convertible to/from `Robj`.
```

extendr-api/src/optional/either.rs
```diff
@@ -1,5 +1,36 @@
...
+
+`either` crate support is currently available in the dev version of `extendr-api` and requires enabling `either` feature:
+```toml
+[dependencies]
+extendr-api = { git = "https://github.com/extendr/extendr" , features = ["either"] }
+```
+
+Accepting `Either<L,R>` as an input requires enabling `extendr` option `use_try_from = TRUE`:
```
multimeric, line 10:
```suggestion
Accepting `Either<L, R>` as an input requires enabling the `extendr` option `use_try_from = TRUE`:
```

multimeric (approved):

extendr-api/src/optional/either.rs
Ilia-Kosenkov: I am not convinced '*to allow*' is grammatically correct here. To me, this reads '*enables smth, to allow smth*', while I'd expect either '*enable smth, allowing smth*' or '*enables smth and allows smth*'.

multimeric: > To me, this reads 'enables smth, to allow smth'

This sounds fine to me. It's short for "This module enable**s** support for the either crate, to allow...". Does that sound better to you?

> 'enable smth, allowing smth' 

"This module enable something"? No that's not right, unless you mean it in the imperative sense, but that doesn't make sense either because we aren't asking the user to do anything.

> 'enables smth and allows smth'

Gramatically correct, but misses the causative relationship between using this module and accepting `Either`

Your original "allowing accepting and returning" was also gramatically correct but I try to avoid two participles like that. It's hard to follow.

Issue #528, opened 2023-04-13
Raw identifiers act as a comment in the `extendr-wrappers.R` file


bicarlsen:
If a [raw identifier](https://doc.rust-lang.org/rust-by-example/compatibility/raw_identifiers.html) is used as a function argument, the `#` is copied in the `extendr-wrappers.R` file, acting as a comment.
e.g.
```rust
#[extendr]
fn my_fcn(r#type: i32) -> i32 {
  ...
}

extendr_module! {
  mod my_lib;
  my_fcn;
}
```
results in the output line `my_fcn <- function(r#type) .Call(wrap__my_fcn, r#type)` in the `extendr-wrappers.R` file.

Ilia-Kosenkov:
Hi! Thanks for reporting this. Yep, I can totally see why it happens like this. We will fix it soon-ish.

CGMossa:
I'm sure @Ilia-Kosenkov will help you out in a minute. Note if you want to chat with us or ask informally about things, some of us hang out on the project Discord sever.

Ilia-Kosenkov:
Oh, cool. I already researched the issue and prototype some fixes, but I am more than happy to guide you through the process!
Here is a rough outline of the first steps that you need to do (let's pretend we follow some TDD):

1. Get the extendr repo locally, go to `tests/extendrtests` and ensure you can test this R project
  - By test I mean you can execute `devtools::test()` and all things pass (or do it within Rstudio or any other tool that you use)
  - There is a caveat -- it may fail because it cannot find extendr-api. You need to tweak `Cargo,toml` in `tests/textendrtest/src/rust` so it points to your local `extendr-api` folder (we use relative paths there, you likely need to add/remove a couple of `../` until it matches the path). That way you will be able to test the changes you make to the `extendr-*` projects
2. Once current tests work, add a new rust module in this test project and write there methods that have `r#` raw identifiers. Start with something similar to your example (but give it better test names).
  - Don't forget to reference this new module in `tests/extendrtests/src/rust/src/lib.rs` so you get some bindings generated for it
3. Run `rextendr::document()` over `tests/extendrtest`. It should fail with an error similar to the one above -- the generated bindings will contain invalid R syntax.

CGMossa:
It's worth it for us to get you up and running lol.
You're more than welcome here.

Ilia-Kosenkov:
I agree that if you want to get some deeper understanding of `extendr`, it is well worth of our time to walk you through.

[...]

PR #531, opened 2023-04-15
Fix bug with raw identifiers in function arguments (#528)

bicarlsen:
Added functionality to remove the raw identifier prefix (`r#`) from function arguments in the R bindings.


Ilia-Kosenkov:
Hey, thanks for the contribution. There a couple of edge cases we need to cover, but right now I do not have time to comment on this. Can you hold on to this PR for a little bit longer?

bicarlsen:
Of course, just let me know.

Ilia-Kosenkov:
Ok so I was thinking about touching this thing: https://github.com/extendr/extendr/blob/8dc13d8ba87818e010542a4ac6977b5281dd98f5/extendr-api/src/metadata.rs#L153-L159

We have a method that sanitizes rust identifiers on the R side. E.g., if you have a `fn my_fn(_x :i32) {}`, `R` cannot handle the `_x` argument, but it instead can handle ``` `_x` ```, which we effectively pass around when we detect an incompatible name. I propose to add another check, like `.contains('#')`, instead of changing generated code. We just sanitize wrappers properly. 

There is another thing, we should also consider a case of functions that have `r#` in the name, I suggest we also add a `r#true()` function to the test package. Now if `sanitize_warppers()` is updated, `r#true()` will be wrapped accordingly, but not the argument passed to the `.Call()` (here it is something like `wrap__mypkg_r#true`, which is invalid). This part I have not yet solved, but we need to also sanitize it the same way we sanitize arguments -- ``` `wrap__mapkg_r#true` ```  is a valid R identifier.

This way we preserve exact 1-to-1 mapping between rust names and R names.

bicarlsen:
I updated the implementation to incorporate the `sanitize_identifiers` method mentioned by @Ilia-Kosenkov above. This did not address the wrapped function names in the `extendr-wrappers.R` file though, so still needed sanitization in the `extendr_macros/wrappers.rs` module.

Ilia-Kosenkov:
Oh shoot, completely forgot about this, my apologies. I will check it today.

Ilia-Kosenkov:
I think you need to merge master in your branch to update CI

bicarlsen:
I don't understand why the Markdown links are causing errors in the tests. This issue was discussed in #534 . I fixed the  typos with the parentheses, so not sure what is causing the issue now.

tests/extendrtests/R/extendr-wrappers.R
```diff
@@ -89,6 +89,23 @@ euclidean_dist <- function(a) .Call(wrap__euclidean_dist, a)
 
 type_aware_sum <- function(input) .Call(wrap__type_aware_sum, input)
 
+#' Test raw identifiers (`r#`) in function arguments are parsed correctly.
+#' See (Issue #582)[https://github.com/extendr/extendr/issues/528] for details.
```
CGMossa, line 93:
```suggestion
#' See [Issue #582](https://github.com/extendr/extendr/issues/528) for details.
```

tests/extendrtests/R/extendr-wrappers.R
```diff
@@ -89,6 +89,23 @@ euclidean_dist <- function(a) .Call(wrap__euclidean_dist, a)
...
+#' @param type : i32 or `NULL`
+#' @export
+raw_identifier_in_fn_args <- function(type = NULL) .Call(wrap__raw_identifier_in_fn_args, type)
+
+#' Test raw identifiers (`r#`) as function names are parsed correctly.
+#' See (Issue #582)[https://github.com/extendr/extendr/issues/528] for details.
```
CGMossa, line 99:
```suggestion
#' See [Issue #582](https://github.com/extendr/extendr/issues/528) for details.
```

tests/extendrtests/R/extendr-wrappers.R
```diff
@@ -89,6 +89,23 @@ euclidean_dist <- function(a) .Call(wrap__euclidean_dist, a)
...
+#' @export
+true <- function() .Call(wrap__true)
+
+#' Combine raw identifiers (`r#`) as a function name and in arguments are parsed correctly.
+#' See (Issue #582)[https://github.com/extendr/extendr/issues/528] for details.
```
CGMossa, line 104:
```suggestion
#' See [Issue #582](https://github.com/extendr/extendr/issues/528) for details.
```

CGMossa (changes requested):
There are more places where the markdown-links are done incorrectly. But here you go!
Just accept these suggestions.

bicarlsen:
Geeze! Thanks @CGMossa.

CGMossa:
@bicarlsen Right, you need to run the `document()` function again, as to regenerate `Rd`-files with the right markdown link notation in them.

bicarlsen:
Okay, I was having trouble running the tests on my local machine, but  noticed that I had to change the [`tests/extendrtests/src/rust/Cargo.toml` > `extendr-api`](https://github.com/extendr/extendr/blob/fc677380fa03c638e7913fdd96140da9ff1a5ad4/tests/extendrtests/src/rust/Cargo.toml#L23) to have an absolute path. I am still receiving an error
```
── Error ('test-graphic-device.R:2:5'): `my_device()` works as expected ────────
Error in `my_device("foo")`: Graphics API version mismatch
Backtrace:
     ▆
  1. ├─testthat::expect_output(...) at test-graphic-device.R:2:4
  2. │ └─testthat:::quasi_capture(...)
  3. │   ├─testthat (local) .capture(...)
  4. │   │ └─testthat::capture_output_lines(code, print, width = width)
  5. │   │   └─testthat:::eval_with_output(code, print = print, width = width)
  6. │   │     ├─withr::with_output_sink(path, withVisible(code))
  7. │   │     │ └─base::force(code)
  8. │   │     └─base::withVisible(code)
  9. │   └─rlang::eval_bare(quo_get_expr(.quo), quo_get_env(.quo))
 10. └─extendrtests::my_device("foo")
 ```
but am not sure why as I didn't touch that code and it seems unrelated to this PR.

**Comment by CGMossa** [comment]
This happened to me the other day as well...
Run an rextendr colon colon clean() and try again.

**Review by CGMossa — COMMENTED** [review]
Alright, that part is resolved. Did you have success with running tests after running `rextendr::clean()`.

**Comment by Ilia-Kosenkov** [comment]
I would say looks good. We have tests to cover these edge cases, everything seems to be in place. Please finalize the change request discussion with @CGMossa and we can proceed.

Oh and BTW, please write a note in the changelog.

**Comment by bicarlsen** [comment]
After a `rextendr::clean()` I still receive the error
```
Error (test-graphic-device.R:2:5): `my_device()` works as expected
Error in `my_device("foo")`: Graphics API version mismatch
Backtrace:
  1. testthat::expect_output(...)
       at test-graphic-device.R:2:4
 10. extendrtests::my_device("foo")
```

**Comment by bicarlsen** [comment]
@Ilia-Kosenkov I browsed the `CHANGELOG` but had a difficult time understanding how exactly to add an entry to it in the correct format. Could you give some insight?

**Comment by CGMossa** [comment]
> After a `rextendr::clean()` I still receive the error
> 
> ```
> Error (test-graphic-device.R:2:5): `my_device()` works as expected
> Error in `my_device("foo")`: Graphics API version mismatch
> Backtrace:
>   1. testthat::expect_output(...)
>        at test-graphic-device.R:2:4
>  10. extendrtests::my_device("foo")
> ```

When I had this error, I couldn't even do `plot.new()` in R. Are you able to do that? Please try to restart your session as well.

**Comment by bicarlsen** [comment]
Restarted the session a few times, `clean`ed and re`document`ed a few times, still the same error. Plotting seems to work for me though.

**Comment by Ilia-Kosenkov** [comment]
Summoning @yutannihilation . You have experience with graphic devices, why are we getting version mismatch here (when executed locally)? Can it be that our bindings are for `R 4.2.3` and are labeled `4.2`, so if you now use `R 4.2.x`, the bindings are resolved to `4.2`, but there still is a mismatch (i.e. if R people bumped graphics version in patch)/

**Comment by Ilia-Kosenkov** [comment]
We (are trying to) maintain CHANGELOG according to [these rules](https://keepachangelog.com/en/1.0.0/). In your case, you need to go to the top of the file and add under `unreleased/fixed` section something that describes your changes. Make it human readable, so it goes something like:
```
[Fixed]
- <existing entry>
- A bug preventing using raw rust literals in exported extendr functions [PR link]
```

Use present tense and simple wording. Give an example of a raw literal (like `r#type`) in and how it is converted to `type` now. I would say 1-2 sentences are enough.

yutannihilation:
> R people bumped graphics version in patch

It can happen.

https://github.com/wch/r-source/blob/ed2b4252a103c2f8494b906c404fbc9a8cd2d5ef/src/include/R_ext/GraphicsEngine.h#L81

But, I guess it's not the case (unless you are using R 4.1). Let me try reproducing it on my local.

bicarlsen:
Shall I push wish the new `CHANGELOG` or wait to resolve the graphics issue?

CGMossa:
Push the changes. Gives us time to review that.

yutannihilation:
Okay, I reproduced it. I think you can just remove your `tests/extendrtests/src/rust/Cargo.lock` to update the dependency crates.

bicarlsen:
> Okay, I reproduced it. I think you can just remove your `tests/extendrtests/src/rust/Cargo.lock` to update the dependency crates.

Yup, that did the trick :)

bicarlsen:
Is there anything else to be done here before we can merge?

CGMossa (approved):
LGTM

CGMossa:
I had already seen this and thought @Ilia-Kosenkov wanted to give it a last look and then submission. Looks great from my end.

PR #153, opened 2021-01-25
Add invisible() to calls that return nothing.

clauswilke:
Closes #148.

PR #535, opened 2023-04-23
Update CI settings for R 4.3

yutannihilation:
I think we'll want to reflect some of the recent changes of libR-sys' CI settings by #153, but let me keep this pull request as minimal. Let's do it later.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -171,14 +171,14 @@ jobs:
       - name: Install dependencies for extendrtests and rcmdcheck
         uses: r-lib/actions/setup-r-dependencies@v2
         with:
-          cache-version: 3
+          cache-version: 4
           working-directory: tests/extendrtests
           extra-packages: any::rcmdcheck
 
       - name: Install dependencies for rextendr
         uses: r-lib/actions/setup-r-dependencies@v2
         with:
-          cache-version: 3
+          cache-version: 4
           working-directory: tests/rextendr
 
       # Regex is used to inject absolute path into Cargo.toml
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,7 +28,6 @@ jobs:
         config:
           - {os: windows-latest, r: 'release', rust-version: 'stable-msvc', rtools-version: '43'}
           - {os: windows-latest, r: 'devel',   rust-version: 'stable-msvc', rtools-version: '43'}
-          - {os: windows-latest, r: 'release', rust-version: 'stable-gnu',  rtools-version: '43'}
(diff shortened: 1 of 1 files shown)


yutannihilation:
This fails until https://github.com/extendr/libR-sys/pull/164 gets merged (the current default bindings point to R 4.2).

CGMossa:
Are you sure it is the right PR in the original post? I don't understand the connection.

I'll greenlight the PR in question now.

I'm still a little puzzled by that no fix to the bindings being passed after being generated have been proposed yet.
But I'm following along as much as I can!

yutannihilation:
Yes, I'm pretty sure. As you can see, the error message says:

```
Error in `my_device("foo")`: Graphics API version mismatch
```

This means the API version embedded in the header file doesn't match with the version of the R session.

yutannihilation:
Thanks for the approval! All green now.

.github/workflows/test.yml
```diff
@@ -26,18 +26,10 @@ jobs:
...
-          #   detect when something gets broken. 
-          # * for R < 4.2, the MSVC toolchain must be used to support
-          #   cross-compilation for the 32-bit. 
-          - {os: windows-latest, r: 'release', rust-version: 'stable-msvc', rtools-version: '42'}
-          - {os: windows-latest, r: 'devel',   rust-version: 'stable-msvc'}
-          - {os: windows-latest, r: 'release', rust-version: 'stable-gnu',  rtools-version: '42'}
-          - {os: windows-latest, r: '4.1',     rust-version: 'stable-msvc'}  # TODO: Remove this runner when we drop the support for R < 4.2
+          - {os: windows-latest, r: 'release', rust-version: 'stable-msvc', rtools-version: '43'}
+          - {os: windows-latest, r: 'devel',   rust-version: 'stable-msvc', rtools-version: '43'}
+          - {os: windows-latest, r: 'release', rust-version: 'stable-gnu',  rtools-version: '43'}
```
Ilia-Kosenkov, line 31: What's the point in supporting `gnu` now? I dropped it from `{rextendr}` CI.

yutannihilation: I don't remember the reason why we included it, so it should be fine.

CGMossa: Can we keep one gnu test for health?

Ilia-Kosenkov: We used it because we made a switch from `gnu` to `msvc` and kept it for 'compatibility', though all of our workflows were using `msvc`

Ilia-Kosenkov (approved):

yutannihilation:
> Can we keep one gnu test for health?

Why? If you feel we should eagerly support the GNU toolchain, let's discuss in another issue. I'm merging this for now.

PR #540, opened 2023-04-27
Address collection of too many `Robj`s

CGMossa:
This has been discussed extensively: #397, and other places.

PR does two things:

* Addresses changes in the latest version of libR-sys, that might have broken `extendr`.
* Refactors `FromIterator`-impls for `Strings` and `List`. R cannot handle protecting more
than 10k elements at a time, but protecting one element, that contain more than 10k is no issue.
Thus intermediate objects are no longer pre-allocated into `Robj`, and then assigned to `Strings` / `List` resulting `Robj`.
Instead immediate objects are written directly into the object.

Furthermore there was an inconsistency before where `Rf_unprotect` was set to `n`, but it would
incidentally unprotect the `result` Robj, and leaving the first intermediate `Robj`.

This code might have affected performance, and thus benchmarking this should be done in a follow-up PR.
Problem is this **didn't work for some cases**, and thus it is more important to make it functional, than say blazingly fast.

## Follow-up PR
- [ ] Currently, there is no way to get an unprotected `Robj` from a rust-compatible type.
We should do that, as the batched allocations in `FromIterator` shouldn't use `ownership`-module
but rather rely on allocating and unprotecting its own intermediaries. 
- [x] Add a gc-torture test to see if our protection mechanisms work.


CGMossa:
Context: Turns out this didn't have anything to do with extendr's `ownership` module: Using r-bindings directly, i.e. without extendr:
```rust
  let total_protected = 10_000 * 6;
  let dummy_string = "Jamie Foxx";
  for _p in 0..total_protected {
    unsafe {
      Rf_protect(Rf_mkCharLenCE(
        dummy_string.as_ptr() as _,
        dummy_string.len() as _,
        cetype_t::CE_UTF8,
      ));
```
This would break.

JosiahParry:
Please see [Discord thread](https://discord.com/channels/751419551132155925/1101214791688785930) for reproducible examples

CGMossa:
I would also like to change this:
https://github.com/extendr/extendr/pull/398#discussion_r847264770

I've removed it and the test that was present in 
extendr\extendr-api\tests\issue-397-memory-allocation.rs
works fine with this PR.

CGMossa:
Suggested edit:
```diff
diff --git a/extendr-api/src/ownership.rs b/extendr-api/src/ownership.rs
index 759303cc..a70a8fe0 100644
--- a/extendr-api/src/ownership.rs
+++ b/extendr-api/src/ownership.rs
@@ -158,7 +158,7 @@ impl Ownership {
     // Garbage collect the tracking structures.
     unsafe fn garbage_collect(&mut self) {
         // println!("garbage_collect {} {}", self.cur_index, self.max_index);
-        let new_size = self.cur_index * 2 + EXTRA_PRESERVATION_SIZE;
+        let new_size = self.cur_index + EXTRA_PRESERVATION_SIZE;
         let new_sexp = Rf_allocVector(VECSXP, new_size as R_xlen_t);
         R_PreserveObject(new_sexp);
         let old_sexp = self.preservation as SEXP;

```

multimeric:
I think this works because of:
> Protecting an R object automatically protects all the R objects pointed to in the corresponding SEXPREC, for example all elements of a protected list are automatically protected.
https://cran.r-project.org/doc/manuals/R-exts.html#Garbage-Collection

Although there is a point at which we have to PROTECT each newly-created Robj before it gets put into the List, and I'm worried that we're not doing that.

multimeric:
From reading `R-exts`s, it sounds like it might be okay to not PROTECT each object as long as we can guarantee that no memory allocations or anything that might trigger the GC happens before it gets put into the list.

CGMossa:
I'm immediately passing the intermediate objects to a protected vector. Isn't that good enough?
Note that the resulting vector is protected by the ownership module.

multimeric:
I think it's okay as long as you can guarantee that the GC **will not** run between the point of you creating the `Robj` and the point of you putting it in the list. Even `SET_VECTOR_ELT` might somehow force GC.

CGMossa:
Right. For the List version I need to protect, assign to the list, and unprotect(1).

multimeric:
Yeah, I think so? It might incur a small performance penalty but it will be safer. Also it won't be anywhere near as bad as the old way because we do it one at a time.

CGMossa:
@multimeric I don't think I need to do that.
This line `let item: Robj = v.into();` presumably invokes
```rust
impl<T> From<T> for Robj
where
    T: ToVectorValue,
```
Which takes the scalar, wraps it in something that is `IntoIterator`, and then calls `.collect_robj()`.
This in turn calls `fixed_size_collect` which uses `Robj::alloc_vector` [extendr `ownership` module]

The only "bad" thing about the `List`'s `FromIterator` impl is that it would be nice if it sidesteps
`ownership`-module as to not clog up the rust gc with something intermediary. 
If you guys can suggest something now for that, then we can get in, otherwise I believe that's better suited for
a follow-up PR.

extendr-api/tests/from_iterator_tests.rs
```diff
@@ -0,0 +1,59 @@
...
+#[test]
+fn test_with_gc_torture() {
+    test!(
+        let x = vec![1, 4, 5, 6];
+        R!("gctorture(on = TRUE)")?;
+        let list: List = x.into_iter().collect();
+        R!("gctorture(on = FALSE)")?;
+        assert_eq!(list, list!(1, 4, 5, 6));
+    );
+}
```
multimeric, line 46: One improvement I would like here. I've noticed that I don't encounter segfaults unless the vector allocations are large (https://stat.ethz.ch/pipermail/r-devel/2023-April/082589.html). Could you try making a list of large vectors (n>100) instead of just scalars?

CGMossa: Good idea. I'll add that later.

extendr-api/src/wrapper/list.rs
```diff
@@ -378,20 +378,17 @@ impl<T: AsRef<str>> KeyValue for (T, Robj) {
...
+        let iter_collect: Vec<_> = iter.into_iter().collect();
+        let len = iter_collect.len();
 
-            let len = values.len();
+        crate::single_threaded(|| unsafe {
             let robj = Robj::alloc_vector(VECSXP, len);
-            for (i, v) in values.into_iter().enumerate() {
-                SET_VECTOR_ELT(robj.get(), i as isize, v);
+            for (i, v) in iter_collect.into_iter().enumerate() {
+                let item: Robj = v.into();
```
multimeric, line 389:
```suggestion
                // We don't PROTECT each element here, as they will be immediately
                // placed into a list which will protect them:
                // https://cran.r-project.org/doc/manuals/R-exts.html#Garbage-Collection
                let item: Robj = v.into();
```

multimeric (approved):

extendr-api/src/wrapper/list.rs
CGMossa: This line yields a protected Robj by the ownership module.
This part is not controversial as everything is protected.
What isn't good is that it is clogging the ownership module.

CGMossa: I was going to make a follow-up PR, where we create an unprotected Robj, (that also need to use the rust-conversion thing we have setup), and then deal with that.. I forgot what the plan would be there...

JosiahParry:
I'm still running into this issue. I'd love to be able to collect more objects from lists. This is a limiting factor in the use of extendr. :/ 

Is there anything I can (try) to do to help this get merged?

JosiahParry:
FWIW to prevent this issue: one cannot collect into a list but rather should build the list iteratively using `List::new()` and `.set_elt()` 

See https://github.com/JosiahParry/sfconversions/commit/d3ad0f30445e9a473c1c6343300a19a416c168f4 for example of required refactor.

CGMossa:
I'm tending to the comments here now. For what it is worth, I believe `List::from_values` is also safe to use.

JosiahParry:
> I'm tending to the comments here now. For what it is worth, I believe `List::from_values` is also safe to use.

Would collecting into `Vec<T>` then `List::from_values()` cause an additional / unwanted allocations or anything that might bloat memory or slow it down?

CGMossa:
You're very correct. The idea would be to pass an interator to `from_values`, thus side-stepping the `Vec<T>` collecting / allocation.

CGMossa:
@Ilia-Kosenkov I've addressed the comments by @multimeric. I have also had this approved. Would you mind giving this a sanity check and merging?

Ilia-Kosenkov:
Looks ok to me, but let's resolve conflicts and run latest CI to be absolutely sure

yutannihilation:
> run latest CI to be absolutely sure

I think you'll probably hit this error, too.

https://github.com/extendr/extendr/pull/586#issuecomment-1637495135

CGMossa:
I ask @Ilia-Kosenkov to kindly give this a look and please merge it if you're satisfied.

Issue #549, opened 2023-05-04
implicit function param conversions leaks memory


sorhawell:
extendr functions/method allow conveniently to set types in function param, to skip explicitly converting any Robj into something else. E.g. below `x: Strings` implicit call something like `robj.as_strings()` under the hood. However any conversion error will activate `throw_r_error`

`throw_r_error()` is easy to use, but known for leaking mem #278 . The history of alternatives are long and there is a broad spectrum of different opinions/patterns in the community. 

extendr relies on `throw_r_error` in this function header conversion. E.g. in this function Robj is converted to Strings. "use_try_from" can be true or false
```rust
#[extendr(use_try_from = true)]
fn give_me_string_vec(x: Strings) -> String {
    format!("thank you for Strings with length {} ...", x.len())
}
```

In this case the leakage is benign, only 3 integers or some Altrep thereof. For use cases where input is small, the user would find status quo completely acceptable.
```R
> polars:::give_me_string_vec(1:3)
Error in polars:::give_me_string_vec(1:3) : Expected Strings got Integers
```

however this will cause a 2Gb leak, which will not be recovered until Rsession is restarted
```R
(function() {
  for (i in 1:100) {
    print(i)
    x = sample(1:10,1e7,replace = TRUE)
    tryCatch(polars:::give_me_string_vec(x), error = \(err) NULL)
  }
})()
```

work arounds is to take any input as `Robj` and use `try_into` and return a result (#473) or panic!. Then there is no leakage.

daniellga:
That's a problem... I use to avoid returning `Robj` so it's easier to identify the r type the function returns... Does the same happen if I use only `#[extendr]`?

sorhawell:
This is a very nice show case. You may need to try floats instead of ints, as the latter may be passed as a very lightweight AltRep.

sorhawell:
> That's a problem... I use to avoid returning `Robj` so it's easier to identify the r type the function returns... Does the same happen if I use only `#[extendr]`?

Everything is eventually converted into an Robj (or panic!) when return. I agree it looks prettier if signature not return Robj, but some more concrete type. But sometimes thats the most convenient, or if to explicit choose the conversion into Robj.

multimeric:
> This is a very nice show case. Maybe floats instead of ints, as the latter may be passed as a very lightweight AltRep.

[...]

Issue #278, opened 2021-09-03
Returning error messages to R


multimeric:
Is there an easy way to return strings from Rust that will become errors, warnings or messages in R? If I remember right, simply returning an Error will just panic since it just gets unwrapped.

multimeric:
Okay it looks like we have an existing wrapper which is `throw_r_error`: https://github.com/extendr/extendr/blob/2103979ab085376194b1f51a5a20b28e2f165691/extendr-api/src/thread_safety.rs#L91-L98

I note there's also an `Rf_warning` which can be used the same way as `Rf_error` for warnings too. There's also `R_ShowMessage`, but neither of these have a nice Rust wrapper so we would have to manually convert the string into a C string as in the above snippet.

andy-thomason:
This sounds like something to look at.

Would you like to try a PR or would you like one of us to give it a try?

multimeric:
If it's as simple as implementing a simple Rust wrapper like this for each of those APIs I'm happy to do it. 

By the way, we have this wrapper for returning errors, but afaik none of the `TryFrom` `Robj` implementations actually use this. So I'm guessing that means that errors in a Rust library will not return a user-friendly R error but will show a less friendly Rust error? Am I right?

multimeric:
Notes for future implementation:

* `R_ShowMessage` is somewhat explained here: https://cran.r-project.org/doc/manuals/R-exts.html#Setting-R-callbacks
* `Rf_warning` and `Rf_error` are explained here: https://cran.r-project.org/doc/manuals/r-release/R-ints.html#Warnings-and-errors

andy-thomason:
We did look at handling Result types in wrappers. It may still do so.

multimeric:
The thing that worries me is this: 
https://github.com/extendr/extendr/blob/7e9cac34187d333abd83b79b0e13887249fb7d42/extendr-api/src/robj/into_robj.rs#L37-L45

So if our Rust function returns a result, and that result is an error, then it will just get unwrapped in Rust. I guess the error message will get printed to the terminal maybe (?) but I suspect this isn't very user friendly. Also will this cause existing Rust objects in R to break, since I assume the memory will be reclaimed when the library crashes?

andy-thomason:
This only works because we catch panics and convert them to R errors.
Panics should unwind the stack, dropping objects along the way.
We should check that we have a test for this, however.

A nicer way would be to have a R class "RustError" or similar
[...]

PR #551, opened 2023-05-07
[PoC] Drop Error before throw_r_error()

yutannihilation:
https://github.com/extendr/extendr/issues/549#issuecomment-1536978399

### DIFF
--- a/extendr-api/src/error.rs
+++ b/extendr-api/src/error.rs
@@ -20,7 +20,9 @@ pub fn unwrap_or_throw<T>(r: std::result::Result<T, &'static str>) -> T {
 pub fn unwrap_or_throw_error<T>(r: std::result::Result<T, Error>) -> T {
     match r {
         Err(e) => {
-            throw_r_error(e.to_string());
+            let msg = e.to_string();
+            drop(e);
+            throw_r_error(msg);
         }
         Ok(v) => v,
     }


multimeric:
Is just dropping `e` enough? What about all the other `Robj` in the parent scope like the input arguments?

yutannihilation:
I'm not sure, but, when creating an `Error`, isn't the `Robj` moved to the `Error`? What do you mean by "all the other `Robj` in the parent scope"? Anyway, this pull request is just for showing the basic idea and I'm not interested in polishing this.

multimeric:
Yeah I guess most of the times this is happening, the `#[extendr]` function has returned a result, so all of its arguments have been dropped (?), and it's only the Robj in the error that hasn't been?

yutannihilation:
In my understanding, extendr assumes it's the author's responsibility to drop all the resources manually before calling `throw_r_error()`. So, I think you are right in that there's nothing we can do to prevent leaking the other arguments that are already converted to `Robj` before the one that fails.

I think the root cause is that the extendr's error handling is basically broken. IMHO, there's no choice to stop calling `Rf_error()` on Rust's side. I mean, [on R's side](https://github.com/extendr/extendr/issues/278#issuecomment-1226938512) or [on C's side](https://github.com/extendr/extendr/issues/278#issuecomment-1426995505).

multimeric:
> In my understanding, extendr assumes it's the author's responsibility to drop all the resources manually before calling throw_r_error().

But in extendr we call `throw_r_error` for the user on panic or error, so we need to handle this automatically if possible.

> I think the root cause is that the extendr's error handling is basically broken. IMHO, there's no choice to stop calling Rf_error() on Rust's side. I mean, https://github.com/extendr/extendr/issues/278#issuecomment-1226938512 or https://github.com/extendr/extendr/issues/278#issuecomment-1426995505.

Why do you say we need to use `Rf_error`? [Soren's PR](https://github.com/extendr/extendr/pull/473) has proposed a good alternative where we just return an object to R that represents an error without needing to use `Rf_error` at all.

multimeric:
Okay I've tested thsi branch and dropping the error doesn't seem to solve the memory leak: https://github.com/multimeric/ExtendrMemoryLeaking/blob/drop/MemoryLeak.md

yutannihilation:
Thanks, then it seems I was wrong. Closing this pull request.

yutannihilation:
> Why do you say we need to use `Rf_error`?

No. I'm saying extendr should stop using `Rf_error()`, so https://github.com/extendr/extendr/pull/473 looks fine to me.

yutannihilation:
@multimeric 
JFYI, I confirmed this pull request would solve the particular issue I thought was the core problem. Not sure if it was the same as the one in your mind, though.

diff: https://github.com/extendr/extendr/compare/ca001033e371ff3f52022cea72aa534f2958f041...yutannihilation:debug-issue-549

``` r
library(rextendr)

patch.crates_io <-
  list(
    `extendr-api` = list(
      git = "https://github.com/yutannihilation/extendr.git",
      branch = "debug-issue-549"
    )
  )

options(rextendr.patch.crates_io = patch.crates_io)

fn <- r"(
fn give_me_string_vec(x: Strings) -> String {
    format!("thank you for Strings with length {} ...", x.len())
})"

# case 1) without drop()

rust_function(fn, extendr_fn_options = list("use_try_from" = TRUE), quiet = TRUE)
give_me_string_vec(1)
#> index: 0, refcount: 0, is_null: true
#> index: 1, refcount: 0, is_null: true
#> index: 2, refcount: 0, is_null: true
#> index: 3, refcount: 1, is_null: false
#> Error in give_me_string_vec(1): Expected Strings got Doubles
give_me_string_vec(1)
#> index: 0, refcount: 0, is_null: true
#> index: 1, refcount: 0, is_null: true
#> index: 2, refcount: 0, is_null: true
#> index: 3, refcount: 1, is_null: false
#> index: 4, refcount: 1, is_null: false
#> Error in give_me_string_vec(1): Expected Strings got Doubles

# case 2) with drop()

#teak
cfg <- rextendr:::features_config
cfg$known_features <- c(cfg$known_features, "use_drop")

rust_function(fn, extendr_fn_options = list("use_try_from" = TRUE), features = "use_drop", quiet = TRUE)
give_me_string_vec(1)
#> index: 0, refcount: 0, is_null: true
#> index: 1, refcount: 0, is_null: true
#> index: 2, refcount: 0, is_null: true
#> index: 3, refcount: 0, is_null: true
#> Error in give_me_string_vec(1): Expected Strings got Doubles
give_me_string_vec(1)
#> index: 0, refcount: 0, is_null: true
#> index: 1, refcount: 0, is_null: true
#> index: 2, refcount: 0, is_null: true
#> index: 3, refcount: 0, is_null: true
#> index: 4, refcount: 0, is_null: true
#> Error in give_me_string_vec(1): Expected Strings got Doubles
```

<sup>Created on 2023-05-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

sorhawell:
In hindsight I think this PR was on the right track, the were just more objects, that needed to be dropped also.

sorhawell:
Maybe more objects were needed to be dropped.

sorhawell:
From my tests it seems this PR would solve leakage in the case where `use_try_from=true` and there is only on input param. However, if there are two input params. Leakage still can occur from first param.

E.g. if a function receives param1 and param2. Where conversion succeeds on 1, but fails on 2. Then only param2 would be dropped here and the conversion of param1 would not be dropped.

yutannihilation:
Ah, got it. Thanks, it seems my understanding was wrong. Since `handle_panic()` wraps only the call of the original function and the conversion from bare `SEXP`s to `Robj` is done before this point (I don't confirm by `cargo expand` yet), these `Robj`s don't get unwound.

https://github.com/extendr/extendr/blob/ca001033e371ff3f52022cea72aa534f2958f041/extendr-macros/src/wrappers.rs#L112-L114

PR #555, opened 2023-05-21
fix leaking on conversion errors and panics

### LINKED ISSUE #549: implicit function param conversions leaks memory

sorhawell:
fixes #549


sorhawell:
[test results](https://github.com/sorhawell/does_it_leak/blob/simple_example/tests/leak_result.txt).

The tiny 50-250 bytes leaks in first test is likely a burn-in fluke.

sorhawell:
example of the new macro expansion:
image

extendr-macros/src/wrappers.rs
```diff
@@ -102,17 +100,44 @@ pub fn make_function_wrappers(
...
     // ```
+    //
     wrappers.push(parse_quote!(
         #[no_mangle]
         #[allow(non_snake_case, clippy::not_unsafe_ptr_arg_deref)]
         pub extern "C" fn #wrap_name(#formal_args) -> extendr_api::SEXP {
-            unsafe {
-                use extendr_api::robj::*;
+            use extendr_api::robj::*;
+            let res_res: std::result::Result<
```
multimeric, line 109: More meaningful name

extendr-macros/src/wrappers.rs
```diff
@@ -102,17 +100,44 @@ pub fn make_function_wrappers(
...
+                std::panic::catch_unwind(||-> std::result::Result<Robj, extendr_api::Error> {
+                    Ok(extendr_api::Robj::from(#call_name(#actual_args)))
+                })
+            };
+            match res_res {
+                Ok(Ok(zz)) => {
+                    return unsafe { zz.get() };
+                }
+                Ok(Err(extendr_err)) => {
+                    let err_string = extendr_err.to_string();
```
multimeric, line 123: I'm confused as to how the `Ok(Err)` path occurs. Won't it always be `Err()` or `Ok(Ok())` given this code?

sorhawell:
> Hmm, this confuses me because it really seems like the only thing that this changes is adding `drop(extendr_err)` which is basically what #551 did. Am I missing something?

The difference is unwind and throw_r_error is very dodgy way to do clean up memory and even though all objects is dropped in one scope, there could be other objects in the nested tree of scopes which aren't dropped when doing longjump.

A quite safer/cleaner way is to return a nested result object from the unsafe scope. Then guaranteed by the rust scope rules, anything but the result-object is then dropped. Then only the result object in a single scope needs to be dealt with  This makes it easier to get it right.

See also [comment in 551](https://github.com/extendr/extendr/pull/551#issuecomment-1556854052).

.. I should though provide a link to the tests 551 fails

extendr-macros/src/wrappers.rs
sorhawell: `Err(Any)` contains any caught panic. `Ok(Err(Error))` contains any conversion error. Since on new line 330 I replace `unwrap_or_throw_error` with ?-operator to bubble error instead of throwing and panicking. There for the expansion of `#actual_args` has one `?` for each parameter. When a conversion fails  will return from closure on line 114.  Notice the type on line 114 is return type of the closure, not from `catch_unwind()`.

sorhawell: if some user function itself returns a Result-type which is Err and since something like #473 is not adopted, `extendr_api::Robj::from` can then also panic in the post-conversion back to R. I define this also as a user induced panic. So a user function will only panic if the user uses panic/unwrap/... or if returning an Err.

sorhawell: image

sorhawell: looking at the expansion again, I see can drop the `map_err(|e| ...)` as `?`-operator does that implicitly.
image

yutannihilation: > I'm confused as to how the `Ok(Err)` path occurs. 

Yeah, this is tricky, but this has to be like this because we have to handle `panic!()`.

Just for information, Lionel Henry kindly suggested that I could write some C code to wrap `R_UnwindProtect()` so that no `panic!()` is needed to escape from the cleanup function. I'll experiment this idea and share the result (it probably won't be very soon, so please forget for now).

sorhawell:
@Ilia-Kosenkov @multimeric 

I believe this PR is ready to merge and all conversations are resolved.

I think it is central to extendr to solve the memory leaking now.

CGMossa:
I'll review it soon.

CGMossa:
Does this **actually** fix the leaks?

I've no comments to add.. But I cannot approve this because there are no tests,
and no message from you that state that this branch, doesn't leak memory.
You even made the leak tests for this! Am I missing something?

(sorry I had food poisoning)

multimeric:
Agreed that tests are important here. Could the R benchmarks be converted into unit tests?

CGMossa:
@multimeric okay, I might understand this now: This message from @sorhawell state that he ran tests on this PR, master, and #551. https://github.com/extendr/extendr/pull/555#issuecomment-1557134693 And this PR passes everything.

@sorhawell I think you should just put these tests in this PR as well.

multimeric:
Yeah, I know he tested this externally, it's just that including them in the PR increases our confidence about it, and prevents regressions.

Ilia-Kosenkov:
I am having trouble mapping the expanded macro from the screenshot above to the actual code right now. Would you be able to show us an expanded macro of a function with, say, three different args?

sorhawell:
> I am having trouble mapping the expanded macro from the screenshot above to the actual code right now. Would you be able to show us an expanded macro of a function with, say, three different args?

One way to see the expansion is to e.g. use helloextendr or similar. Refer to my repo + branch in Cargo.toml. Then run `cargo expand` .

I will try to post an example at some point.

sorhawell:
@Ilia-Kosenkov  I have made an example from helloextendr with 3 args and use_try_from  = true and false

https://github.com/sorhawell/three_arg_expansion/blob/main/README.md

extendr-macros/src/wrappers.rs
```diff
@@ -102,17 +100,48 @@ pub fn make_function_wrappers(
...
+
+
+            match wrap_result_state {
+                Ok(Ok(zz)) => {
+                    return unsafe { zz.get() };
+                }
+                // any conversion error bubbled from #actual_args conversions of incomming args from R.
+                Ok(Err(conversion_err)) => {
+                    let err_string = conversion_err.to_string();
+                    drop(conversion_err); // try_from=true errors contain Robj, this must be dropped to not leak
```
Ilia-Kosenkov, line 128: I may be missing the context, but why is it not dropped at the end of the scope (as all Rust things are)? Is it because `::throw_r_error()` breaks things up?

sorhawell: Paraphrasing the writings of @yutannihilation . Rust has a stack and perhaps some heap allocations. Performing longjump will not unwind the stack and free heap allocations. But also panic unwind is discouraged in rust docs and stated as UB when returning to another language as R.

Previous attempts to manually drop before throw_r_error() did not fully succeed because they only dropped one or some scopes on top of the stack. By first returning `Result` the whole conversion stack can be dropped, and then only a single object needs to be dealt with. That is a saner way to deal with all edge cases and it emperically solves any known case of leaking out of ~150 test cases.

I have used with r-polars and there is none of 1200 unit tests which fail.

Ilia-Kosenkov:
So what is the consensus here? Do we need integration tests or perhaps separate benchmarks in the project?

sorhawell:
I was looking extendr tests yesterday, and was thinking to place one or two extendr-impls like  [here](https://github.com/extendr/extendr/blob/master/tests/extendrtests/src/rust/src/optional_either.rs)

and then write a test in testthat folder.

sorhawell:
I have added the benchmark as integration test and reduced testing from ~150 to 20 tests.

yutannihilation:
Yes, as I wrote "except for Windows," nothing new.

sorhawell:
I do not plan to add more to this PR for now.
For my own pet project I will adopt 555, 553 and 473.

Ilia-Kosenkov:
Sorry, I was overwhelmed at work recently and did not check these PRs. I will review them & we will hopefully merge.

sorhawell:
@yutannihilation 
Now when this [other unrelated win-build error](https://github.com/extendr/extendr/actions/runs/5200353819/jobs/9378963508) ever went away, what is your opinion on merging this PR?

yutannihilation:
I'm in [a different direction](https://github.com/extendr/extendr/pull/555#discussion_r1207849111), so I'm not the right one to judge this, but, to me, there seems no reason not to merge this pull request. It looks far better than the current code.

CGMossa:
@multimeric should we just merge this then?

Ilia-Kosenkov (approved):
This should also be mentioned in `CHANGELOG.md`

eitsupi:
@sorhawell Could you update the changelog?

sorhawell:
@Ilia-Kosenkov  changelog updated

CGMossa:
Thanks @Ilia-Kosenkov

CGMossa:
@yutannihilation Interesting, have you seen this https://www.tidyverse.org/blog/2019/05/resource-cleanup-in-c-and-the-r-api/ ? Did you already link to it?

yutannihilation:
No.

PR #558, opened 2023-06-05
Use libR-sys v0.5.0

yutannihilation:


CGMossa (approved):
Perfection

yutannihilation:
It seem things are not so easy :)

Maybe I need some tweak on rextendr? I forget everything about what was needed when we updated libR-sys last time...

CGMossa:
Windows failed. First thing to check is length of paths. If it exceeds a
certain threshold then we are in trouble.

yutannihilation:
It seems this problem is related to Rust 1.70. I see such errors like

```
...snip...\std\src\sys\windows/handle.rs:290: undefined reference to `NtWriteFile'
```

I'm wondering why we didn't catch this with the nightly toolchain, but maybe we don't have any CI runner with nightly on Windows...?

yutannihilation:
The line 290 of the fine is this.

https://github.com/rust-lang/rust/blame/1.70.0/library/std/src/sys/windows/handle.rs#L290

This line is not very new. The code has been here from 1 year ago, so I guess this is related to stabilization of some feature. Maybe `OnceCell`?

yutannihilation:
Note: It can be solved by adding `-lntdll`.

yutannihilation:
Windows tests still fail, but let's merge this to let other runners pass.

PR #560, opened 2023-06-09
Add -lntdll and ignore warnings on Windows oldrel

yutannihilation:
A follow-up of https://github.com/extendr/rextendr/pull/285.


CGMossa (approved):

yutannihilation:
Thanks for approving. While there might be some better option to deal with the warning on oldrel, I believe this is better than doing nothing and let the CI keep failing. I'm merging this for now.

PR #562, opened 2023-06-22
DO NOT MERGE: test if extendr tests currently fails due to win toolchain

sorhawell:


yutannihilation:
Pro-tip: you can use `--allow-empty` to create an empty commit.

yutannihilation:
Can we close this?

PR #563, opened 2023-06-22
Fix win toolchain

sorhawell:
[main extendr tests currently fails latest nightly because cannot install toolchain for 1-2 msvc targets](https://github.com/extendr/extendr/actions/runs/5345995596/jobs/9692375964?pr=562)

When I google and find similar errors. It appears to have been a reoccurring issue the last 5 years in rust.

[one solution which worked when the problem appeared in 2019 was to temporarily fix nightly](https://github.com/rust-lang/rust/issues/57488#issuecomment-453040972)

py-polars and r-polars runs on the same nightly version.

If accepting this PR, I suggest to make ("DO NOT MERGE PR: reverting fix nightly") to monitor when the error goes away again.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -270,8 +270,8 @@ jobs:
           # When the MSVC target is used, since it's cross-compilation from MSVC
           # to GNU, the doc tests are usually skipped. Adding `-Zdoctest-xcompile`
           # lets the doctests run, which accordingly require the nightly toolchain.
-          - {os: windows-latest, r: 'release', rust-version: 'nightly-msvc', rust-targets: ['x86_64-pc-windows-gnu'], rtools-version: '43', extra-args: ['-Zdoctest-xcompile']}
-          - {os: windows-latest, r: 'oldrel',  rust-version: 'nightly-msvc', rust-targets: ['x86_64-pc-windows-gnu'], rtools-version: '42', extra-args: ['-Zdoctest-xcompile']}
+          - {os: windows-latest, r: 'release', rust-version: 'nightly-2023-04-11-msvc', rust-targets: ['x86_64-pc-windows-gnu'], rtools-version: '43', extra-args: ['-Zdoctest-xcompile']}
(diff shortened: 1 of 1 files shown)


sorhawell:
@yutannihilation @Ilia-Kosenkov  I believe this PR will fix the remaining error in #555 . The error was unrelated to 555 but is likely due to a temporary incompatibility between nightly and msvc toolchain.

yutannihilation:
Hmm, thanks for investigating! However, if this is a temporary issue, I feel it's easier to wait rather than pinning it.

CGMossa:
It's blocking the very difficult process of reviewing #555. I think we should strive to make main ci to pass, as waiting is affecting other activities.

sorhawell:
> It's blocking the very difficult process of reviewing #555. I think we should strive to make main ci to pass, as waiting is affecting other activities.

Sorry I might have to take it back if it was related to 555.

pinning or not, I think is  a fair matter of preference. I personally like them green badges and to see all tests complete as much as possible.

yutannihilation:
Okay, let's merge then. I have no strong opinion here, but I believe no one want to get pinned to the outdated nightly. Please don't forget to file an issue to revert this change.

yutannihilation (approved):

CGMossa (approved):

sorhawell:
I was trying to take back the relation to 555 above.

I had the error in all my private branches today.

https://github.com/extendr/extendr/actions/runs/5345995596/jobs/9692375964?pr=562 here is an example of before merge where it is a problem in main extendr today.

I will set up the revert PR to track when it resolved again.

It could otherwise be I have analyzed the error wrong and it is really a flaky error.

yutannihilation:
Yeah, I know it actually failed. But, you pushed [a commit of a blank line](https://github.com/extendr/extendr/pull/555/commits/a1159ce03ed9c4a34692237ab623540b0d07bdff) to #555 before this gets merged, and all checks passed. Am I wrong?

https://github.com/extendr/extendr/actions/runs/5346377680/jobs/9693322158

Issue #565, opened 2023-06-26
rprintln! cannot print `%` via extendr and may cause segfault


sorhawell:
## fail to print `%`
See below for testcase  that % does not work with extendr. `println` works fine.

## segfault 
Regarding segfaults, I cannot produce a minimal example as it occurs so far [only when printing polars DataFrame with `%`](https://github.com/pola-rs/r-polars/issues/258). Some times it causes R to crash with segfault error, sometimes it will not print any column name, sometimes it will print some gibberish (I guess this is a half segfault, just not caught by OS yet, or NOT, but rather some formatted character).

## reason:
dunno maybe related to `%` is special character in C-like string format function, like `sprintf`

For one of the tests I add an extra inserted field `{}, "also me"`. Some gibberish is shown in Rstudio IDE but cannot be copied or reproduced with reprex
<img width="296" alt="image" src="https://github.com/extendr/extendr/assets/8015271/c2f9c4cc-adfa-4032-b90d-32d9726e2582">
The reason is likely it is printed not to R terminal but to standard out. This supports that extendr does not correctly escape the `%` and this will trigger print failure and potentially something worse.


#### EDIT: I previously believed that `'%'` was somehow worse than `%` . Test below does not reproduce that.


## [testcase derived from helloextendr](https://github.com/sorhawell/percentage_print) 

```rust
#[extendr]
fn hello_world() -> &'static str {
    rprintln!("------------------------------------");
    rprintln!("\n\n print pct-char ");
    rprintln!("123%321");
    rprintln!("print completed");

    rprintln!("\n\n print quoted pct-char ");
    rprintln!("123'%'321");
    rprintln!("print completed");

    rprintln!("\n\n print inserted pct-char : ");
    rprintln!("123{}321", "%");
    rprintln!("print completed");

    rprintln!("\n\n print quoted inserted pct-char : ");
    rprintln!("123{}321{}321", "'%'", "also me");
    rprintln!("print completed");

    rprintln!("------------------------------------");
    rprintln!("\n\n print dollarsign-char ");
    rprintln!("123$321");
    rprintln!("print completed");

    rprintln!("\n\n print quoted dollarsign-char ");
    rprintln!("123'$'321");
    rprintln!("print completed");

    rprintln!("\n\n print inserted dollarsign-char : ");
    rprintln!("123{}321", "$");
    rprintln!("print completed");

    rprintln!("\n\n print quoted inserted dollarsign-char : ");
    rprintln!("123{}321", "'$'");
[...]

PR #566, opened 2023-06-26
try fix print pct

sorhawell:


sorhawell:
fixes #565 

no unit test yet

extendr-api/src/rmacros.rs
```diff
@@ -144,7 +144,7 @@ macro_rules! rprint {
     () => {
     };
     ($($rest: tt)*) => {
-        print_r_output(format!($($rest)*));
+        print_r_output(format!($($rest)*)).replace("%","%%");
```
CGMossa, line 147: Isn't the way to escape these is to use `.replace("%", "\%")`?

sorhawell: fair question

[this reference ](https://en.cppreference.com/w/cpp/io/c/fprintf) says it should be `%%`, [this SO says that](https://stackoverflow.com/questions/20583827/difference-between-and) `\%` is illegal because it does not have any defined meaning.

printf uses a literal `%` as the only special character
the escaped `%` as `\%` does not have a defined meaning and should be avoided
printf uses a `%` the special character to escape `%` when needed literally, therefore `%%`

CGMossa: No, you're right about the `%%`.

CGMossa:
Maybe we need a reference to this page: https://en.cppreference.com/w/c/io/fprintf

```c


void Rprintf(const char *format, ...)
{
    va_list(ap);

    va_start(ap, format);
    Rvprintf(format, ap);
    va_end(ap);
}
```

```c

void Rvprintf(const char *format, va_list arg)
{
    int i=0, con_num=R_OutputCon;
    Rconnection con;
    va_list argcopy;
    static int printcount = 0;

    if (++printcount > 100) {
	R_CheckUserInterrupt();
	printcount = 0 ;
    }

    do{
      con = getConnection(con_num);
      va_copy(argcopy, arg);
      /* Parentheses added for Fedora with -D_FORTIFY_SOURCE=2 */
      (con->vfprintf)(con, format, argcopy);
      va_end(argcopy);
      con->fflush(con);
      con_num = getActiveSink(i++);
    } while(con_num>0);
}
```

And https://en.cppreference.com/w/c/io/vfprintf which is similar.

sorhawell:
a fair decision is whether to keep the macros `rprintln! , reprinln!, rprint! and reprint!`as is and add warnings in docs about bad printing and at worst segfault. googling `printf` it seems various character combinations can trigger a segfault. 

I would personally prefer to one extra string allocation to not think twice about when to be careful about `rprinln!`.
I would personally add to docs that the more low-level and unsafe print_r_output() exists, where the use can avoid 1-2 string allocations.

sorhawell:
in theory
[<String>::replace() ](https://doc.rust-lang.org/src/alloc/str.rs.html#271)could be optimized to scan the string and only reallocate if any `%`-char is found. Otherwise no extra allocation, just one linear scan.

Ilia-Kosenkov:
Consulting with all-knowing Chat-GPT, I suspect something like this could be possible
```rust
pub fn print_r_output<T: Into<Vec<u8>>>(s: T) {
    let fmt = CString::new("%s").expect("NulError"); // Based on C convention for formatting, can possible be cached?
    let cs = CString::new(s).expect("NulError");
    unsafe {
        Rprintf(fmg.as_ptr(), cs.as_ptr());
    }
}
```

sorhawell:
That is more efficient.

Ilia-Kosenkov:
Can you try this out in the PR branch btw?

sorhawell:
@Ilia-Kosenkov 

how about this one?

```rust
const PRINTF_NO_FMT_CSTRING: &'static [i8] = &[37, 115, 0]; // same as "%s\0"
#[doc(hidden)]
pub fn print_r_output<T: Into<Vec<u8>>>(s: T) {
    let cs = CString::new(s).expect("NulError");
    unsafe {
        Rprintf(PRINTF_NO_FMT_CSTRING.as_ptr(), cs.as_ptr());
    }
}

#[doc(hidden)]
pub fn print_r_error<T: Into<Vec<u8>>>(s: T) {
    let cs = CString::new(s).expect("NulError");
    unsafe {
        REprintf(PRINTF_NO_FMT_CSTRING.as_ptr(), cs.as_ptr());
    }
}
```

Ilia-Kosenkov:
Yep. But did it fix the issue? :)

sorhawell:
Yes. Now the unit test prints `Hello World %%!`. The benefit of testing with `%%` is will yield `%%` of success and `%` if fail. No segfault or strange stuff.

Ilia-Kosenkov (approved):

eitsupi:
@extendr Could anyone merge this?

eitsupi:
Thanks for merging!

PR #568, opened 2023-06-27
ci: add dependabot.yml for rust and github actions

eitsupi:
I think we can automate the updating of dependencies.

However, a MSRV checking workflow may be needed to keep Rust versions compatible.
So please merge #569 first.


Ilia-Kosenkov (approved):
Weird. I was positive I already approved this thing. Anyway, here it goes

PR #573, opened 2023-07-02
`PartialOrd` for scalars

### LINKED ISSUE #570: Support `PartialOrd` trait for R scalars
At present, there is no way to compare the max or min of two `Rfloat`s. To do this, we need to grab the inner f64.  For example using an `Rfloat` `x` and `y` and return an Rfloat this is done: 

`Rfloat::from(x.0.min(y.0))`

Consider treating `NA` as `NaN` so no comparison is done.

Ilia-Kosenkov:
- `Scalar<T>` to initiate abstraction of common scalar properties
- `PartialOrd` for `Rint`, `Rfloat`, `Rbool` (`NA`-aware)
- `min()/max()` on `Rint` & `Rfloat` based on `PartialOrd`
- `From<Rint>` and `From<i32>` for `Rfloat` (note that `i32::MIN_VALUE` will not be treated as `NA_real_`)

Closes #570

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Added
 
 - [**either**] `TryFrom<&Robj> for Either<T, R>` and `From<Either<T, R>> for Robj` if `T` and `R` are themselves implement these traits. This unblocks scenarios like accepting any numeric vector from R via `Either<Integers, Doubles>` without extra memory allocation [[#480]](https://github.com/extendr/extendr/pull/480)
-- `PartialOrd` trait implementation for `Rfloat`, `Rint` and `Rbool`. `Rfloat` and `Rint` gained `min()` and `max()` methids [[#573]](https://github.com/extendr/extendr/pull/573)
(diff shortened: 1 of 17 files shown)


Ilia-Kosenkov:
@JosiahParry , ready for initial review :-)

CGMossa (approved):
LGTM, if it passes test.. then it should be merged.

Ilia-Kosenkov:
Funny thing it does not really solve the original problem -- `min()` and `max()` require `Ord` , not `PartialOrd`. We cannot provide `Ord` of course, but at least now it should work like 
```rust
vec![Rint::from(1), Rint::from(2)].iter().filter(|&x| x >= 2);
```

yutannihilation:
That's what I was wondering about when I commented on https://github.com/extendr/extendr/issues/570. The reason `f64` has `min()` and `max()` is not that `f64` is `Ord`, but it implements its own `min()` and `max()`.

https://doc.rust-lang.org/std/primitive.f64.html#method.min

Ilia-Kosenkov:
Ah yeah, no problem! We already have some methods defined (mostly R-inspired), I can extend it to also include `min()`, `max()`, etc

PR #578, opened 2023-07-08
Add `...` support

### LINKED ISSUE #576: Add support for `...` parameter in extendr functions
Here is a potential design of the feature:
- Introduce a parameter level attribute, similar to `#[default]`, say, `#[dotdotdot]`, or `#[ellipsis]` and allow marking (ideally, at most one) `List` parameter
- On the side of R, update wrapper generation to accept a single `...` and pass it to Rust function as `list(...)`.

Ilia-Kosenkov:
- Add support for `...` parameters in `extendr`-functions.
  - Introduce `Ellipsis` type that lazily consumes `...`
  - Allow iterating over raw values (either `Promise` or `missing_arg`)
  - Allow quickly collecting all values by evaluating `Promise`s, with at most one allowed trailing missing arg (`rlang::list2`-style)
- Implement conversion from `Robj` in two steps:
  - If `Robj` is `Environment`, try searching for `...` element in the environment and then
    - If `...` is `missing_arg`, return empty `Ellipsis` (case when nothing is captured by `...`)
    - If `...` is `DOTSXP`, capture it in `Ellipsis`
- Add `#[ellipsis]` attribute to signal wrapper generator to rename parameter & capture `environment()`
- Fix incorrect conversion when `use_try_from = true`:  `TryFrom<Robj>` was used instead of `TryFrom<&Robj>`, which is the preferred one

Closes #576

### DIFF
--- a/extendr-api/src/metadata.rs
+++ b/extendr-api/src/metadata.rs
@@ -56,8 +56,7 @@ pub struct Metadata {
 
 struct RArg {
     name: String,
-    default: Option<&'static str>,
-    is_ellipsis: bool,
+    modifier: ArgModifier,
 }
 
 impl RArg {
@@ -66,24 +65,19 @@ impl RArg {
     }
 
     fn to_actual_arg(&self) -> String {
-        if self.is_ellipsis {
-            "environment()".to_string()
+        if matches!(self.modifier, ArgModifier::Ellipsis) {
+            "environment()".into()
         } else {
             self.name.clone()
         }
     }
 
     fn to_formal_arg(&self) -> String {
-        match self {
-            Self {
-                default: Some(default_val),
-                ..
-            } => {
+        match self.modifier {
+            ArgModifier::Default(default_val) => {
                 format!("{} = {}", self.name, default_val)
             }
-            Self {
-                is_ellipsis: true, ..
-            } => "...".to_string(),
+            ArgModifier::Ellipsis => "...".to_string(),
             _ => self.name.clone(),
         }
     }
@@ -93,12 +87,7 @@ impl From<&Arg> for RArg {
     fn from(arg: &Arg) -> Self {
         Self {
             name: sanitize_identifier(arg.name),
-            default: if let ArgModifier::Default(default) = arg.modifier {
-                Some(default)
-            } else {
-                None
-            },
-            is_ellipsis: matches!(arg.modifier, ArgModifier::Ellipsis),
+            modifier: arg.modifier,
         }
     }
 }
--- a/extendr-api/src/metadata.rs
+++ b/extendr-api/src/metadata.rs
@@ -6,12 +6,19 @@ use crate::robj::IntoRobj;
 use crate::*;
 use std::io::Write;
 
+#[derive(Debug, PartialEq, Copy, Clone)]
+pub enum ArgModifier {
+    None,
+    Default(&'static str),
+    Ellipsis,
+}
+
 /// Metadata function argument.
 #[derive(Debug, PartialEq)]
 pub struct Arg {
     pub name: &'static str,
     pub arg_type: &'static str,
-    pub default: Option<&'static str>,
+    pub modifier: ArgModifier,
 }
 
 /// Metadata function.
@@ -69,7 +76,11 @@ impl From<&Arg> for RArg {
     fn from(arg: &Arg) -> Self {
         Self {
             name: sanitize_identifier(arg.name),
-            default: arg.default,
+            default: if let ArgModifier::Default(default) = arg.modifier {
+                Some(default)
+            } else {
+                None
+            },
         }
     }
 }
--- a/extendr-api/src/metadata.rs
+++ b/extendr-api/src/metadata.rs
@@ -6,10 +6,14 @@ use crate::robj::IntoRobj;
 use crate::*;
 use std::io::Write;
 
+/// Argument modifier.
 #[derive(Debug, PartialEq, Copy, Clone)]
 pub enum ArgModifier {
+    /// No modifier applied.
     None,
+    /// Default value is set via `#[default = "<value>"]`.
     Default(&'static str),
+    /// Argument is marked as `...` using `#[ellipsis]`.
     Ellipsis,
 }
 
--- a/extendr-api/src/metadata.rs
+++ b/extendr-api/src/metadata.rs
@@ -53,6 +53,7 @@ pub struct Metadata {
 struct RArg {
     name: String,
     default: Option<&'static str>,
+    is_ellipsis: bool,
 }
 
 impl RArg {
@@ -61,13 +62,25 @@ impl RArg {
     }
 
     fn to_actual_arg(&self) -> String {
-        self.name.clone()
+        if self.is_ellipsis {
+            "environment()".to_string()
+        } else {
+            self.name.clone()
+        }
     }
 
     fn to_formal_arg(&self) -> String {
-        match self.default {
-            Some(default_val) => format!("{} = {}", self.name, default_val),
-            None => self.name.clone(),
+        match self {
+            Self {
+                default: Some(default_val),
+                ..
+            } => {
+                format!("{} = {}", self.name, default_val)
+            }
+            Self {
+                is_ellipsis: true, ..
+            } => "...".to_string(),
+            _ => self.name.clone(),
         }
     }
 }
@@ -81,6 +94,7 @@ impl From<&Arg> for RArg {
             } else {
                 None
             },
+            is_ellipsis: matches!(arg.modifier, ArgModifier::Ellipsis),
         }
     }
 }
--- a/extendr-api/src/wrapper/ellipsis.rs
+++ b/extendr-api/src/wrapper/ellipsis.rs
@@ -39,6 +39,10 @@ impl Ellipsis {
             .map(|x| (x.name, x.value.to_promise()))
             .collect::<Vec<_>>();
 
+        if values.len() == 0 {
+            return Ok(vec![]);
+        }
+
         let n = values.len() - 1;
 
         values
@@ -57,7 +61,7 @@ impl Ellipsis {
                         let name = name
                             .clone()
                             .map(|nm| format!("`{}`", nm.as_str()))
-                            .unwrap_or_else(|| format!("[{}] element", i));
+                            .unwrap_or_else(|| format!("[{}] element", i + 1)); // To match R element numbering.
                         Some(Err(Error::Other(format!("Missing value for {}", name))))
                     }
                 }
@@ -75,6 +79,24 @@ impl<'a> TryFrom<&'a Robj> for Ellipsis {
             Rtype::Environment => try_from_env(robj),
             Rtype::Symbol if robj.is_missing_arg() => Ok(Ellipsis::new()),
             tp => Err(Error::Other(format!("Got {:?}: {:?}", tp, robj))),
+        }
+    }
+}
+
+impl TryFrom<Robj> for Ellipsis {
+    type Error = Error;
+
+    fn try_from(value: Robj) -> std::result::Result<Self, Self::Error> {
+        Ellipsis::try_from(&value)
+    }
+}
+
+impl<'a> FromRobj<'a> for Ellipsis {
+    fn from_robj(robj: &'a Robj) -> std::result::Result<Self, &'static str> {
+        if let Ok(f) = Ellipsis::try_from(robj) {
+            Ok(f)
+        } else {
+            Err("Not an ellipsis")
         }
     }
 }
@@ -136,14 +158,6 @@ impl TryFrom<&Robj> for EllipsisItemValue {
         } else {
             <Promise as TryFrom<&Robj>>::try_from(value).map(EllipsisItemValue::Promise)
         }
-    }
-}
-
-impl TryFrom<Robj> for EllipsisItemValue {
-    type Error = Error;
-
-    fn try_from(value: Robj) -> std::result::Result<Self, Self::Error> {
-        <EllipsisItemValue as TryFrom<&Robj>>::try_from(&value)
     }
 }
 
--- a/extendr-api/src/wrapper/ellipsis.rs
+++ b/extendr-api/src/wrapper/ellipsis.rs
@@ -1,0 +1,6 @@
+use crate::Robj;
+
+#[derive(PartialEq, Clone)]
+pub struct Ellipsis {
+    pub(crate) robj: Robj,
+}
--- a/extendr-api/src/wrapper/ellipsis.rs
+++ b/extendr-api/src/wrapper/ellipsis.rs
@@ -1,12 +1,28 @@
 use crate::prelude::*;
-use crate::{Error, Robj, Rtype, Types};
+use crate::{Error, R_NilValue, Robj, Rtype, Types, SEXP};
 
 #[derive(PartialEq, Clone)]
 pub struct Ellipsis {
     pub(crate) robj: Robj,
 }
 
-impl Ellipsis {}
+impl Ellipsis {
+    pub fn iter(&self) -> EllipsisIter {
+        unsafe {
+            EllipsisIter {
+                robj: Some(&self.robj),
+                list_elem: self.robj.get(),
+            }
+        }
+    }
+
+    pub fn collect_values(&self) -> Result<Vec<Robj>> {
+        self.iter()
+            .filter_map(|x| <Promise as TryFrom<&Robj>>::try_from(&x).ok())
+            .map(|p| p.eval())
+            .collect()
+    }
+}
 
 impl<'a> TryFrom<&'a Robj> for Ellipsis {
     type Error = Error;
@@ -15,7 +31,7 @@ impl<'a> TryFrom<&'a Robj> for Ellipsis {
         match robj.rtype() {
             Rtype::Dot => Ok(Self { robj: robj.clone() }),
             Rtype::Environment => try_from_env(robj),
-            _ => Err(Error::Other("Placeholder".into())),
+            tp => Err(Error::Other(format!("Got {:?}", tp))),
         }
     }
 }
@@ -28,4 +44,43 @@ fn try_from_env(env: &Robj) -> Result<Ellipsis> {
                 .ok_or(Error::Other("Ellipsis missing".into()))
         })
         .and_then(|(_, v)| <Ellipsis as TryFrom<&Robj>>::try_from(&v))
+}
+
+#[derive(Clone)]
+pub struct EllipsisIter<'a> {
+    pub(crate) robj: Option<&'a Robj>,
+    pub(crate) list_elem: SEXP,
+}
+
+impl<'a> Default for EllipsisIter<'a> {
+    fn default() -> Self {
+        EllipsisIter::new()
+    }
+}
+
+impl<'a> EllipsisIter<'a> {
+    pub fn new() -> Self {
+        unsafe {
+            Self {
+                robj: None,
+                list_elem: R_NilValue,
+            }
+        }
+    }
+}
+
+impl<'a> Iterator for EllipsisIter<'a> {
+    type Item = Robj;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        unsafe {
+            if self.robj.is_none() || self.list_elem == R_NilValue {
+                return None;
+            }
+
+            let elem = libR_sys::CAR(self.list_elem);
+            self.list_elem = libR_sys::CDR(self.list_elem);
+            Some(Robj::from_sexp(elem))
+        }
+    }
 }
--- a/extendr-api/src/wrapper/ellipsis.rs
+++ b/extendr-api/src/wrapper/ellipsis.rs
@@ -1,6 +1,31 @@
-use crate::Robj;
+use crate::prelude::*;
+use crate::{Error, Robj, Rtype, Types};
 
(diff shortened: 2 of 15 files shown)


extendr-api/src/metadata.rs
```diff
@@ -6,12 +6,19 @@ use crate::robj::IntoRobj;
 use crate::*;
 use std::io::Write;
 
+#[derive(Debug, PartialEq, Copy, Clone)]
+pub enum ArgModifier {
```
multimeric, line 10: Comment this and its variants please, explaining how it relates to R arguments

extendr-api/src/metadata.rs
```diff
@@ -46,6 +53,7 @@ pub struct Metadata {
 struct RArg {
     name: String,
     default: Option<&'static str>,
+    is_ellipsis: bool,
```
multimeric, line 56: Why is this modelled so that an `RArg` can *both* be ellipsis *and* have a default, whereas an `Arg` cannot be both? Can we be consistent here?

extendr-api/src/metadata.rs
```diff
@@ -54,13 +62,25 @@ impl RArg {
     }
 
     fn to_actual_arg(&self) -> String {
-        self.name.clone()
+        if self.is_ellipsis {
+            "environment()".to_string()
```
multimeric, line 66: This is wacky, can you explain what you're doing here?

extendr-api/src/wrapper/ellipsis.rs
```diff
@@ -0,0 +1,247 @@
+/*!
+Enables support for `...` parameters in R functions.
+`Ellipsis` can be used as a regular extendr-function argument. At most one `Ellipsis` is allowed per function,
+and it requires `#[ellipsis]` attribute in front of it in the signature.
+When encountered, `#[ellipsis]` attribute transforms parameter name into `...` on R side and captures current `environment()`.
```
multimeric, line 5:
```suggestion
When encountered, `#[ellipsis]` attribute transforms the parameter name into `...` on R side and captures current `environment()`.
```

extendr-api/src/wrapper/ellipsis.rs
```diff
@@ -0,0 +1,247 @@
+/*!
+Enables support for `...` parameters in R functions.
+`Ellipsis` can be used as a regular extendr-function argument. At most one `Ellipsis` is allowed per function,
+and it requires `#[ellipsis]` attribute in front of it in the signature.
```
multimeric, line 4:
```suggestion
and it requires the `#[ellipsis]` attribute in front of it in the signature.
```

multimeric (commented):
Some general comments before I get stuck into the minutiae of the PR:

* Since the Ellipsis type already marks a `...` param, and makes no sense for that type to appear elsewhere, do we actually need the `#[ellipsis]` macro as well?
* Is `environment()` really the best way to capture arguments? Why don't we just make a wrapper for `DOTSXP` to do this the "true" way? I feel sure there will be an edge case where the two behave differently

extendr-api/src/metadata.rs
Ilia-Kosenkov: I was lazy to properly handle this, but I can do that for sure!

Ilia-Kosenkov:
Also forgot to add examples & update the changelog.

multimeric:
> I followed the `rlang:"list2()` approach, since passing `...` directly did not work for me as expected. (I can try once again though).

Do you mean that `list2` captures `...` using `environment()`? Hmm, I wonder why. One thing I wonder is how you will handle distinguishing formal arguments from `...` arguments in the environment? Also I can't even get it to show `...` args with `environment()`, but maybe I'm doing it wrong:
```R
> y = function(z=4, ...){ print(ls(environment())) }
> y(b = 32, c= 3)
[1] "z"
```

> `#[ellipsis]` is needed because I found no way to reliably identify `Ellipsis` type using `syn` -- I do not want to just mark all inputs that are named `Ellipsis` as `...`, so I had to add this opt-in option. It is consumed by wrappers.

Oh you mean that you can identify the `Ellipsis` type, but you want to be sure it's *our* `Ellipsis` type? Hmm I wouldn't be overly concerned by this. I agree that the parser probably can't know that. What about `ellipsis = "foo"` in the function decorator? Would that be simpler?

Ilia-Kosenkov:
Take a look at my integration tests, it is shown there. `... ` captures everything that does not match formals. Environment has a hidden `... ` element which you can try accessing. It contains `DOTSXP`.

I'd prefer to keep parameter annotations at parameters. If we find a better solution in the future, we can update implementation, but now we will at least have `... ` support.

multimeric:
Okay, you've convinced me that `environment()` does capture `...`. For reference:
```R
> y = function(z=4, ...){ print(ls(environment(), all.names=TRUE)) }
> y(b = 32, c= 3)
[1] "..." "z"
```

I just feel that this would be a cool opportunity to support `DOTSXP`, but I guess there's no significant advantage in passing `...` directly.

Ilia-Kosenkov:
The thing is, I suspect that if you just pass `...` to a function call, it unwraps them to `n` different arguments, while our function actually contains a fixed number of parameters. So I believe an environment is the way to go. It also allows writing our own NSE, since we capture promises without evaluating them. As an alternative, we can substitute ```environment()$`...` ``` in the wrappers, but my code should already handle such an approach (it can convert `Ellipsis` from `Environment` and from `DOTSXP`).

JosiahParry:
I've tried using this and there are two important things missing, I think. I want to be able to force the evaluation of an `EllipsisIterItem` with either an `.eval()` method or `.into_robj()`. Right now I can grab the nth element from dots with the `.nth()` method from the iterator but I cannot get that as an extendr value. 

Also, is there any way to _not_ include the #[ellipsis] macro like so `fn fx(#[ellipsis] dots: Ellipsis)`? Can it be rewritten like `fn fx(dots: #[ellipsis])` or something a little less verbose (and honestly low key confusing to a novice like me! I thought this always had to be above something! 🤯)

Ilia-Kosenkov:
My goal is to iteratively improve on this feature and not check in ginormous PR with multiple changes affecting all of the extendr.
To answer your questions:
- You absolutely can extract n-th item and evaluate it. If you use `.iter()`, you get a by-ref iterator returning 
image
which contains this
image
You should be able to do something like `.iter().nth(n).and_then(|x| x.to_promise())`, which, if I did everything right here, will give you an `Option<&Promise>` or something similar. Which you can unwrap and evaluate to get an `Robj`. 
This is however not recommended, since unless you are implementing your own NSE support, there should not be a reason for evaluating only `nth()` element of `...` without the rest.

- How would you communicate that `dots : Ellipsis` should be mapped to `...` argument? Rust macros do not have access to compilation units and are oblivious to type information (at least to my knowledge). If i do not import `Ellipsis` from `extendr_api::prelude`, but rather define my own type, how would you handle that?
My current solution -- use an attribute to transform parameters of R function and arguments passed to `.Call()`. With next iteration, perhaps there will be a better solution.


P.s.: Perhaps my function signature is suboptimal, it should be
```rust
fn to_promise(&self) -> Option<&Promise> {todo!()}
```
but you can write it for yourself using standard `match` sytnax

sorhawell:
I have a new suggestion soon :)

sorhawell:
Calling R from rust and raising an error  `R!("stop("boom")")` or via `throw_r_error()` is problematic and should be avoided. AFAIK they don't leak memory anymore, but a consequence hereof is, the errors produced roam freely and can never be caught again with tryCatch. For that alone the current `Ellipsis`-PR has insufficient error handling. If the error actually could be caught, it would likely mean no return to rust after LONGJMP, hence no unwind, hence memory leak.

 - I agree with @JosiahParry and have made a Promise-iterator. Where next() return a Ok(Promise + Opt-name) or Err(syntax error). The iterator just skips any valid trailing comma.
 - I found that edge case where trailing arg is missing but named is currently allowed, but it should not be
 - The new iterator does not perform any intermediary collect
 -  I have re-implemented both current behavoir `eval_values()` and a more correct behavior `trycatch_values()`. The latter allow to access/catch the eval error from a promise on rust side.
 - I try to not panic on errors but return them to R
- I compare list2() with all implementations.

I have mad a [single page `lib.rs` extension to @Ilia-Kosenkov  PR to keep it simple.](https://github.com/sorhawell/dotdotdot_whos_there/blob/main/src/rust/src/lib.rs)

Conclusion:
 - Only `trycatch_values_result()` is fully equivalent to `list2()`
 - No implementaion appears to leak mem see `CheckMemReleased: OK`
 - promise.eval() is inherrently problematic as it also emits "uncatchable errors"
 - Personally I would prefer the `NamedPromiseIterator` for must use cases.
 
 
prints of [test script(try_ellipsis.R)](https://github.com/sorhawell/dotdotdot_whos_there/blob/main/tests/try_ellipsis.R) printed below
 
 ``` r
library(rlang)
library(helloextendr)

f_err = \() stop("boom")


#simple unwrap
unwrap = \(result) {
  if(!is.list(result) || !identical(names(result),c("ok","err"))) {
    stop("internal type error: cannot unwrap non result")
  }
  if(is.null(result$err)) {
    result$ok
  } else {
    stop(errorCondition(
      message = paste(capture.output(print(result$err)),collapse = "\n"),
      value = if(is.character(result$ok)) NULL else result$ok
    ))
  }
}

# normal.  -> ALL THE SAME
iter_dots(1,x=2,)
#> CheckMemReleased: OK
#> valid EllipsisNamedPromise: EllipsisNamedPromise { name: None, promise: Promise { code: 1.0, environment: global_env() } }
#> valid EllipsisNamedPromise: EllipsisNamedPromise { name: Some("x"), promise: Promise { code: 2.0, environment: global_env() } }
trycatch_dots(1,x=2,)
#> CheckMemReleased: OK
#> [[1]]
#> [1] 1
#> 
#> $x
#> [1] 2
trycatch_dots_result(1,x=2,) |> unwrap()
#> Error in trycatch_dots_result(1, x = 2, ): could not find function "trycatch_dots_result"
eval_dots(1,x=2,)
#> CheckMemReleased: OK
#> [[1]]
#> [1] 1
#> 
#> $x
#> [1] 2
collect_dots(1,x=2,)
#> CheckMemReleased: OK
#> [[1]]
#> [1] 1
#> 
#> $x
#> [1] 2
list2(1,x=2,)
#> [[1]]
#> [1] 1
#> 
#> $x
#> [1] 2


#invalid syntax before arg Error. -> NOT THE SAME
iter_dots(,f_err())
#> CheckMemReleased: OK
#> oups ... only allow trailing missing args: Index(1)
trycatch_dots(,f_err()) #A: panic on syntax Error
#> CheckMemReleased: OK
#> Error in trycatch_dots(, f_err()): user function panicked: trycatch_dots
trycatch_dots_result(,f_err()) |> unwrap() #B: raise syntax error
#> Error in trycatch_dots_result(, f_err()): could not find function "trycatch_dots_result"
eval_dots(,f_err())     # A: panic on syntax Error
#> CheckMemReleased: OK
#> Error in eval_dots(, f_err()): user function panicked: eval_dots
collect_dots(,f_err())  # A: panic on syntax Error
#> CheckMemReleased: OK
#> Error in collect_dots(, f_err()): user function panicked: collect_dots
list2(,f_err())  #B: raise syntax error
#> Error in `list2()`:
#> ! Argument 1 can't be empty.
#> Backtrace:
#>     ▆
#>  1. ├─rlang::list2(, f_err())
#>  2. └─rlang::abort(message = message)


#arg Error before invalid syntax -> NOT THE SAME
iter_dots(f_err(),,)
#> CheckMemReleased: OK
#> valid EllipsisNamedPromise: EllipsisNamedPromise { name: None, promise: Promise { code: lang!(sym!(f_err)), environment: global_env() } }
#> oups ... only allow trailing missing args: Index(2)
trycatch_dots(f_err(),,)  #A: panic on arg Error
#> CheckMemReleased: OK
#> Error in trycatch_dots(f_err(), , ): user function panicked: trycatch_dots
trycatch_dots_result(f_err(),,) |> unwrap()  #B: Raise Arg Error
#> Error in trycatch_dots_result(f_err(), , ): could not find function "trycatch_dots_result"
eval_dots(f_err(),,)     #C: insta throw arg error, panic on generic extendr-error
#> CheckMemReleased: OK
#> Error in eval_dots(f_err(), , ): user function panicked: eval_dots
collect_dots(f_err(),,)  #C: insta throw arg error, panic on generic extendr-error
#> CheckMemReleased: OK
#> Error in collect_dots(f_err(), , ): user function panicked: collect_dots
list2(f_err(),,)         #B: Raise Arg Error
#> Error in f_err(): boom

#TOO LATE TO CTACH f_err() from R
tryCatch(eval_dots(f_err()),error=\(err) as.character(err)) |> print()
#> CheckMemReleased: OK
#> [1] "Error in eval_dots(f_err()): user function panicked: eval_dots\n"
tryCatch(collect_dots(f_err()),error=\(err) as.character(err)) |> print()
#> CheckMemReleased: OK
#> [1] "Error in collect_dots(f_err()): user function panicked: collect_dots\n"


#error NamedMissingTrailingArg -> NOT ALL the same
iter_dots(1,x=)
#> CheckMemReleased: OK
#> valid EllipsisNamedPromise: EllipsisNamedPromise { name: None, promise: Promise { code: 1.0, environment: global_env() } }
#> another error surprisingly happended Other("trailing [x=] arg in ... was named but not defined")
trycatch_dots(1,x=) #A Panic on syntax error
#> CheckMemReleased: OK
#> Error in trycatch_dots(1, x = ): user function panicked: trycatch_dots
trycatch_dots_result(1,x=) |> unwrap()  #B: Raise syntax Error
#> Error in trycatch_dots_result(1, x = ): could not find function "trycatch_dots_result"
eval_dots(1,x=) #A: panic on syntax error
#> CheckMemReleased: OK
#> Error in eval_dots(1, x = ): user function panicked: eval_dots
collect_dots(1,x=)  #C: ACCEPT TRAILING NAMED ARG  <-- THIS IS A BUG I THINK
#> CheckMemReleased: OK
#> [[1]]
#> [1] 1
list(1,x=) #B: Raise syntax Error
#> Error in list(1, x = ): argument 2 is empty
```

<sup>Created on 2023-07-21 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

**Comment by Ilia-Kosenkov** [comment]
How do you reconcile this with https://github.com/extendr/extendr/pull/578#issuecomment-1638826251 ?

**Comment by sorhawell** [comment]
Fast analogy: stream of bytes -> Utf-8 encoding (chars) -> words.
We can iterate with different levels of abstraction. It means different things to "get 5th element" of a byte stream, a char or a word even though originating from the same base stream/iterator.

The lower level is `EllipsisIter` (think very loosly stream of bytes: very flexible it could be anything). The `EllipsisIter` yield `EllipsisIterItem` which is `Option<Symbol>` + the enum `EllipsisIterValue(Promise)` or `EllipsisIterValue(MissingArg).

@JosiahParry  On this raw low-level  taking the 5th item means accessing whatever symbols and promise is between the 4th and 5th comma-`,`. Imagine an R function called this`f(,,,,x=2,)`the `<Ellipsis>.iter().nth(5)` will yield Symbol 'x' + Promise `2`.

From here on any arg-syntax can be build on top e.g. `list()` `list2()`  `data.frame[,]`or `data.table[]`. We have mostly been interested in the `list2` syntax but any user could re-implement data.table-like syntax because the low-lever iter is public. That is great. Actually that makes me very happy ;)

I use my`List2PromiseIter::new(ellipsis_iter)` which lazily iterates the low-level `EllipsisIter` according to `list2`-syntax rules. It yieds a Result<NamedPromise, extendr_api::Error>.  If `Ok` it means the arg is syntactically valid and contains a `NamedPromise` which a `Option<String>` +  `Promise`. If `Err`it returns the syntax Error.

@JosiahParry  On the level taking the 5th element means  getting name of valid(`Ok`) arg (if any) Option<String> and a `Promise`or getting the error of the invalid arg(`Err`). Extendr_api::Error:missingArg conveniently contains either index or name. The `List2PromiseIter`completely ignores any valid trailing arg, such that a trailing comma does not matter. Same outcome. If the 5th element is a trailing unnamed arg, you cant take it because the iterator has depleted. That is essentially list2.

On the  3rd level we get to iterate over evaluated promises. Promises will either yield a value or throw an Error. I do not recommend use of `promise.eval()`unless the user don't care much about error handling or want to iterate 100k elements. I recommend `fn trycatch_promise(p: Promise) -> Result<Robj> {...}` which will capture the error as an Robj. The Robj is an R `ErrorCondition(message,value,call,...)`. But any user can pick their "eval"-style.

@JosiahParry  On this level getting the 5th elements just means getting a Result<Robj, Error>. Where the error is either a syntax error or the eval error.

Bonus level for 2 and 3!  Often we want to fold/collect the 3rd `Vec<Result<Robj,Error>>`or the 2nd level `Vec<Result<NamedPromise,Error>>` into one outer result like `Result<Vec<Robj>,Error>`.  Then we get e.g. 2nd a Ok Vec<NamedPromis> if all args syntactically ok, or for 3rd level `Vec<Robj>` if both syntax and evaluation was good. If not Err we get the first encountered error. We implement an iterator also which will deplete on first error.

**Comment by sorhawell** [comment]
## Example taking the fifth promise

```rust
/// @export
fn take_fifth_promise(#[ellipsis] dots: Ellipsis) -> Robj {
    let _ = CheckMemRelease::new();
    //get vector Named Promises
    let mut takable_promises = List2PromiseIter::new(dots.iter())
        .takable_promises()
        .expect("no syntax errors in any arg");

    let named_promise = takable_promises
        .get_mut(5)
        .expect("has at least 5+1 elements")
        .take()
        .expect("not already taken");

    let robj = trycatch_promise(named_promise.promise).expect("no eval error");
    let name = named_promise.name;

    rprintln!("fifth value was {robj:?} and had arg name {name:?}");

    robj
}
```

``` r
library(rlang)
library(helloextendr)

f = \(i) \() {cat("func no.",i,"was touched\n");i}  

delayedAssign("test1",f(1)())
test1
#> func no. 1 was touched
#> [1] 1

delayedAssign("f2",f(2)())
delayedAssign("f5",f(5)())
delayedAssign("f6",f(6)())
helloextendr::take_fifth_promise(0,1,f2,3,4,f5,f6) #tada only touched f5
#> CheckMemReleased: OK
#> func no. 5 was touched
#> fifth value was 5.0 and had arg name None
#> [1] 5


#let's try with a regular vector which will touch all promises
delayedAssign("f2",f(2)())
delayedAssign("f5",f(5)())
delayedAssign("f6",f(6)())
c(0,1,f2,3,4,f5,f6)
#> func no. 2 was touched
#> func no. 5 was touched
#> func no. 6 was touched
#> [1] 0 1 2 3 4 5 6
```

<sup>Created on 2023-07-22 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

sorhawell:
@Ilia-Kosenkov I realize your PR is very customizable and the user can do anything. If you see this as a "nightly" feature which has not fully stabillized yet. I think this PR is good to go or we van merge in some of suggestions. I personally suggest to separate concerns in above mentioned levels,  and do not like promise.eval(), but what ever. Maybe this is best revisited after extendr can return errors. You do have single minor bug see comment "ACCEPT TRAILING NAMED ARG  <-- THIS IS A BUG I THINK".

sorhawell:
> P.s.: Perhaps my function signature is suboptimal, it should be 
```rust
fn to_promise(&self) -> Option<&Promise> {todo!()}
```

That might not work as Promise cannot be dereferenced my compiler tells me. I think one have to own the Promise to eval it.

Ilia-Kosenkov:
Thanks for the feedback, I did not quite realize how thorough your QA actually is!

I will go through all of your examples over this weekend. I agree there is a bug with trailing named arg.

I am also curious about memory leekage, so I will be looking at that.

There is one solution to eval that I tried out -- there is a catch_r_error function which can intercept that something went wrong. However we deliberately pause R unwinding there. If we continue unwinding on error, it works closely to how it should, but likely at the cost of not de-allocating rust memory. If I can fix that, that soliton will be superior to running any version of `tryCatch `

JosiahParry:
This has surpassed my knowledge! As long there is a documented way to grab the nth element of dots much like `..1` and `..5` for example that would meet my expectations.

JosiahParry:
One more thought on this before I leave it alone: if this is going to be an experimental API for dots then I think it would be great if there could be some sort of lifecycle-type badge like:   included in the documentation to communicate that it will likely change. It's inclusion is so that more testing of it can be done.

eitsupi:
Any update?

Ilia-Kosenkov:
I got stuck with an intermediate version and since there is an overhaul of some extendr functions, decided not to ibtroduce new hacks. But I clean this and merge as is to provide initial out of the box support for the feature.

JosiahParry:
Thank you, @Ilia-Kosenkov! It's a wonderful feature. More imperfect features is better than no perfect one 💕

JosiahParry:
I am working on https://github.com/JosiahParry/extendr-vctrs and would like to have this feature to have a correct implementation of the `format()` method which requires passing dots. I think this would be a wonderful feature. 

@Ilia-Kosenkov is there anything I can do to assist in merging this?

Ilia-Kosenkov:
The first prototype is blocked by my inability to properly catch errors. If evaluating an element of dots results in error, we want to re-throw it correctly and unwind Rust-allocated resource in-between. We can try to revisit this and perhaps @CGMossa now knows more about unwinding & could help us. I will do a rebase/merge in the evening

Issue #580, opened 2023-07-08
Should not use libR-sys 0.5.0


eitsupi:
See mlverse/tok#4

One of the flavors on CRAN seems to fail to build libR-sys 0.5.0.

yutannihilation:
`  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', /data/gannet/ripley/R/packages/tests-MKL/tok/src/.cargo/registry/src/github.com-1ecc6299db9ec823/libR-sys-0.5.0/build.rs:637:40`

This line:

https://github.com/extendr/libR-sys/blob/a1047c2cd9a2909c2a48ed615657dd18ff18d4fb/build.rs#L637C1-L637C58

yutannihilation:
Maybe `.canonicalize()` added in this pull request make some problem on [the Fedora machine](https://www.stats.ox.ac.uk/pub/bdr/Rblas/README.txt)? If so, I think we can opt-out the canonicalization for non-Windows platforms.

https://github.com/extendr/libR-sys/pull/154

But, as other Fedora machines pass without problems, it might be a problem on the machine setup.

yutannihilation:
Just released my package with libR-sys 0.5.0. Let's see if this fails as well.

https://cran.r-project.org/web/checks/check_results_string2path.html (v0.1.4 is the version)

yutannihilation:
Hmm, strange. I got the acceptance email, but my package is still in the pretest queue for 10 hours. It seems something is wrong on CRAN servers.

eitsupi:
Thanks for taking a look at this!

I did a little research and couldn't find any mention of a problem of `canonicalize()` on Linux.
Is it possible that `r_paths.library` did not exist in the first place instead of `canonicalize()`'s problem? If so, would libR-sys 0.4.0 be expected to not work?
Could you consider using `unwrap_or_default` or something instead of `unwrap` there?

(Perhaps this issue needs to be transferred to extendr/libR-sys?)

yutannihilation:
> Is it possible that `r_paths.library` did not exist

Yeah, it's the only possibility I could come up with. libR-sys assumes the library is on `{R_HOME}/lib`, but it might not be always true. I have no idea in what case R's shared libs are placed in some different location.

https://github.com/extendr/libR-sys/blob/e1a740e25b91a028617afe52e54bc476b45870c1/build.rs#L203C28-L203C28

[...]

PR #586, opened 2023-07-17
Update libR-sys to v0.6.0

yutannihilation:
Close https://github.com/extendr/extendr/issues/580

### DIFF
--- a/extendr-api/Cargo.toml
+++ b/extendr-api/Cargo.toml
@@ -17,7 +17,7 @@ repository = "https://github.com/extendr/extendr"
 rust-version = "1.60"
 
 [dependencies]
-libR-sys = "0.5.0"
+libR-sys = "0.6.0"
 extendr-macros = { path = "../extendr-macros", version = "0.4.0" }
 extendr-engine = { path = "../extendr-engine", version = "0.4.0" }
 lazy_static = "1.4"
--- a/extendr-engine/Cargo.toml
+++ b/extendr-engine/Cargo.toml
@@ -15,7 +15,7 @@ license = "MIT"
 repository = "https://github.com/extendr/extendr"
 
 [dependencies]
-libR-sys = "0.5.0"
+libR-sys = "0.6.0"
 
 [features]
 default = []


yutannihilation:
The failure on Windows-oldrel seems related to ndarray feature.

`  = note: x86_64-w64-mingw32.static.posix-gcc.exe: error: C:/Users/runneradmin/AppData/Local/Temp/Rtmp4umJGk/working_dir/RtmpEdhhH6/file140c5226deb/target\x86_64-pc-windows-gnu\debug\deps\ndarray_example77009d8b8404fecef8caf59b232f854830.ndarray_example77009d8b8404fecef8caf59b232f854830.89c8dac2a25d80e4-cgu.10.rcgu.o: No such file or directory`

I believe the failure isn't related to updating libR-sys, because the diff of the bindings is only the version header.

```
❯ git diff v0.5.0...v0.6.0 .\bindings\bindings-windows-x86_64-R4.2.rs
diff --git a/bindings/bindings-windows-x86_64-R4.2.rs b/bindings/bindings-windows-x86_64-R4.2.rs
index 4f37ff4..8ac2a52 100644
--- a/bindings/bindings-windows-x86_64-R4.2.rs
+++ b/bindings/bindings-windows-x86_64-R4.2.rs
@@ -1,6 +1,6 @@
 /* automatically generated by rust-bindgen 0.64.0 */

-/* libR-sys version: 0.5.0 */
+/* libR-sys version: 0.6.0 */
 /* bindgen clang version: clang version 15.0.7 */
 /* clang-rs version: clang version 15.0.7 */
 /* r version: 4.2.3 */
```
full diff: https://github.com/extendr/libR-sys/compare/v0.5.0...v0.6.0

yutannihilation:
It seems I cannot reproduce this with my local R 4.2.2.

Ilia-Kosenkov:
Can we just re-run it here? Sometimes it is something flaky in builds/runner setup

yutannihilation:
I already tried.

1st run: https://github.com/extendr/extendr/actions/runs/5572135550/jobs/10177843937
2nd run: https://github.com/extendr/extendr/actions/runs/5572135550/jobs/10178386289

Ilia-Kosenkov:
I just noticed that it is actually rextendr misbehaving. I will try to check it when I have time

yutannihilation:
Thanks. Yes, this is about rextendr. I forgot to mention here, but I confirmed rextendr's CI also fails. So, something should have happened after the last commit of rextendr.

https://github.com/extendr/rextendr/pull/301

Ilia-Kosenkov:
Oh no, I know what it is -- path to ndarray library is `260` symbols long. It is an ancient Widows path limitation issue, which is why it is not reproduced locally

yutannihilation:
Oh..., so, should we disable the tests? I'm not sure why it worked before 2 weeks ago, though.

Ilia-Kosenkov:
A good question, right now I am not 100% sure. I wonder if we can temporarily reduce the path length.

Why it misbehaves - also not sure. New version of cargo with different paths?

yutannihilation:
> New version of cargo with different paths?

Ah, I don't know about the details of the update in 1.71, but it might matter. Let's try with 1.70.

yutannihilation:
Confirmed. I ran the CI with Rust 1.70 and 1.71, and it seems 1.70 passes, while 1.71 fails.

https://github.com/extendr/rextendr/actions/runs/5574728108/jobs/10183703676?pr=302

I guess the difference between R 4.2 and R 4.3 (more specifically, Rtools 42 and Rtools 43) is the long path support.

https://blog.r-project.org/2023/03/07/path-length-limit-on-windows/

Ilia-Kosenkov (approved):

PR #587, opened 2023-07-22
Update syn crate to version 2

yutannihilation:
Disclaimer: I don't intend to merge this because it would conflicts with several ongoing pull requests. The main purpose of this is is to show how updating is possible, and (the second purpose is to train myself).

As discussed in https://github.com/extendr/extendr/pull/574, extendr seems to have some motivation to update the syn crate. Here's some code and references:

* [Release note of Syn 2.0](https://github.com/dtolnay/syn/releases/tag/2.0.0)
* [`syn::parse`](https://docs.rs/syn/latest/syn/parse/index.html)
* [`syn::meta::parser`](https://docs.rs/syn/latest/syn/meta/fn.parser.html)
* [`syn::Attribute`](https://docs.rs/syn/latest/syn/struct.Attribute.html)
* [When to choose `Attribute::parse_nested_meta` or `syn::meta::parser`](https://docs.rs/syn/latest/syn/meta/struct.ParseNestedMeta.html#examples)

Sorry for no explanations about the changes. I might write some in future if I get some time.

### DIFF
--- a/extendr-macros/Cargo.toml
+++ b/extendr-macros/Cargo.toml
@@ -7,7 +7,7 @@ authors = [
     "Mossa Merhi Reimert <mossa@sund.ku.dk>",
     "Claus O. Wilke <wilke@austin.utexas.edu>",
     "Hiroaki Yutani",
-    "Ilia A. Kosenkov <ilia.kosenkov@outlook.com>"
+    "Ilia A. Kosenkov <ilia.kosenkov@outlook.com>",
 ]
 edition = "2021"
 description = "Generate bindings from R to Rust."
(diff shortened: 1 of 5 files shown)


eitsupi:
Currently prqlr includes two versions of syn.
https://github.com/cran/prqlr/blob/9b1cb4c827eaf303d5bbbe74df7ec5e5fb320d7d/LICENSE.note#L556-L572

It would be great if this PR is merged.

CGMossa:
I have looked at this PR, and looked at doing the upgrade myself, but when I did it, I faced so many more issues. Maybe I choose the wrong way to upgrade the code.
This looks very succinct.

CGMossa:
Dear @yutannihilation, I like this and I think we should merge this.

The PRs that are currently up are mostly abandoned (due to inaction from our side), or otherwise
will need attention. I think a good path is to work on the PRs, that we all understand, and value, and then
we can unblock our attention and focus on those larger PRs.

Upgrading syn now is better. 

Let me know how I can assist here.

yutannihilation:
Sure. I think this should work.

CGMossa (approved):
LGTM

PR #591, opened 2023-08-24
Return featureful iterator from `Robj`s

### LINKED ISSUE #590: Why 2 `iter` methods for some structs
As far as it goes, Integers, and probably other types, have 2 `iter` methods implemented.

https://extendr.github.io/extendr/extendr_api/wrapper/integers/struct.Integers.html#method.iter
https://extendr.github.io/extendr/extendr_api/wrapper/integers/struct.Integers.html#method.iter-1

Is there any reason for that? The additional iter implementation seem to introduce some handicaps, for example I can't use some `std::slice::Iter` methods like `rev` for it. For example:

This works:
```
let a: Integers = (0..2).map(|z| Rint::from(z)).collect();
let b = <[Rint]>::iter(&a);
let c: Integers = b.rev().copied().collect();
```
This doesn't:
```
let a: Integers = (0..2).map(|z| Rint::from(z)).collect();
let b = a.iter().rev();
```

CGMossa:
Essentially, the issue is

```rust
let a: Integers = (0..2).map(|z| Rint::from(z)).collect();
let b = <[Rint]>::iter(&a);
let c: Integers = b.rev().copied().collect();
```
But this line fails:
```rust
let d = a.iter().rev();
```

I believe this is because we use `impl Iterator<Item = &TYPE>`,
which means trait impls on the underlying slice-iterator aren't passed on.
This can be remedied by using `std::slice::Iter` as a return type instead.

Motivated by a discussion in discord by @daniellga, as seen in #590.

Fixes #590

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,8 @@
 
 - [**either**] `TryFrom<&Robj> for Either<T, R>` and `From<Either<T, R>> for Robj` if `T` and `R` are themselves implement these traits. This unblocks scenarios like accepting any numeric vector from R via `Either<Integers, Doubles>` without extra memory allocation [[#480]](https://github.com/extendr/extendr/pull/480)
 - `PartialOrd` trait implementation for `Rfloat`, `Rint` and `Rbool`. `Rfloat` and `Rint` gained `min()` and `max()` methods [[#573]](https://github.com/extendr/extendr/pull/573)
-- Iterators from `Robj` now implements `DoubleEndedIterator` and `ExactSizeIterator`
-etc. allowing for more combinators to be used [[#591]](https://github.com/extendr/extendr/pull/591)
+- Iterators from `Robj` now implements `DoubleEndedIterator` and `ExactSizeIterator` allowing for more combinators to be used
+    [[#591]](https://github.com/extendr/extendr/pull/591)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,7 @@
 
 - [**either**] `TryFrom<&Robj> for Either<T, R>` and `From<Either<T, R>> for Robj` if `T` and `R` are themselves implement these traits. This unblocks scenarios like accepting any numeric vector from R via `Either<Integers, Doubles>` without extra memory allocation [[#480]](https://github.com/extendr/extendr/pull/480)
 - `PartialOrd` trait implementation for `Rfloat`, `Rint` and `Rbool`. `Rfloat` and `Rint` gained `min()` and `max()` methods [[#573]](https://github.com/extendr/extendr/pull/573)
-- Iterators from `Robj` now returns `std::slice::Iter` / `std::slice::IterMut`,
-which implements `DoubleEndedIterator`, `ExactSizeIterator`, `FusedIterator`,
+- Iterators from `Robj` now implements `DoubleEndedIterator` and `ExactSizeIterator`
 etc. allowing for more combinators to be used [[#591]](https://github.com/extendr/extendr/pull/591)
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 - [**either**] `TryFrom<&Robj> for Either<T, R>` and `From<Either<T, R>> for Robj` if `T` and `R` are themselves implement these traits. This unblocks scenarios like accepting any numeric vector from R via `Either<Integers, Doubles>` without extra memory allocation [[#480]](https://github.com/extendr/extendr/pull/480)
 - `PartialOrd` trait implementation for `Rfloat`, `Rint` and `Rbool`. `Rfloat` and `Rint` gained `min()` and `max()` methods [[#573]](https://github.com/extendr/extendr/pull/573)
+- Iterators from `Robj` now returns `std::slice::Iter` / `std::slice::IterMut`,
+which implements `DoubleEndedIterator`, `ExactSizeIterator`, `FusedIterator`,
+etc. allowing for more combinators to be used [[#591]](https://github.com/extendr/extendr/pull/591)
 
 ### Fixed
 
--- a/extendr-api/src/robj/mod.rs
+++ b/extendr-api/src/robj/mod.rs
@@ -426,7 +426,7 @@ impl Robj {
     ///     assert_eq!(num_na, 1);
     /// }
     /// ```
-    pub fn as_logical_iter(&self) -> Option<impl Iterator<Item = &Rbool>> {
+    pub fn as_logical_iter(&self) -> Option<std::slice::Iter<Rbool>> {
         self.as_logical_slice().map(|slice| slice.iter())
     }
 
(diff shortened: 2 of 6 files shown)


CGMossa:
Ready for review. I think there is a value in adding this, and a more thorough approach is not easily feasible;

The main issue is that `Deref` / `DerefMut` isn't flexible enough to determine `Target` on what happens
on the call-site. Therefore, we use our own `.iter`, `.iter_mut`, as the specified type there goes back to
`.as_typed_slice`, thus giving the flexibility seen before.

I'm open to suggestions, especially in an issue or follow-up PR.

CGMossa:
Fixes #590

Ilia-Kosenkov:
The reason `.rev()` does not work is because it requires `impl DobuleEndedIterator`. You can likely solve this by returning `Option<impl DoubleEndedIterator<Item = &T>>`. It might be better to go this way (if it solves problems) to keep it abstract.
Also, I see explicit changes to `as_real_iter()` and `as_logical_iter()`, but aren't there other iters that should be updated accordingly?

CGMossa:
The backbone of most conversions is `AsTypedSlice`, meaning there is really no need for an "abstract" iterator return type.

I'd like to clarify that we are not returning an abstract type, but a minimal type.

Fixing things so only `.rev()` is possible is an odd choice if you ask me. I haven't checked what other iterator combinators are available, but I'm sure this will play better with packages like `itertools` as well.

Only iterators remaining are the ones for character strings. I'm not changing those because strings are special.

I hope you can give this another look, and see that it is a free-lunch benefit.

Ilia-Kosenkov:
Then perhaps returning a custom iterator rather than returning a slice iterator could be a way to go? As for other methods, most of them require either `Iterator` (one way from the start), `DoubleEndedIterator` (from the start or from the end), or `ExactSizeIterator`, which knows its size.

CGMossa:
> Then perhaps returning a custom iterator rather than returning a slice iterator could be a way to go? As for other methods, most of them require either `Iterator` (one way from the start), `DoubleEndedIterator` (from the start or from the end), or `ExactSizeIterator`, which knows its size.

What problem does this solve? It's a slice iterator in the end, so we pass that on.

If there is no immediate benefit for it.

Ilia-Kosenkov:
`Integers` and friends also wrap altreps, so if we are honest, the iterator is not always a `slice` iterator. It is that the current implementation just materializes everything (if I read it correctly). If we return an `impl Trait`, we can then properly handle alterp vs no alterp.

But I am not against your current approach, just want to understand the whole picture.

CGMossa:
`ALTREP` sounds like a good reason not to return the slice-iterator.

Could you list steps you see needed to implement a custom iterator type?
I'll try an approach, but I'm flying blind here..

CGMossa:
I've just opted for the addition of `DoubleEndedIterator` and `ExactSizeIterator`. This is a little better than having to stick with slice-iterator, now that we have `ALTREP` support (are we sure it works?).
But this still says that the altrep version will promise `DoubleEndedIterator` and `ExactSizeIterator`, which I guess are innocent enough.

Ilia-Kosenkov:
The issue here is that if we want to support altrep as is, I believe the best we can do is `ExactSizeIterator` (though I am not sure all altreps have known lengths). So we could expose generic iterator with minimum features and optional rich iterator that may or may not be obtained based on the underlying type.

CGMossa:
This is getting rather unpleasantly complicated.
At the moment the `iter` implementation acts more like an `into_iter`, 
as we don't get references back. I could imagine that an altrep-iterator would be one with references;

daniellga:
Wouldn't it be easier if there were separated types for altrep? Like IntegersAltRep, DoublesAltRep, etc? Or is this iter thing the only  place where an altered and nonaltrep implementation are divergent?

CGMossa:
I think the issue is that `Robj` is the funnel. Maybe we are wrong in this perception, but everything is an SEXP, and `Robj` encodes how `SEXP`s should be treated in Rust. Entirely. From the calling of bindings, to the protection in `ownership`-module, and so on.

AltRep will have to go through that as well. 

But... the whole extendr API is not made for the altrep thing. We have some support, that I don't understand, for it. But everything right now forces the materialisation of values.

I cannot make a PR that can promise not to break ALTREP API support in the future, and frankly I don't think it is supported *fully* yet either.

CGMossa:
I deleted my reply because I thought it was double posted. It wasn't. If anyone received my reply by email please feel free to paste it in again. I've lost it.

daniellga:
Sure. Here's your reply:

> I think the issue is that Robj is the funnel. Maybe we are wrong in this perception, but everything is an SEXP, and Robj encodes how SEXPs should be treated in Rust. Entirely. From the calling of bindings, to the protection in ownership-module, and so on.
> 
> AltRep will have to go through that as well.
> 
> But... the whole extendr API is not made for the altrep thing. We have some support, that I don't understand, for it. But everything right now forces the materialisation of values.
> 
> I cannot make a PR that can promise not to break ALTREP API support in the future, and frankly I don't think it is supported fully yet either.

JosiahParry:
I'm using this as a learning exercise for myself. I'd like to try and summarize the problem. 

**The issue:** Vector type objects such as `Doubles`, `Integers`, `Complexes` etc only implement `Iterator`. It would be nice if they also implemented `DoubleEndedIterator` and `ExactsizeIterator`. 

This PR adds both `DoubleEndedIterator` and `ExactSizeIterator` for objects with scalar type iterator components. This is awesome and a big win.

However, there is hesitation in taking this approach because extendr provides ALTREPs of these same types and it cannot be guaranteed that this change will break them. 

If this is the case, then it sounds like there needs to be additional tests for iterators on altreps?

If i understand correctly, `ExactSizeIterator`s require a `len()` method. For ALTREPs, the `AltrepImpl` requires the `length()` traits so that should not necessarily be a problem. The problem then is for `DoubleEndedIterator` which requires `next_back()` and `nth_back()` methods that return `Option<Self::Item>`.

For extendr "native" types this can be done simply. But may not be guaranteed for ALTREPs. My understanding right now is that ALTREPs that have been created by extendr do not have these traits _anyways_ so there wouldn't be any breaking changes by adding them. Additionally, ALTREPs require manual trait implementations. If the default iterator approaches do not work for an ALTREP, is the user prevented from creating their own `DoubleEndedIterator` or `ExactSizeIterator` for their type? 

To my novice eyes it seems like an easy win without any breaking changes (assuming CI passes).

CGMossa:
@Ilia-Kosenkov can you respond to @JosiahParry on this?

Ilia-Kosenkov:
Sure. I somehow missed it. So you convinced me about `ExactSizeIterator`, if `length()` is guaranteed to be available. But `DoubleEndedIterator` is more problematic. I found [this](https://stat.ethz.ch/pipermail/r-devel/2019-September/078481.html) and [that](https://svn.r-project.org/R/trunk/src/main/sort.c), which suggests there is a way to optimally iterate over ALTREPs using regions, and I do not remember if we do that now. Now what we do is we use a slice, which materializes all altreps (which is bad of course). Maybe we should go with another solution, having a separate `DoubleEndedIterator` which may or may not be available for a given type. Currently all types will just support it because of materialization, but in the future we could return `None` if underlying type is `ALTREP` so no materialization happens, and iterate over regions wih forward-only iterator, preserving memory (so that you can do something like `sum(1:9999999999999999999999999)` without actually allocating vector).

PR #594, opened 2023-08-25
Add `[T; N]` for `Robj`.

CGMossa:
Rust added support for `[T; N]`, but we haven't taken advantage of it yet.

Previous attempts got scope creeped, thus this time we'll stick with a very small PR for now.

### DIFF
--- a/extendr-api/src/robj/into_robj.rs
+++ b/extendr-api/src/robj/into_robj.rs
@@ -663,23 +663,6 @@ where
     }
 }
 
-// We would love to do a blanket IntoIterator impl.
-// But the matching rules would clash with the above.
-macro_rules! impl_from_iter {
-    ($t: ty) => {
-        impl<'a, T> From<$t> for Robj
-        where
-            Self: 'a,
-            T: Clone + 'a,
-            T: ToVectorValue,
-        {
-            fn from(val: $t) -> Self {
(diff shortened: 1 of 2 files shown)


extendr-api/src/robj/into_robj.rs
```diff
@@ -721,28 +704,49 @@ macro_rules! impl_from_as_iterator {
...
+}
+
+impl<T: ToVectorValue> From<Vec<T>> for Robj {
+    fn from(value: Vec<T>) -> Self {
+        let len = value.len();
+        fixed_size_collect(value.into_iter(), len)
+    }
+}
 
 impl_from_into_iter! {&'a [T]}
```
Ilia-Kosenkov, line 751: If this is the only use of the macro now, should it be kept here, or maybe we can drop the macro in favor of just writing it out as is?

CGMossa: I did make a PR where I tried to get rid of everything. Frankly, I'm just hoping for hyper scoped PRs, that get merged, and then I'll do whatever. But I'll take a look at this now.

CGMossa: Oh this can be inlined as you say.. Oops.. I got rid of it now.

Ilia-Kosenkov (approved):
Looks clean & mich more readable.

PR #595, opened 2023-08-25
Subset of `cargo clippy` fixes

CGMossa:
Again, with a new version of Rust, comes many new clippy/rustc lints.
I've gone through some of them, and I propose we merge these.

There are more, meaning `cargo clippy` won't be successful with this PR, but
these make sense in my opinion.


Ilia-Kosenkov (approved):
If tests pass, I guess it does not break anything. Thanks!

PR #596, opened 2023-08-26
this fails.. why?

CGMossa:


CGMossa:
Copy from previous message: 
This revealed an issue. 
```rust
// Only valid i32 are 0, 1 and i32:MIN.
let minus: Result<Logicals> = vec![-1].try_into();
dbg!(&minus);
// // assert!(minus.is_err());
let minus: Result<Logicals> = vec![-1, -24].try_into();
dbg!(&minus);
```

```
running 1 test
[extendr-api/src/wrapper/logicals.rs:175] &minus = Ok(
    TRUE,
)
[extendr-api/src/wrapper/logicals.rs:178] &minus = Ok(
    [thread 'wrapper::logicals::tests::test_vec_i32_logicals_conversion' panicked at 'called `Option::unwrap()` on a `None` value', extendr-api/src/wrapper/logicals.rs:24:1
stack backtrace:
```

> Translation: A -1 gets converted to TRUE, and non-scalar one panics while generating those elements.

CGMossa:
The issue is that `ToVecterValue` for `i32` was being used when trying to construct `Logicals`. Circumventing this wasn't trivial.
Major conflict here is that it is not clear how to convert `i32` to `Rbool`. Should only `0, 1, i32::MIN` be accepted?
Note that since we cannot impl both `From<i32>` and `TryFrom<i32>` for the same time, I've made alternative naming that should probably be bikeshed some more.

CGMossa:
@Ilia-Kosenkov Can I ask for you to provide your opinion and guideline on this? I'm really a bit indecisive about the whole thing. Being able to **just** go from `Vec<i32>` to whatever equivalent in R you want seems like an obvious UX benefit, but is this the way forward?

Ilia-Kosenkov:
How about we implement `From<i32> for Rbool`, which does the same as in C languages, i.e. `0` is `false`, not `0` is `true`, and `i32::MIN` is `NA`?

CGMossa:
Good idea.
That's what I implemented first hand.

But if is different from `from_values` which follows Rs convention.

extendr-api/src/scalar/rbool.rs
```diff
@@ -3,6 +3,9 @@ use crate::scalar::Scalar;
 use crate::*;
 use std::convert::TryFrom;
 
+/// Integer (`i32`) values that [`Rbool`] is allowed to take.
+const VALID_VALUES: [i32; 3] = [0, 1, i32::MIN];
```
JosiahParry, line 7: Is this `const` scoped to this file only? If so, that checks out. If not, I think it should be renamed something like `VALID_RBOOL_VALUES`

extendr-api/src/wrapper/logicals.rs
```diff
@@ -41,6 +41,21 @@ impl Logicals {
     }
 }
 
+impl TryFrom<Vec<i32>> for Logicals {
+    type Error = Error;
+
+    fn try_from(value: Vec<i32>) -> std::result::Result<Self, Self::Error> {
+        let value_rbools: std::result::Result<Vec<Rbool>, _> = value
+            .into_iter()
+            .map(|v| Rbool::try_from_strict(v))
```
JosiahParry, line 50: We want the strict try_from here because if it can't convert _everything_ into Rbool we want Err? 
What's somewhat frustrating is that I would want `From<Vec<i32>> for Logicals` to behave like `from_values()` but we can't have that, right? They would be conflicting implementations?

extendr-api/src/scalar/rbool.rs
CGMossa: I think so, otherwise it would have to be a `pub const`.

extendr-api/src/wrapper/logicals.rs
CGMossa: I know it is too much to ask for.. But can you make an overview of the choices? Because @Ilia-Kosenkov said https://github.com/extendr/extendr/pull/596#issuecomment-1701648049, but
`from_values`, uses the current `From<Rbool>` rule, which is.. I'mma guess not the same as the proposed version with @Ilia-Kosenkov 's

Ilia-Kosenkov: At this point, I have no strong opinion. It could be either `From` using the logic I mentioned (C-like), or it could be `TryFrom`, which would be more 'precise'. Note that in the case of `From`, we ensure that the resulting `Rbool` is either `1`, `0`, or `NA` regardless of the `i32` input.

CGMossa: so.. Merge this?

extendr-api/src/scalar/rbool.rs
Ilia-Kosenkov, line 7: Should we reference NA value here directly? NA value is an implementation detail & hardcoding it is a bit of an odd thing.

JosiahParry: Could you use `Rbool::na().inner()` or something like that to avoid hard coding?

PR #458, opened 2023-01-23
Use the 2021 edition

yutannihilation:
https://github.com/extendr/extendr/issues/457

PR #597, opened 2023-08-27
Set cargo's dependencies resolver to latest version

CGMossa:
Warning:
```
cargo doc
warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
 
 members = [
 	"extendr-api",


yutannihilation (approved):
It seems this is my oversight when I did https://github.com/extendr/extendr/pull/458. Thanks for catching!

PR #598, opened 2023-08-27
[internal] Add `with_r` function for testing instead of `test!`

CGMossa:
Draft, and open for discussion;
This is a feature for testing `extendr`, and likely isn't to be part of the user experience.

Today, one has to invoke `test!` to use R-specific functions in Rust.
Macros are horrendous for the developer experience and especially Rust Analyzer
isn't particularly good with that.

Secondly, `cargo fmt` doesn't do anything with `test!`.

I believe, that we can get far enough with a function instead. The pattern is

```rust
fn with_env(f: impl FnOnce()) {
  start();
  f();
  end();
}
```

This is essentially what is in `test!`. What else is in `test!` is that the `f` is
a `impl FnOnce() -> Result<()>`. This is to enable `?` within `test!` macro.
First, I'd say that not everyone writing tests knew this, as the tests are full of `.unwrap()`. Also, the `?` here does indeed merely `.unwrap()`. While this is a
home-made `try`-block, I don't see a particular advantage in having it.

Quite frankly, the typical Rust user rarely use `?` in tests anyways. 

Secondly, the `extendr-engine` doesn't do any parallel / concurrency for the initated R instances. Therefore I've added an `atomic` to keep track of this across threads. 

Lastly, I've ported over many of the tests to `test!`, and I have made modules / separate
files of the tests found in extendr to their own files.


CGMossa:
I'd like to clarify that the high diff count is due to tests being moved around, and switching between `test!` and `with_r(||`; It is necessary to convert some of these to see it in action.

CGMossa:
For `extendr-engine`, an embedded r instance is initiated / started, but it is never shut off. Presumably this is fine for tests;
I've experimented with using an atomic to keep track of r instances and shut them off. This didn't work immediately. I suspect we have concurrency bugs in extendr.
For now, I'll make a version of `with_r(||` that works as well as `test!`.
Next I'll experiment with adding a `dtor` from the crate `ctor`, and see if that works,
to atleast invoke the `end_r` and clean-up the internal r-instance.

That would be it for this PR, because things are too broken to be improved further.

JosiahParry:
Can this be closed because of #599?

PR #599, opened 2023-08-28
[internal] Function-version of `test!` 

CGMossa:
Adding `with_r` a function-version of `test!`. 

- This function doesn't allow the use of `?` inside of tests like `test!` does;
- It's not a macro, thus rust-analyzer, `cargo fmt`, and other tools behave nicely with it

Please see #598 for tests where `with_r` has replaced many `test!` cases. 

There are many other issues and many things to dwell on, but for now, let's take a look at this.

### DIFF
--- a/extendr-engine/src/lib.rs
+++ b/extendr-engine/src/lib.rs
@@ -63,7 +63,7 @@ pub fn end_r() {
 }
 
 /// Ensures that an embedded R instance is present when evaluating
-/// `f`. 
+/// `f`.
 pub fn with_r(f: impl FnOnce()) {
     start_r();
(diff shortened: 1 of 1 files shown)


extendr-engine/src/lib.rs
```diff
@@ -62,6 +62,15 @@ pub fn end_r() {
     }
 }
 
+/// Ensures that an embedded R instance is present when evaluating
+/// `f`.
```
CGMossa, line 66:
```suggestion
/// `f`.
```

Ilia-Kosenkov (approved):

PR #600, opened 2023-08-29
Add ALTLIST 

JosiahParry:
ALTLIST representations were recently added to base R. 
This PR adds functionality to create an ALTREP for List objects.


JosiahParry:
I think I may have messed up my fork sync. If that's the case let me know and I will blow it all up and make a new branch :P

JosiahParry:
The CI errors are due to not finding the ALTLIST functions in libRsys which should be available for R 4.1 onwards. They're present in 0.6.0 as well. https://docs.rs/libR-sys/latest/libR_sys/?search=altlist

yutannihilation:
Yes, one small correction is the version with ALTLIST support is R 4.3.

CGMossa:
@JosiahParry I hope it is okay: But I added the feature-gate. Let's see if it passes tests.

JosiahParry:
I only just finished eating breakfast! Y'all are fast. Thank you both for the comments. @CGMossa i see you pushed some commits to address them. Thank you!

extendr-api/build.rs
```diff
@@ -26,4 +26,8 @@ fn main() {
     if &*major >= "4" && &*minor >= "2" {
         println!("cargo:rustc-cfg=use_r_ge_version_15");
     }
+
+    if &*major >= "4" && &*minor >= "3" {
+        println!("cargo:rustc-cfg=use_r_version_16");
```
yutannihilation, line 31:
```suggestion
        println!("cargo:rustc-cfg=use_r_altlist");
```

Ah, sorry that I wasn't very clear. I meant you should use some meaningful name for the feature.

CGMossa: I can rename it, but this isn't a feature where you *enable* `altlist`. It is more like a gate.
A `use_r_altlist` would enable it or you'd expect it to disable it.

What do you think? Let me know, and I'll change it.

JosiahParry: I've only written 1 feature flag before so I'm not an expert. Is there any way to enable it such that it's on by default for R >= 4.3 but disabled otherwise?

CGMossa: That's what it is right now.
But the name I'd say is confusing from a "logic"/words point of view. If @yutannihilation disagrees,
then I'll rename it right away.. I initially thought it should be `use_altlist` too..

JosiahParry: If its okay, I'd like to have `use_r_altlist` setting it to false is clear what the decision is.

CGMossa: Done

yutannihilation (approved):
Looks good! Could you add an item to CHANGELOG.md?

JosiahParry:
@yutannihilation pushed!

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Adds support for `ALTLIST` allowing users to represent structs as R list objects
```
Ilia-Kosenkov, line 5: Incorrect bullet point placement. We follow [this](https://keepachangelog.com/en/1.0.0/), according to which you should put this sentence under `### Added`, like so:
```
### Added
- `ALTLIST` support, allowing users ...
```

extendr-api/src/wrapper/altrep.rs
```diff
@@ -1005,6 +1019,39 @@ impl Altrep {
...
+        name: &str,
+        base: &str,
+    ) -> Robj {
+        #![allow(non_snake_case)]
+
+        single_threaded(|| unsafe {
+            let class = Altrep::altrep_class::<StateType>(Rtype::List, name, base);
+            let class_ptr = R_altrep_class_t { ptr: class.get() };
+
+            unsafe extern "C" fn altlist_Elt<StateType: AltListImpl + 'static>(
```
Ilia-Kosenkov, line 1033: Any reason to spell it `altlist_Elt` and not `altlist_elt`?

extendr-api/src/wrapper/altrep.rs
```diff
@@ -1005,6 +1019,39 @@ impl Altrep {
...
+                x: SEXP,
+                i: R_xlen_t,
+            ) -> SEXP {
+                Altrep::get_state::<StateType>(x).elt(i as usize).get()
+            }
+
+            unsafe extern "C" fn altlist_Set_elt<StateType: AltListImpl + 'static>(
```
Ilia-Kosenkov, line 1040: Same question about the name.

extendr-api/src/wrapper/altrep.rs
```diff
@@ -455,6 +455,16 @@ pub trait AltStringImpl {
...
 }
 
+#[cfg(use_r_altlist)]
+pub trait AltListImpl {
+    /// Get a single element from this vector
+    // a single element of a list can be any Robj
+    fn elt(&self, _index: usize) -> Robj;
+
+    /// Set a single element in this list.
+    fn set_elt(&mut self, _index: usize, _value: Robj) {}
```
Ilia-Kosenkov, line 465: Why do we have a default implementation for `set_elt()` here?

extendr-api/tests/altrep_tests.rs
```diff
@@ -231,7 +232,51 @@ fn test_altstring() {
...
+
+        // confirm it is altlist
+        assert!(obj.is_altlist());
+
+        // confirm method is accurate
+        assert_eq!(obj.len(), 3);
+
+        // convert to a list and test the .elt() method
+        let l = List::try_from(obj.into_robj()).unwrap();
+        let li = l.elt(1).unwrap();
```
Ilia-Kosenkov, line 278: I think it should be just trivially indexable with `[1]`, no? If `L` is of `Lilst` type and it does not support `[]`, it is an oversight.

Ilia-Kosenkov (commented):
Cool! If this is working reliably (and I'd say we might need more tests to be 100% positive), we need an update in documentation (README or otherwise).

extendr-api/src/wrapper/altrep.rs
yutannihilation: I think this just follows other altrep-imlementation. 
I cannot decide which is good, but making it read-only by default is probably not a very strange idea.

Ilia-Kosenkov: But it is not readonly? If you implement a trait with default method implementation, you get access to this method anyway? Or you can pass around this type as `T : Trait` or `impl Trait`, which exposes all methods. In which case, settings becomes a `noop`, which could be misleading. If it is our go-to approach, then no objection, but it is worth investigating how it works and why we chose this approach.

yutannihilation: I have no idea, sorry. My main point was that this isn't @JosiahParry's fault so we should not block for this reason.

> it is worth investigating how it works and why we chose this approach.

I agree with you. Let's merge this and discuss in another issue.

JosiahParry: This is copying evrey other altrep trait implementation. The `set_elt()` and `elt()` are required for the ALTLIST.

I think the atlrep story is unfnished and can be improved upon a lot. This is a baby step PR towards that direction.

extendr-api/tests/altrep_tests.rs
JosiahParry: Hm, doing so results in the following: 

Error: cannot take a writable DATAPTR of an ALTLIST [class: li, pkg: mypkg]
Fatal error: unable to initialize the JIT

JosiahParry: Do you have an idea of what this means?

extendr-api/src/wrapper/altrep.rs
JosiahParry: Following convention of other altrep implementations which I think follows the convetion in the C file
https://github.com/wch/r-source/blob/49de65f04d675e485e783685e9786139fcb3b4c7/src/main/altrep.c#L202C34-L202C34

extendr-api/tests/altrep_tests.rs
CGMossa: >
    /* Disallow taking the writable `DATAPTR()` of an ALTLIST. This
       check could be moved to `DATAPTR()` to catch more faulty
       usages. */

Ilia-Kosenkov:
As a final step, it would be nice to also jump into `{extendrtests}` and write an integration test for `ALTLIST` there. It will ensure that it is handled correctly from R perspective & also provide an example implementation to which we can refer people.

JosiahParry:
@Ilia-Kosenkov Done! I think there needs to be a series of follow up issues to make the ALTREP experience better. There is not any way (that I am aware of) of capturing the Altrep and converting back to the original struct type.

yutannihilation:
Sorry, I forgot that the CI run needs an approval. I just approved it, but it seems tests you added fail. I think

* Probably you forget to add the altlist module to `lib.rs`
* The compilation and test won't work on oldrel.

JosiahParry:
@yutannihilation thanks for catching that! I've committed the change to `lib.rs`. Is there a way to ensure that function won't get compiled for oldrel? Would adding a feature gate work?

yutannihilation:
Yes, I think you need these feature gate and `build.rs` like below. I forgot `DEP_R_R_VERSION_*` is available without directly adding libR-sys to the dependencies, but it should be possible.

https://github.com/extendr/extendr/blob/master/extendr-api/build.rs

I hope there's some simpler solution, but I don't come up with a nice one yet.

JosiahParry:
Thank you @Ilia-Kosenkov and @yutannihilation for suggesting the tests. The ALTLIST integration test failed with a panic from `extendr-api/wrapper/altrep.rs` line 522. To investigate further, I have added additional integrations tests for ALTINTEGER and ALTSTRING. The same failure exists for ALTINTEGER. ALTSTRING results in a segfault in the memory leak integration tests. 

I have pushed these integration tests but have commented them out. This is because it will cause CI to _always_ fail. 

In an effort to avoid scope creep, I suggest that this PR be merged as it mimics other ALTREP implementations and works similarly to the others. In addition, it passes the rust tests. This would be an incremental improvement to extendr. The PRs scope is limited but has revealed a place where extendr can be further improved. 

I also suggest that there be an immediate follow up issue / PR to identify the source of the issue and rectify. I believe that making ALTREP work for extendr can open up doors for the project.

Ilia-Kosenkov:
Judging by the error message, the issue should be 'fun' -- pointer misalignment might be a bug in our implementation or in R. I will try to take a look when I get an opportunity.

extendr-api/src/wrapper/altrep.rs
```diff
@@ -455,6 +455,16 @@ pub trait AltStringImpl {
     }
 }
 
+#[cfg(use_r_altlist)]
+pub trait AltListImpl {
+    /// Get a single element from this vector
+    // a single element of a list can be any Robj
```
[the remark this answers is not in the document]
CGMossa: @JosiahParry do you see this?

CGMossa (commented):
Maybe these comments I've left has been invisible to you?

yutannihilation:
As the CI passes, can we merge this for now?

yutannihilation:
Thanks. I'll address the test failures.

PR #601, opened 2023-08-29
[internal] Shutdown R after tests

CGMossa:
`extendr-engine` never enacts the `end_r` function. This means that we never see the consequences of the finalizers for `ExternalPtr`, or for `ALTREP`. 

Also, if our tests make temporary directories/files through R those are never cleaned properly.

The only lean way to do this I found is with `ctor`, a crate that provides a way to add "init function" for a library; In this case, we use the "shutdown" for a library, i.e. `dtor`.

This is a draft for now, as it may uncover unforseen bugs.

### DIFF
--- a/extendr-engine/src/lib.rs
+++ b/extendr-engine/src/lib.rs
@@ -53,7 +53,7 @@ pub fn start_r() {
 
 /// Close down the R interpreter. Note you won't be able to
 /// Restart it, so use with care or not at all.
-pub fn end_r() {
+fn end_r() {
     unsafe {
         //Rf_endEmbeddedR(0);
         R_RunExitFinalizers();
--- a/extendr-engine/src/lib.rs
+++ b/extendr-engine/src/lib.rs
@@ -71,6 +71,13 @@ pub fn with_r(f: impl FnOnce()) {
     // is no `end_r()` call here.
 }
 
+#[ctor::dtor]
+fn shutdown_r() {
+    if START_R.is_completed() {
+        end_r();
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/extendr-engine/Cargo.toml
+++ b/extendr-engine/Cargo.toml
@@ -16,6 +16,7 @@ repository = "https://github.com/extendr/extendr"
 
 [dependencies]
 libR-sys = "0.6.0"
+ctor = "0.2.4"
 
 [features]
 default = []


CGMossa:
While test is a draft, it does show various errors that happens, simply because we are cleaning up after tests.
If anyone wants to dig deeper to these issues, they are more than welcome. Tagging expert bug chasers: @yutannihilation @Ilia-Kosenkov

yutannihilation:
No idea. Again, this is one of the reasons why I think extendr should give up its embedded usages. While you think it's a shortcut to write tests without a real R session, on the other hand, it consumes your time for investigating such errors like this.

CGMossa:
You're absolutely right. But the extendr-developer is blind to what happens when R shutsdown,
whilst containing extendr-things.

We tracked a bug down yesterday, in which R would segfaults, on exit. Turns out it was an `R --save` exit.
And then an altrep-string object was being asked for string pointer (by `rlang`, due to reasons), and then
an erroneous assumption about this pointer was made, and crash.

My attitude towards this is frankly either we improve the testing, or as you say, we take it out.
But then the extendr-developer experience is indeed strange.


`jlrs` the Julia/Rust wrapper project started out being a "run julia in R" first, and was that for atleast a year,
before providing bindings for Rust things in Julia. So there is that perspective as well...

CGMossa:
@yutannihilation wonderful! Your latest find made this PR pass tests! Thank you so much.

yutannihilation:
Ah, was that the cause? Glad to know it helps here.

extendr-engine/src/lib.rs
```diff
@@ -71,6 +71,13 @@ pub fn with_r(f: impl FnOnce()) {
     // is no `end_r()` call here.
 }
 
+#[ctor::dtor]
+fn shutdown_r() {
+    if START_R.is_completed() {
+        end_r();
+    }
+}
```
yutannihilation, line 79: If I understand correctly, this means `shitdown_r()` is always called regardless the user already ended the session (e.g. by calling `end_r()` by themselves) or not. Is it safe to call `end_r()` multiple times?

CGMossa: I hope the is completed will ensure that end r is only called if the cell was called.
That happens only once.
And the shutdown is in a dtor which is a function that is executed only when extendr engine is shutdown.

yutannihilation: Yes, I know shutdown_r() is ensured to be called only once. But, as end_r() is a public function, users can call it freely, right?

CGMossa: Right. I should also make it private.
It's not safe to call it multiple times.

yutannihilation: Or, probably like start_r(), you can wrap it with std::sync::Once().

CGMossa: The thing is, `test!` is defined in `extendr-api`, which uses `start_r`.
And it is done like that because of reasons due to.. keeping out non-API symbols from the library file.
Atleast in theory;

CGMossa: I don't think `end_r` is safe to run multiple times at all. I don't think `test!` should contain `end_r` either.
I think it should be private and with this, end_r is appropriately ran at the end of the program.

Unless one is doing something particularly tricky, with threads, and each thread spawning its own process.. then this setup will suffice maybe.

yutannihilation: Thanks, makes sense.

yutannihilation (approved):

PR #602, opened 2023-08-29
`ExternalPtr` clean-up type-name

CGMossa:
Great care is taken to deallocate the Rust object from `ExternalPtr`, but what about `tag` and `prot`?

- Using `r(())` we get `R_NilValue` that is protected by `ownership`-module.

But this is convoluted.

- `ExternalPtr` protects values,
`tag` and `prot`, so they need to
be "released" to unprotected.

### DIFF
--- a/extendr-api/src/wrapper/altrep.rs
+++ b/extendr-api/src/wrapper/altrep.rs
@@ -496,9 +496,9 @@ impl Altrep {
             }
 
             let ptr: *mut StateType = Box::into_raw(Box::new(state));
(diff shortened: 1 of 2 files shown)


CGMossa:
Okay, I've pushed a new change. I've decided that it is important to set the embedded pointer in `ExternalPtr` to C `NULL` after we have dropped the value within. 
I believe this is good if for some reason something does a last minute cleaning of resources in R.

Ilia-Kosenkov (approved):

PR #603, opened 2023-08-29
[internal] Migrate tests to `with_r`

CGMossa:
This superceeds #598. 

It is a mouth-ful but might be better in the long run.


Ilia-Kosenkov:
I restarted the failed job. It may be unstable because it is `nightly` and we run doctests in x-compilation regime. However, since I missed the whole discussion on the topic, what is the point of replacing existing tests with `with_r()`?

Ilia-Kosenkov:
Honestly, I am not convinced existing tests should be migrated -- I prefer `test!{}` since it is easier to use (except for formating/syntax highlighting). Also, `[test_with_r]` is bad because it will disallow usage of other test frameworks like `rstest`, which is super useful. Ideally, it would be nice to integrate `extendr` with `rstest` so that test cases are evaluated in the `R` session, which is not possible right now.

CGMossa:
This is not in a hurry. Sleep on it.

CGMossa:
I still believe that tests should be formatted properly, and completion and syntax highlighting should be prioritised in tests, but this is a point I don't want to fight for. I'm closing this PR down, but it can definitely be brought up in the future.

PR #606, opened 2023-08-31
ExternalPtr are no longer stringly typed

CGMossa:
Currently, `tag` in `ExternalPtr` is set to a string containing `std::any::type_name`.
This means, that every time an `Robj` is to be interpreted as an `ExternalPtr`, a
string comparison occurs.

R recommends that `tag` is set to an `install("Class name")`, which is a symbol, which is a pointer, and thus pointers are faster to compare than strings.

Proposed is a strategy where what is stored in `ExternalPtr` isn't an owning pointer `T*`, but rather something ressembling `Box<dyn Any>`, which contains type information.
This type information is extracted through `downcast`. 

This is a draft, and I'd like your collaboration in making this better, as I'm struggling to figure out the type-stuff.

### DIFF
--- a/extendr-api/src/wrapper/externalptr.rs
+++ b/extendr-api/src/wrapper/externalptr.rs
@@ -135,7 +135,8 @@ impl<T: Any + Debug> ExternalPtr<T> {
     pub fn addr(&self) -> &T {
         unsafe {
             let ptr = R_ExternalPtrAddr(self.robj.get()) as *const Box<dyn Any>;
-            ptr.as_ref().unwrap().downcast_ref().unwrap()
+            let ptr_ref: &Box<dyn Any> = ptr.as_ref().unwrap();
+            ptr_ref.downcast_ref().unwrap()
         }
     }
 
@@ -144,7 +145,8 @@ impl<T: Any + Debug> ExternalPtr<T> {
     pub fn addr_mut(&mut self) -> &mut T {
         unsafe {
             let ptr = R_ExternalPtrAddr(self.robj.get()) as *mut Box<dyn Any>;
-            ptr.as_mut().unwrap().downcast_mut().unwrap()
+            let ref_mut_ptr: &mut Box<dyn Any> = ptr.as_mut().unwrap();
+            ref_mut_ptr.downcast_mut().unwrap()
         }
     }
 }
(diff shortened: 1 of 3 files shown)


extendr-api/src/wrapper/externalptr.rs
```diff
@@ -82,24 +82,25 @@ impl<T: Any + Debug> ExternalPtr<T> {
...
     pub fn new(val: T) -> Self {
+        use std::ffi::c_void;
         unsafe {
             // This allocates some memory for our object and moves the object into it.
-            let boxed = Box::new(val);
+            let v: Box<dyn Any> = Box::new(val);
 
             // This constructs an external pointer to our boxed data.
             // into_raw() converts the box to a malloced pointer.
-            let robj = Robj::make_external_ptr(Box::into_raw(boxed), r!(()));
```
Ilia-Kosenkov, line 91: Does `Robj::make_external_ptr()` is used anywhere else? can we resort to calling `extendr_api` methods rather than digging into `libR-sys`?

Ilia-Kosenkov: Also, here I see double indirection, which I am not sure is necessary or efficient.

CGMossa: When I submitted this PR, the tests passed locally. Now I'm changing it because something very odd is happening.

I don't want to use `Robj::make_external_ptr`, because it wants to protect `tag` and `prot`, but
those are protected by being part of an `ExternalPtr`.

The indirection is what gives me the ability to pick up the type information from the pointer.
As (apparently) the typeid is stored in a `Box<dyn Any>` rather than a `*mut T`. 
I don't know where this is written :(

CGMossa:
Okay; The failures on CI are surprising: I can see that `extendr-macros/src/extendr_impl.rs` does its own thing; So I would need to change the type-checking there as well.

But the question is now, why are other checks *passing*? They should all be broken on `extendrtests`.

**PR IS STILL IN DRAFT MODE I GUESS**

extendr-api/src/wrapper/externalptr.rs
```diff
@@ -131,39 +135,49 @@ impl<T: Any + Debug> ExternalPtr<T> {
...
+        single_threaded(|| unsafe {
+            let ptr = R_ExternalPtrAddr(self.robj.get()) as *mut ExternalData;
+            let ptr_ref = &mut *ptr;
+            let ptr_ref_downcast = ptr_ref.0.downcast_mut::<T>().unwrap();
+            ptr_ref_downcast
+        })
     }
 }
 
 impl<T: Any + Debug> TryFrom<&Robj> for ExternalPtr<T> {
```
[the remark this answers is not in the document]
CGMossa: Right now, we have `single_threaded(|| {})` everywhere when we access the R-API. And I don't know if that is good enough, I do see spurious issues;

But thanks for the note! Any comment, anywhere, is very appreciated.

CGMossa:
Too many things are broken here; Especially the `downcast_ref`/`downcast_mut` requiring `'static`. I'll try another approach suggested by @Wolvereness.

CGMossa:
I thought I had closed this. I abandoned this strategy because it is a little cluttered.

Please read the new draft PR.

PR #609, opened 2023-08-31
Integrate `ExternalPtr<T>` and `#[extendr]` for `impl`-blocks

CGMossa:
# [Help wanted]

Before this PR, `ExternalPtr<T>` and `#[extendr]`-impl were not connected at all.
In fact, it was doubly implemented. **Also**, runtime type-checks between passing R the
`ExternalPtr<T>`s was through strings. R recommends using symbols for the comparisons,
and while we did do that, the symbols are `ownership`-checked, and we were generating
them every time types were passed.

## Progress

* Added `ExternalData<T>` which adds a `TypeId` to what's covered by `ExternalPtr<T>`.
Thus, we have `ExternalPtr<ExternalData<T>>`. By virtue of `repr(C)`, we may extract first
field (here `TypeId`) from an opaque pointer to an unknown `ExternalPtr`. 

We use `addr_of!` because it could be bad to create an intermediate reference (for the pointer) of
a thing that isn't all that valid (`ExternalData<()>`). It is some dark-magic that someone suggested (named @wolvereness).

* Deprecating / Removing `check_external_ptr_type` as it is not a recommended way of doing
reflection, see documentation on `std::any::type_name` for this.

* From now on `#[extendr]`-impl types must be `Debug`. 

* Removed `From<&T> for Robj` as it doesn't make sense to create a `ExternalPtr` from a
reference; We need to have an owning pointer, to even finalize the object to begin with.

* Added static methods on `ExternalPtr`, as the issue is we cannot return references to
local data, and thus instantiating an `ExternalPtr<T>` in say `FromRobj` is not feasible.

## Hurdle

`FromRobj` is not playing nicely; Changing that will increase the foot-print of this PR,
and it is not obvious to me how to change it.

### DIFF
--- a/extendr-api/src/wrapper/externalptr.rs
+++ b/extendr-api/src/wrapper/externalptr.rs
@@ -1,6 +1,7 @@
 use super::*;
-use std::any::Any;
+use std::any::{Any, TypeId};
 use std::fmt::Debug;
+use std::marker::PhantomData;
 
 /// Wrapper for creating R objects containing any Rust object.
 ///
@@ -21,7 +22,7 @@ pub struct ExternalPtr<T: Debug + 'static> {
     pub(crate) robj: Robj,
 
     /// This is a zero-length object that holds the type of the object.
-    marker: std::marker::PhantomData<T>,
+    marker: PhantomData<T>,
 }
 
 impl<T: Debug + 'static> robj::GetSexp for ExternalPtr<T> {
@@ -74,6 +75,27 @@ impl<T: Debug + 'static> DerefMut for ExternalPtr<T> {
     }
 }
 
+/// Internal type to represent a Rust type tagged it's [`TypeId`].
+///
+/// This means that [`R_MakeExternalPtr`] returns a tagged pointer, that
+/// we can use to test the validity of the type-casting at runtime.
+///
+#[derive(Debug)]
+#[repr(C)]
+struct ExternalData<T> {
+    type_id: TypeId,
+    data: T,
+}
+
+impl<T: 'static> ExternalData<T> {
+    fn new(data: T) -> Self {
+        Self {
+            type_id: TypeId::of::<T>(),
+            data,
+        }
+    }
+}
+
 impl<T: Any + Debug> ExternalPtr<T> {
     /// Construct an external pointer object from any type T.
     /// In this case, the R object owns the data and will drop the Rust object
@@ -82,20 +104,19 @@ impl<T: Any + Debug> ExternalPtr<T> {
     /// An ExternalPtr behaves like a Box except that the information is
     /// tracked by a R object.
     pub fn new(val: T) -> Self {
+        use std::ffi::c_void;
+        // This allocates some memory for our object and moves the object into it.
+        let boxed = Box::new(ExternalData::new(val));
         unsafe {
-            // This allocates some memory for our object and moves the object into it.
-            let boxed = Box::new(val);
-
             // This constructs an external pointer to our boxed data.
             // into_raw() converts the box to a malloced pointer.
-            let robj = Robj::make_external_ptr(Box::into_raw(boxed), r!(()));
-
+            // let robj = Robj::make_external_ptr(Box::into_raw(boxed), r!(()));
+            let robj = Robj::from_sexp(single_threaded(|| {
+                R_MakeExternalPtr(Box::into_raw(boxed) as *mut c_void, R_NilValue, R_NilValue)
+            }));
             extern "C" fn finalizer<T>(x: SEXP) {
                 unsafe {
-                    let ptr = R_ExternalPtrAddr(x) as *mut T;
-
-                    // Free the `tag`, which is the type-name
-                    R_SetExternalPtrTag(x, R_NilValue);
+                    let ptr = R_ExternalPtrAddr(x) as *mut ExternalData<T>;
 
                     // Convert the pointer to a box and drop it implictly.
                     // This frees up the memory we have used and calls the "T::drop" method if there is one.
@@ -133,8 +154,8 @@ impl<T: Any + Debug> ExternalPtr<T> {
     /// Normally, we will use Deref to do this.
     pub fn addr<'a>(&self) -> &'a T {
         unsafe {
-            let ptr = R_ExternalPtrAddr(self.robj.get()) as *const T;
-            &*ptr as &'a T
+            let ptr = R_ExternalPtrAddr(self.robj.get()) as *const ExternalData<T>;
+            &(*ptr).data
         }
     }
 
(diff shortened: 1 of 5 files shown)


CGMossa:
On the rust discord, [message link](https://discordapp.com/channels/442252698964721669/443150878111694848/1146857190909558976), someone suggested
```rs
pub struct Robj;

pub trait FromRobj {
    fn from_robj(robj: &Robj) -> std::result::Result<&Self, &'static str>;
}

struct ExternalPtr<T>([T; 0]);

impl<T: FromRobj> ExternalPtr<T> {
    pub fn try_get_from_robj(robj: &Robj) -> std::result::Result<&T, String> {
        T::from_robj(robj).map_err(From::from)
    }
}
```

as a way where what I presented could compile.

extendr-api/src/wrapper/externalptr.rs
```diff
@@ -82,20 +104,19 @@ impl<T: Any + Debug> ExternalPtr<T> {
...
-            // This allocates some memory for our object and moves the object into it.
-            let boxed = Box::new(val);
-
             // This constructs an external pointer to our boxed data.
             // into_raw() converts the box to a malloced pointer.
-            let robj = Robj::make_external_ptr(Box::into_raw(boxed), r!(()));
-
+            // let robj = Robj::make_external_ptr(Box::into_raw(boxed), r!(()));
+            let robj = Robj::from_sexp(single_threaded(|| {
+                R_MakeExternalPtr(Box::into_raw(boxed) as *mut c_void, R_NilValue, R_NilValue)
```
Ilia-Kosenkov, line 115: `Robj::make_external_ptr()` sets class name, basically second argument in this call, to type name. If you do not do that, I believe you will lose the type annotation visible in R (since from R all external ptrs will be of identical 'type').

Wolvereness: The type name in Rust is not considered a "stable" API. It is an arbitrary representation of a type, only intended for debug purposes, and not considered to be consistent across Rust versions or even compilation configurations.

CGMossa: @iliak the idea behind this PR is to stop using tag for type checking.

Ilia-Kosenkov: For type checking yes, but it is used not only for that.

CGMossa: What else is it used for?

Ilia-Kosenkov: Ok I take it back, it seems `tag` is not directly exposed to user in R session.

Ilia-Kosenkov: Apart from something like this 
image

CGMossa: I think that's valuable enough. But this has to happen once, and it is protected by
the `ExternalPtr`. I'll do it in a follow-up PR.

CGMossa: Done.

extendr-macros/src/extendr_impl.rs
```diff
@@ -139,38 +139,7 @@ pub fn extendr_impl(mut item_impl: ItemImpl) -> TokenStream {
...
-            fn from(value: &'a #self_ty) -> Self {
-                unsafe {
-                    let ptr = Box::into_raw(Box::new(value));
-                    let res = Robj::make_external_ptr(ptr, Robj::from(()));
-                    res.set_attrib(class_symbol(), #self_ty_name).unwrap();
-                    res.register_c_finalizer(Some(#finalizer_name));
-                    res
-                }
-            }
-        }
```
CGMossa, line 163: Unsound

yutannihilation:
Naive question: should the type be checked on Rust's side at all? Since extendr's R wrapper functions related to an `extendr`-class are bundled in an environment, there's no chance that an unexpected type of external pointer is passed to extendr's side as long as the user uses the R package as intended. Actually, it seems it's a common practice to use `R_NilValue` for tag.

cpp11: https://github.com/r-lib/cpp11/blob/500f642b4ea132ec8c168fc70a28e81e9510ece3/inst/include/cpp11/external_pointer.hpp#L59C20-L59C37
arrow: https://github.com/apache/arrow/blob/44811ba18477560711d512939535c8389dd7787b/r/src/arrow_cpp11.h#L97
xml2: https://github.com/r-lib/xml2/blob/ef2310ba246aab7842100b271a610ebfa353ee16/inst/include/xml2_types.h#L21

CGMossa:
* Tag doesn't do anything. In the first iteration, I removed it. @Ilia-Kosenkov pointed out that
`.Internal(inspect(x))` would display the `tag` if it could, so that would yield better debugging experience.
* Using `type_name` for type-checking is not intended. In fact, `type_name` documentation states that this should be avoided.
* I'd like to support many things, especially the ability to use traits, and thus proper type checking will
be valuable with that. (I have an idea :P) 
* Casting and dereferencing a pointer is dangerous. If we are to gracefully panic, we need to check
for these things, instead of causing a segfault.

yutannihilation:
> * Tag doesn't do anything. In the first iteration, I removed it.

Oh, sorry, I didn't follow the discussion. I see.

> * Casting and dereferencing a pointer is dangerous.

I understand your fear. But, please look at the actual R wrappers. The external pointer itself is stored in the hidden variable `self` and the Rust function is always called against it. In order to abuse this function, the user need to corrupt the environment or define the function to `.Call()` directly by themselves, which I don't feel need to be cared.

https://github.com/extendr/extendr/blob/593be3da4a627c8efe8c2eccf99440a90fa00714/tests/extendrtests/R/extendr-wrappers.R#L138-L150

That said, since I don't understand the whole picture here, I'm not sure what I'm saying makes any sense. Let me think for a while. Thanks for answering.

CGMossa:
> > * Tag doesn't do anything. In the first iteration, I removed it.
> 
> Oh, sorry, I didn't follow the discussion. I see.
> 
> > * Casting and dereferencing a pointer is dangerous.
> 
> I understand your fear. But, please look at the actual R wrappers. The external pointer itself is stored in the hidden variable `self` and the Rust function is always called against it. In order to abuse this function, the user need to corrupt the environment or define the function to `.Call()` directly by themselves, which I don't feel need to be cared.
> 
> https://github.com/extendr/extendr/blob/593be3da4a627c8efe8c2eccf99440a90fa00714/tests/extendrtests/R/extendr-wrappers.R#L138-L150
> 
> That said, since I don't understand the whole picture here, I'm not sure what I'm saying makes any sense. Let me think for a while. Thanks for answering.

I now understand what this means. I've also figured out some stuff here, so I will close this PR and open a new one.

PR #610, opened 2023-08-31
`Rarray` rebuild

daniellga:
I have been using `RArray` for some time so I made changes that I think will make its use easier for an inexperienced user.
So this PR is an `RArray` rebuild, making it easier to create an `RArray` and also removing `RColumn` and `RMatrix3D`, which I don't think make that much sense.
`collect_rarray` now also accepts dynamic arguments like `Vec`.

### DIFF
--- a/extendr-api/src/wrapper/array.rs
+++ b/extendr-api/src/wrapper/array.rs
@@ -38,7 +38,7 @@ pub struct RArray<T, D> {
 
 pub type RColumn<T> = RArray<T, [usize; 1]>;
 pub type RMatrix<T> = RArray<T, [usize; 2]>;
-pub type RMatrix3D<T> = RArray<T, [usize; 3]>;
+pub type RArray3D<T> = RArray<T, [usize; 3]>;
 
 const BASE: usize = 0;
 
@@ -74,24 +74,25 @@ impl<T> Offset<[usize; 3]> for RArray<T, [usize; 3]> {
     /// Get the offset into the array for a given index.
     fn offset(&self, index: [usize; 3]) -> usize {
         if index[0] - BASE > self.dim[0] {
-            panic!("RMatrix3D index: row overflow");
+            panic!("RArray3D index: row overflow");
         }
         if index[1] - BASE > self.dim[1] {
-            panic!("RMatrix3D index: column overflow");
+            panic!("RArray3D index: column overflow");
         }
         if index[2] - BASE > self.dim[2] {
-            panic!("RMatrix3D index: submatrix overflow");
+            panic!("RArray3D index: submatrix overflow");
         }
         (index[0] - BASE) + self.dim[0] * (index[1] - BASE + self.dim[1] * (index[2] - BASE))
     }
 }
 
 impl<T, D> RArray<T, D> {
+    /// Make a new RArray type.
     pub fn from_parts(robj: Robj, data: *mut T, dim: D) -> Self {
         Self { robj, data, dim }
     }
 
-    /// Get the underlying data fro this array.
+    /// Get the underlying data from this RArray.
     pub fn data(&self) -> &[T] {
         unsafe { std::slice::from_raw_parts(self.data, self.robj.len()) }
     }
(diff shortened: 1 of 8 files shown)


CGMossa:
Yes! I support this fully. I also hope it has multiple dimension support.
I'll definitely review this thoroughly.

CGMossa:
Can you request reviews from me / us when you're ready?

Ilia-Kosenkov:
Maybe there is some issue with access that you cannot ask for reviews?
Anyway, looks impressive, I need time to go through these things.

What about unit tests? Did you manage to make it work with old ones or do we have no tests for `RMatrix`? If it is the latter, could you add some tests, especially for creation/type conversion?

extendr-api/src/wrapper/array.rs
```diff
@@ -0,0 +1,339 @@
...
+        D: AsRef<[usize]> + Clone,
+    {
+        let dims_ref = dims.as_ref();
+        let slice_ref = slice.as_ref();
+        let robj = slice_ref.iter().cloned().collect_robj();
+
+        let prod = dims_ref.iter().product::<usize>();
+        if prod != robj.len() {
+            return Err(Error::Other(format!(
+                "The vector length ({}) does not match the length implied by the dimensions ({})",
```
Ilia-Kosenkov, line 75: I saw this error twice in the codebase, I believe it deserves its own error type with perhaps two parameters and this string moved to `impl Debug`

daniellga: I was trying to implement a generic `ValuesMismatch` error where it accepts 2 values that implement the debug trait but the partialeq derive for Error blocked it... I think it would be better to create an issue in order to implement it another time...
I implemented a `DimensionMismatch` error anyway, but I ended up not too happy about it since it is too specialized...

extendr-api/src/wrapper/array.rs
```diff
@@ -0,0 +1,339 @@
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn from_slice() {
+        test! {
+            // Create 2D RArray.
+            let slice = [1,2,3,4,5,6];
+            let dim = [2,3];
+            let array = RArray::from_slice(slice, dim).unwrap();
```
JosiahParry, line 280: Are there other construction methods besides `from_slice()`?

daniellga: There's also `from_slice_unchecked`. Since creating a new `RArray` creates a copy of the inner data, these were the only 2 methods I thought were needed. Note that you can use `from_slice()` directly on a `Vec` or on a rust array in this case.
I made `RArray::new()` available only to be used internally, since it doesn't add the `dim` attribute for the Robj.
Let me know if you are thinking of another type of constructor.

JosiahParry: Is there a way to create one without copying data? That would be quite nice. Additionally, it would be handy to be able to determine the col/row major order or at least provide a closure similar to the matrix method.

daniellga: I don't think it is possible to create an `Robj` without copying data.

Is it ok for you if I add an example on how to build row major matrices? Or is the `RMatrix` constructor using a closure really desirable? I removed it because think the data manipulation should be left for the user to figure out and also because it was a somewhat complicated function to understand how to use... 

Maybe @CGMossa and @Ilia-Kosenkov  can give their opinion on this.
```
    #[test]
    fn test_collect_rarray_matrix_row_major() {
        test! {
            // Test building an RMatrix in row major order.
            let v: Vec<i32> = (1i32..=12).collect();
            let nrow = 4_usize;
            let ncol = 3_usize;
            let rmat = (0..ncol)
                .flat_map(|i| {
                    v.iter()
                        .skip(i)
                        .step_by(ncol)
                        .cloned()
                })
                .collect_rarray([nrow, ncol]);
            assert!(rmat.is_ok());
            assert_eq!(Robj::from(rmat), R!("matrix(1:12, nrow=4, byrow = TRUE)").unwrap());
        }
    }
```

CGMossa:
Hello @daniellga. I am honestly very sorry that we haven't resolved this. I want to solve it within the next two weeks. If you have availability to get it over the finish line, please tell me.

extendr-api/src/wrapper/array.rs
JosiahParry: I would say that the closure is very important. I can find 8 examples where I'm using it in existing code. 
https://github.com/search?q=owner%3AJosiahParry%20new_matrix&type=code

It's very helpful in that I often iterate over rows and store a `Vec<T>` where `T` is some structure that is on a per-row basis. Presently I am writing code that creates a matrix from `Vec<[f64;2]>` with the closure `|r, c| x[r][c]`.

daniellga: The closure is definitely useful, but I find it very confusing. That's why I removed it in the first place. The changes I made are very opinionated as I think these transformations made by the closure should be done outside extendr. When creating the array, I think the user should already have the Vec (or slice) already ready to be converted to an RArray.

JosiahParry: It requires a whole lot of mental gymnastics to understand how to create a flattened vector in the appropriate shape for R's column-major matrices. Not many people will be able to figure that out--myself included. QOL >>> imo

daniellga:
@CGMossa and @JosiahParry I don't have time at all at the moment... Family stuff. Feel free to modify it if you think changes are worth. I think you are allowed to do it, right? (not very used to github)

JosiahParry:
I still would like to improve this experience here. I continually find myself needing to make matrixes _by row_. Here is a current use case and makes it _harder_ to achieve by removal of the closure approach which, I must admit is clunky, but better than nothing. 

Given a  vector of a struct with a `const usize` parameter, `Vec<EsriCoord<N>>`, I need to make  a matrix with dimension `(x.len(), N)`. Constructing a matrix right now is so awkward. This PR gets some of the way there but intentional excludes this use case. 

```rust
pub struct EsriCoord<const N: usize>(pub [f64; N]);
```

PR #615, opened 2023-09-04
Snapshot-test extendr macros expansions

Ilia-Kosenkov:
- Test generated code
- Visualize expansion for reference

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -61,7 +61,7 @@ jobs:
         shell: pwsh
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Set up Rust
         uses: dtolnay/rust-toolchain@master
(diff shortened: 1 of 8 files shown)


Ilia-Kosenkov:
Blocked by https://github.com/dtolnay/install/issues/12. If it will not be resolved by the author/us, I will switch to just `cargo install` cmd.

Ilia-Kosenkov:
Dayum, snapshot tests are disabled by default on CRAN so they were not executed here >_<. I was debugging unreachable code on CI

CGMossa (approved):
Honestly, I don't know this stuff anyways, but nothing came up when looking through it...

I find the `nzchar` function odd. It just returns `TRUE/FALSE`, but that's not what the `?nzchar` says about it...

LGTM

Ilia-Kosenkov:
Wait, what's wrong with `nzchar`? I did not get it

CGMossa:
>nchar takes a character vector as an argument and returns a vector whose elements contain the sizes of the corresponding elements of x.
but it returns boolean/logical.

Ilia-Kosenkov:
Ah, but there are two functions. `nchar` counts characters, and `nzchar` (I read it as non-zero characters) just tells us if string is not empty.

PR #618, opened 2023-09-09
Use once_cell instead of lazy_static

### LINKED ISSUE #579: Replace lazy_static with once_cell
The lazy_static crate is going to be deprecated, so it's better to migrate to the once_cell crate (and maybe then to `std::sync::OnceLock` or `std::sync::LazyLock`).

https://github.com/rust-lang-nursery/lazy-static.rs/issues/214

As of Rust 1.70, `std::cell:OnceCell` and `std::sync::OnceLock` (thread-safe version) are stabilized. The lazy version of these two, `std::cell::LazyCell` and `std::sync::LazyLock`, are still unstable at the time of writing this.

yutannihilation:
Close #579 

As I described on #579, it seems the lazy_static crate is going to be deprecated (https://github.com/rust-lang-nursery/lazy-static.rs/issues/214) and the implementation based on once_cell is getting stabilized in std. It's not that the lazy-static will become unusable soon, so this isn't very urgent.

### DIFF
--- a/extendr-api/Cargo.toml
+++ b/extendr-api/Cargo.toml
@@ -18,9 +18,9 @@ rust-version = "1.60"
 
 [dependencies]
 libR-sys = { workspace = true }
-extendr-macros = { path = "../extendr-macros", version="0.4.0" }
-extendr-engine = { path = "../extendr-engine", version="0.4.0" }
-lazy_static = "1.4"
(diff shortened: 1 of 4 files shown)


CGMossa (approved):
LGTM

CGMossa:
@yutannihilation should I merge this?

yutannihilation:
Sorry for not merging. Thanks for reminding.

PR #619, opened 2023-09-09
[DO NOT MERGE THIS] Altrep test failures

yutannihilation:
This branch is a tweaked version of #600 to show the test failures discussed in https://github.com/extendr/extendr/pull/600#issuecomment-1706978167.

I could reproduce the error on my local, and the error message says it happens here:

https://github.com/extendr/extendr/blob/1c32f6904274f55e109e132bf1ef0e2e4b11647c/extendr-api/src/wrapper/altrep.rs#L533

To investigate what's happening, I expanded the line as below in the assumption that the problem is that extendr tries to extract the pointer address from a wrong data. As expected, I saw this error message on my local, so it's probably means my assumption is right.

``` rs
            let sexp = R_altrep_data1(x);
            if TYPEOF(sexp) as u32 != libR_sys::EXTPTRSXP {
                panic!("This isn't an external pointer! (type {})", TYPEOF(sexp));
            };
            let state_ptr = R_ExternalPtrAddr(sexp);
```

So, what's the problem? ~~My guess is that the `R_altrep_data1()` above is a mistake~~ (edit: sorry, this itself is correct, but this should not be used in finalizer). Since this finalizer is defined for the external pointer inside the altrep, not for the altrep itself, `R_ExternalPtrAddr()` should be applied to `x` instead of `R_altrep_data1(x)`.

https://github.com/extendr/extendr/blob/1c32f6904274f55e109e132bf1ef0e2e4b11647c/extendr-api/src/wrapper/altrep.rs#L492-L496

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+- `ALTLIST` support allowing users to represent structs as R list objects
 - [**either**] `TryFrom<&Robj> for Either<T, R>` and `From<Either<T, R>> for Robj` if `T` and `R` are themselves implement these traits. This unblocks scenarios like accepting any numeric vector from R via `Either<Integers, Doubles>` without extra memory allocation [[#480]](https://github.com/extendr/extendr/pull/480)
 - `PartialOrd` trait implementation for `Rfloat`, `Rint` and `Rbool`. `Rfloat` and `Rint` gained `min()` and `max()` methods [[#573]](https://github.com/extendr/extendr/pull/573)
 - `use_rng` option for the `extendr` attribute macro, which enables the use of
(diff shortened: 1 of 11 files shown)


yutannihilation:
Confirmed it's reproducible on the CI as well.

https://github.com/extendr/extendr/actions/runs/6128634718/job/16635888669#step:16:178

yutannihilation:
This pull request isn't intended as the fix to be merged, but I'd like to check if my understanding is correct. Sorry for the noise.

yutannihilation:
Confirmed (The R-oldrel failures are expected as the current code use AltList unconditionally. The R-devel failure seems a temporary one).

CGMossa:
Very good find.. I've been going through this stuff lately, and especially this, and didn't catch this.. Will check again myself. I assume you'll put a PR up;

PR #620, opened 2023-09-09
Fix ALTREP test failures

yutannihilation:
## Problem

This pull request aims to fix the integration test failure which was found during #600. The error message says

> `thread '<unnamed>' panicked at 'misaligned pointer dereference: address must be a multiple of 0x8 but is 0x310073656e696c', D:\a\extendr\extendr\extendr-api\src\wrapper\altrep.rs:544:13`

This indicates the extendr extracts the pointer address from what should not be treated as a pointer. The line the message indicates is here:

https://github.com/extendr/extendr/blob/581329f49e3742f3dc4f7b3e5c5e0b87ced0e40a/extendr-api/src/wrapper/altrep.rs#L543-L544

## Cause and fix

ALTREP is a pairlist of two SEXP objects.

``` c
SEXP R_new_altrep(R_altrep_class_t aclass, SEXP data1, SEXP data2)
{
    SEXP sclass = R_SEXP(aclass);
    int type = ALTREP_CLASS_BASE_TYPE(sclass);
    SEXP ans = CONS(data1, data2);
    SET_TYPEOF(ans, type);
    SET_ALTREP_CLASS(ans, sclass);
    return ans;
}
```
https://github.com/wch/r-source/blob/08a6e5c3f1c10ba269a02a391d33454404119ed2/src/main/altrep.c#L1150-L1158

In extendr's case, `data1` is an external pointer (`EXTPTRSXP`), and `data2` is `NULL`.

https://github.com/extendr/extendr/blob/581329f49e3742f3dc4f7b3e5c5e0b87ced0e40a/extendr-api/src/wrapper/altrep.rs#L508-L515

`EXTPTRSXP` has a destructor, and extendr's implementation uses ` Altrep::get_state_mut()` to acquire the pointer address. But, since it's applied to the external pointer, not the ALTREP itself, it should use `R_ExternalPtrAddr()` directly. This is the main thing of this pull request.

## Tweak related to ALTLIST

One tricky thing is that list ALTREP, or ALTLIST, is not supported on R < 4.3. So, #600 needed to use a feature gate and enable it depending on the version in `build.rs`. This needs the same tweak. More precisely,

* add `libR-sys` as the dependency because the `build.rs` needs the envvars of R versions (`DEP_R_R_VERSION_...`)
* skip alt-list integration tests on R < 4.3.
* skip macro snapshot tests as the implementation is actually diverged on R < 4.3


CGMossa:
Idea: Maybe `extendr-api` should emit a compile-time `cfg` to dependent crates on whether or not ALTLIST is available. This is done in `extendr-api/build.rs` by using  [cargo:KEY=VALUE](https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key), and emitting `use_r_altlist`, which then for downstream users `DEP_EXTENDR_USE_R_ALTLIST`. I am not sure it is possible to rename this.
Of course you know this already, as you taught us this already..

yutannihilation:
Oh, thanks. Since it was long ago I investigated how to propagate the settings or versions, I didn't remember how to do it (or even it's possible). It really helps!

yutannihilation:
Hmm, sorry, I couldn't figure out how to use `DEP_` envvars in the case of non-sys crates. I think I'm hitting the same limit as this, but I'm not sure. I'm giving up this time...

https://github.com/extendr/libR-sys/pull/159#discussion_r1167616378

yutannihilation:
At this point, it is confirmed that

* all compilation succeed, and
* all integration tests fail

as expected. Next, I'm pushing the fix.

yutannihilation:
I updated the description and marked as ready for review.

Ilia-Kosenkov:
Can you merge `master`? I suspect that since you touched `{extendrtests}`, the snapshot test may fail after merge.

yutannihilation:
Yeah, it failed and I addressed.

tests/extendrtests/src/rust/Cargo.toml
```diff
@@ -16,6 +16,10 @@ crate-type = ["staticlib"]
 [dependencies]
 extendr-api = { version = "*", features = ["graphics", "ndarray", "either"] }
```
Ilia-Kosenkov, line 17: I am not sure, but can we just enable `use_r_altlist` here? It is a test project, I doubt we need to ensure that we have R 4.3. Then you could simplify a lot of code that enables this feature.

CGMossa: These things don't mix. It is not possible to enable compile-time cfg within cargo features.
Cargo features are cfg, with `cfg(feature = some_cargo_feature)`.

Ilia-Kosenkov: Thanks for explaining, did not know that. Then LGTM

yutannihilation: I too was wondering if it's really not possible to do this in some simple way, but I couldn't come up with any nice solution. Even if extendr expose the feature gate, AltList-related APIs doesn't exist in the bindings of libR-sys, so extendr would need to provide dummy implementation for R < 4.3, which doesn't sound very useful either.

I think this problem (switching implementation based on the R version) is a general one and I need to find some solution at some point. But, I gave up in this pull request.

CGMossa (approved):

CGMossa (approved):
LGTM

yutannihilation:
Thanks, merging.

PR #621, opened 2023-09-10
Use R_RegisterCFinalizerEx() instead of R_RegisterCFinalizer()

yutannihilation:
(This isn't a big problem. Just for housekeeping.)

During reading the source code of cpp11, I found they use `R_RegisterCFinalizerEx()` instead of `R_RegisterCFinalizer()`. The difference is the additional `onexit` argument, which is explaind in [WRE](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#External-pointers-and-weak-references) as below:

> R does not perform a garbage collection when shutting down, and the `onexit` argument of the extended forms can be used to ask that the finalizer be run during a normal shutdown of the R session. It is suggested that it is good practice to clear the pointer on finalization.

Actually, `R_RegisterCFinalizer()` is equivalent to `R_RegisterCFinalizerEx(..., FALSE)`.

https://github.com/wch/r-source/blob/6691d53058214c16e5ab8fd6eb41c06e7fb3227f/src/main/memory.c#L1619

I'm not 100% sure if there's no case that this should be `FALSE`, but in extendr's case, it seems there's no interface that users can supply their own finalizers and the current two are what should also be called on the shutdown. So, I think it's fine to pass `TRUE` fixedly.

https://github.com/extendr/extendr/blob/593be3da4a627c8efe8c2eccf99440a90fa00714/extendr-api/src/wrapper/altrep.rs#L502-L506
https://github.com/extendr/extendr/blob/593be3da4a627c8efe8c2eccf99440a90fa00714/extendr-api/src/wrapper/externalptr.rs#L93-L107

### DIFF
--- a/extendr-api/src/robj/rinternals.rs
+++ b/extendr-api/src/robj/rinternals.rs
@@ -272,7 +272,9 @@ pub trait Rinternals: Types + Conversions {
 
     #[doc(hidden)]
     unsafe fn register_c_finalizer(&self, func: R_CFinalizer_t) {
-        single_threaded(|| R_RegisterCFinalizer(self.get(), func));
+        // Use R_RegisterCFinalizerEx() and set onexit to 1 (TRUE) to invoke the
+        // finalizer on a shutdown of the R session as well.
+        single_threaded(|| R_RegisterCFinalizerEx(self.get(), func, 1));
     }
 
     /// Copy a vector and resize it.
--- a/extendr-api/src/wrapper/altrep.rs
+++ b/extendr-api/src/wrapper/altrep.rs
@@ -509,7 +509,10 @@ impl Altrep {
             let tag = R_NilValue;
             let prot = R_NilValue;
             let state = R_MakeExternalPtr(ptr as *mut c_void, tag, prot);
-            R_RegisterCFinalizer(state, Some(finalizer::<StateType>));
+
+            // Use R_RegisterCFinalizerEx() and set onexit to 1 (TRUE) to invoke
+            // the finalizer on a shutdown of the R session as well.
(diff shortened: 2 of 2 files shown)


CGMossa:
I have recently gone through this stuff. I thought `onexit` was about the function scope, not r-session itself.
This is a giant bug honestly, that we don't clean-up on shutdown of r-session.

I had previously looked at this:
```c

/* use fatal !=0 for emergency bail out */
void Rf_endEmbeddedR(int fatal)
{
    R_RunExitFinalizers();
    CleanEd();
    R_CleanTempDir();
    if(!fatal){
	Rf_KillAllDevices();
	AllDevicesKilled = TRUE;
    }
    if(!fatal && R_CollectWarnings)
	PrintWarnings();	/* from device close and .Last */
    app_cleanup();
}
```
and thought that `R_RunExitFinalizers` was ran when R shutdown.
I think we should merge this ASAP.. Let me read the code.

CGMossa (approved):
LGTM

I would say that you also need to look in `extendr_impl.rs` as there is manually / macro
coded this stuff, and if you could patch that, that would be great. 

Approving if you don't want to do it here.

yutannihilation:
> I would say that you also need to look in `extendr_impl.rs` as there is manually / macro
> coded this stuff, and if you could patch that, that would be great.

Thanks. For some reason, I oversaw this. But, as it also sets this via `register_c_finalizer()`, I think this pull request already covers.

https://github.com/extendr/extendr/blob/593be3da4a627c8efe8c2eccf99440a90fa00714/extendr-macros/src/extendr_impl.rs#L165-L175

I'm merging for now.

PR #623, opened 2023-09-11
Patch current `extendr`-impl macro

CGMossa:


CGMossa:
@Ilia-Kosenkov do you know why this is failing?
```
Show testthat output
  find: ‘check’: No such file or directory
```

Ilia-Kosenkov:
I will check.

Ilia-Kosenkov:
Oh you dropped `From` and now itis confused. Even if there is `TryFrom`, `impl`s might not utilize it yet (I do not know).

CGMossa:
Fundamentally, `From<&T> for Robj`, shouldn't be implemented in context of `ExternalPtr`. We are supposed to get the reference from the provided `Robj`, by accessing the stored pointer.
Even a `TryFrom<&T> for Robj` would be wrong as well.

Ilia-Kosenkov:
But how else would you return a `Self` to R?

CGMossa:
There is `From<T> for Robj`. That is the ownership is taken, and the object is "leaked". Now, R owns it. R destroys/drops it. R yields back `Robj` from which we only may extract `&T` and `&mut T`, or any other pointer-type.

PR #625, opened 2023-09-17
Fix a typo

yutannihilation:
Just happened to find. Maybe this doesn't matter because the internal representation of `LGLSXP` and `INTSXP` are the same?


CGMossa (approved):
LGTM

Nice find and good that we clear these up.......
A lot of my PRs matter only sometimes anyways, so this is like as good as anything going on atm.

Issue #608, opened 2023-08-31
Introduce unprotected `Robj`


CGMossa:
I'd like to propose that we introduce unprotected `Robj`. These will be `SEXP`s
that will not be added to the protection list of `ownership`-module.

- Implementation of iterators had a memory constraint that was due to not being
able to protect and unprotect stuff, as R has a limit of how many standalone `PROTECT` calls,
(10k or maybe 22k).
- We use symbols everywhere, and symbols should not be protected. Nor should `r!(())`. 
- Useful when embedding types in `ExternalPtr`. We may want to embed robjects in `ExternalPtr`,
for intricate use-cases of ALTREP, (or even intricate uses of `ExternalPtr` even).

Why an issue? Isn't it trivial to introduce such a type?

Unfortunately the path to introducing this isn't clear for me yet. An `Robj` will have to
somehow contain this type, as we need the type-inference to work for this.

I'm mainly creating this issue to discuss

- Name: `UnsafeRobj`, `FreeRobj`, `UnprotectedRobj`. 
- Implementation strategy, concerns, things you know one has to ensure works..

CGMossa:
Amending with two more points:

* From #626, `SEXP` / `Robj` we receive from R are protected by the caller. Thus, those don't need to be protected.
* We generate `Robj` from `&Robj` and those "clones" need not be protected either.

CGMossa:
I am currently attempting a full rewrite using this new definition for `Robj`:

```rs
pub enum Robj {
    /// Protected by R, or a symbol
    Permanent(RawRobj),
    /// Unprotected / free R object that needs custom protection
    Raw(RawRobj),
    /// Owned (by Rust), meaning it has to be unleashed by Rust
    Owned(RawRobj),
}
```

I'm also using some Rust tricks to make this a seamless transition, by providing `Deref/DerefMut` implementations, and similar considerations.

This rewrite is uncovering a few things, hence the many PRs. If I succeed with this rewrite, the PR will be too big to review and approve. But it might come down to it -- I'm doing my best, but this project is a beast to untangle.

This `Robj` would act more like `Box`, rather than what was previously thought of.. which I would say more of an `Rc`. I do not think the `Rc` semantics fit with "protection"/"unprotect". :(

PR #626, opened 2023-09-18
[PoC] Skip protection for SEXPs passed from R session

yutannihilation:
[DO NOT MERGE THIS]

Currently, extendr protects all SEXPs, but I believe it can be skipped if the object is already protected by the caller. This pull request contains only a minimum subset of such cases because there are too many `Robj::from_sexp()` to review.

While this might look good, this brings a critical bug in cases like this:

``` rust
struct Foo {
    inner: Robj
};

#[extendr]
impl Foo {
    fn new(x: Robj) -> Self {
        Self { inner: x }
    }
}
```

In this case, apparently the input must be protected on Rust's side because it's possible that `x` gets GC-ed immediately after a new `Foo` instance is created.

It's valid to say that it's the package author's responsibility to protect and unprotect it like below, but it might feel a bit too low-level for extendr's ergonomics. Also, `protect()` and `unprotect()` are not exported (while [the document](https://extendr.github.io/extendr/extendr_api/ownership/index.html) indicates they are exported)

``` rust
struct Foo {
    inner: Robj
};

#[extendr]
impl Foo {
    fn new(x: Robj) -> Self {
        protect(x);
        Self { inner: x }
    }
}

impl Drop for Foo {
    fn drop(&mut self) {
        unprotect(self.inner);
    }
}
```

It seems the supposed way to protect an R object associated with a struct is to pass the `SEXP` to the `prot` arugment of `R_MakeExternalPtr()`. [WRE](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#External-pointers-and-weak-references) says:

> the `prot` field for protecting the memory that the external pointer represents, if that memory is allocated from the R heap

But, I doubt it can be done automatically on extendr's side. So, I have no idea.

---

For the details about the idea of distinguishing external and owned SEXPs, I wrote a blog post, if you are interested: https://yutani.rbind.io/post/intro-to-savvy-part1/

### DIFF
--- a/extendr-api/src/robj/mod.rs
+++ b/extendr-api/src/robj/mod.rs
@@ -110,11 +110,16 @@ pub use rinternals::Rinternals;
 ///
 pub struct Robj {
     inner: SEXP,
+    protect: bool,
 }
 
 impl Clone for Robj {
     fn clone(&self) -> Self {
-        unsafe { Robj::from_sexp(self.get()) }
+        if self.protect {
+            unsafe { Robj::from_sexp(self.get()) }
+        } else {
+            unsafe { Robj::from_protected_sexp(self.get()) }
+        }
     }
 }
 
@@ -216,8 +221,15 @@ impl Robj {
     pub fn from_sexp(sexp: SEXP) -> Self {
         single_threaded(|| {
             unsafe { ownership::protect(sexp) };
-            Robj { inner: sexp }
-        })
+        });
+        Self::from_protected_sexp(sexp)
+    }
+
+    pub fn from_protected_sexp(sexp: SEXP) -> Self {
+        Robj {
+            inner: sexp,
+            protect: true,
+        }
     }
 
     /// A ref of an robj can be constructed from a ref to a SEXP
@@ -1078,6 +1090,11 @@ pub(crate) unsafe fn to_str<'a>(ptr: *const u8) -> &'a str {
 /// Release any owned objects.
 impl Drop for Robj {
     fn drop(&mut self) {
+        // Do nothing if it's not protected on extendr's side.
+        if !self.protect {
+            return;
+        }
+
         unsafe {
             ownership::unprotect(self.inner);
         }
--- a/extendr-api/src/robj/mod.rs
+++ b/extendr-api/src/robj/mod.rs
@@ -222,13 +222,16 @@ impl Robj {
         single_threaded(|| {
             unsafe { ownership::protect(sexp) };
         });
-        Self::from_protected_sexp(sexp)
+        Robj {
+            inner: sexp,
+            protect: true,
+        }
     }
 
     pub fn from_protected_sexp(sexp: SEXP) -> Self {
         Robj {
             inner: sexp,
-            protect: true,
+            protect: false,
         }
     }
 
--- a/extendr-macros/src/extendr_impl.rs
+++ b/extendr-macros/src/extendr_impl.rs
@@ -165,7 +165,7 @@ pub fn extendr_impl(mut item_impl: ItemImpl) -> TokenStream {
         // Function to free memory for this type.
         extern "C" fn #finalizer_name (sexp: extendr_api::SEXP) {
             unsafe {
-                let robj = extendr_api::robj::Robj::from_sexp(sexp);
+                let robj = extendr_api::robj::Robj::from_protected_sexp(sexp);
                 if robj.check_external_ptr_type::<#self_ty>() {
                     //eprintln!("finalize {}", #self_ty_name);
                     let ptr = robj.external_ptr_addr::<#self_ty>();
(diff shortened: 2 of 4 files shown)


yutannihilation:
As expected, this test fails fatally.

https://github.com/extendr/extendr/blob/2992da0b3c71b7be2622ed9b5a4bc72869e3ccbc/extendr-api/tests/dataframe_tests.rs#L8-L12

CGMossa:
This is related in spirit to my #608.

Can you explain why you'd expect that latest test to fail?

I also think we could potentially make the external ptr thing work.. But we'll have to introduce some weirdness that I now am motivated to try out..

yutannihilation:
Hmm, I thought it's a problem I described above, but, apparently it's not the case when the R obj is passed from outside. Sorry.

It seems the error happens on these two lines, so, it's probably something related to `CHARSXP` protection. `into_dataframe()` calls R function `data.frame()` internally, and the result is probably not bound to any variable in the R session. But, I'm not sure about the details yet.

https://github.com/extendr/extendr/blob/2992da0b3c71b7be2622ed9b5a4bc72869e3ccbc/extendr-api/tests/dataframe_tests.rs#L21
https://github.com/extendr/extendr/blob/2992da0b3c71b7be2622ed9b5a4bc72869e3ccbc/extendr-api/tests/dataframe_tests.rs#L29

yutannihilation:
A quick update. I was wrong on this.

> It seems the error happens on these two lines, so, it's probably something related to `CHARSXP` protection.

The error happens regardless of the type. Even this fails. But, if I change the index of last line to `list[0]`, it's not. So, the error seems to happen if the index is larger than 0. It's so weird and I have no idea.

```rust
#[test]
fn test_derive_into_dataframe() {
    test! {
        use extendr_api::prelude::*;

        #[derive(Debug, IntoDataFrameRow)]
        struct MyStruct {
            x: Rint,
            y: Rint,
        }

        let v = vec![MyStruct { x: 0.into(), y: 0.into() }, MyStruct { x: 1.into(), y: 2.into() }];
        let df = v.into_dataframe()?;

        assert!(df.inherits("data.frame"));

        let list : List = df.as_list().unwrap();
        println!("{:?}", list[1].as_integers());
    }
}
```

CGMossa:
I've tried to investigate this. My latest guess is that `Robj::from` is broken due to this PR.
I tried to track down the issue, first, there are several tests that break with this, not only data-frame:
```
error: 4 targets failed:
    `-p extendr-api --test dataframe_tests`
    `-p extendr-api --test debug_tests`
    `-p extendr-api --test vector_tests`
    `-p extendr-api --doc`
```
Second, I tried to use `gdb` to track it, but I'm not getting anything useful...

CGMossa:
Like, I've boiled it down to:
```rs

        let s = List::from_values(["x", "y", "z"]);
        dbg!(&s[0]);
        dbg!(&s[1]);
        dbg!(&s[2]);

[extendr-api/tests/unprotected.rs:8] &s[0] = ["x"]
[extendr-api/tests/unprotected.rs:9] &s[1] = ["z"]
[extendr-api/tests/unprotected.rs:10] &s[2] = 
 *** caught segfault ***
address (nil), cause 'unknown'
```

So `[0]` is `x`, that makes sense, `[1]` is `z`!? And then `[2]` is messed up??!!?!?

CGMossa:
@yutannihilation I've found it! There are unclever uses of `as Robj`, see
```rs
    /// Get the list a slice of `Robj`s.
    pub fn as_slice(&self) -> &[Robj] {
        unsafe {
            let data = DATAPTR(self.robj.get()) as *const Robj;
            let len = self.robj.len();
            std::slice::from_raw_parts(data, len)
        }
    }
```
here the equivalence between `Robj` and `SEXP` is exploited. It is however wrong here, and this is what is causing the issue.

yutannihilation:
Wow, thanks so much! I couldn't come up with such a usage.

yutannihilation:
But, I have no idea how I can tweak this. As this pull request changes the data structure of `Robj`, `&[SEXP]` cannot be converted to `&[Robj]` without allocation, which is against [the naming convention of `as_`](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv).

CGMossa:
Alright, I've tried and asked around.. This is simply not possible to do. Even if you dare to make an allocation in there, you'd be returning a reference to a locally allocated thing, which isn't allowed. Then you could `Box`/heap allocate it (I guess?) but that doesn't work either.

This is one of the reasons why I suggested to make a `UnprotectedRobj` (as in Rust doesn't ensure its protection), as that would help with such issues.

I'd like this to move forward, but I don't see it happening by merely tagging `Robj` :(

yutannihilation:
Thanks for analyzing. Yeah, I think you are right. I too chose making a different type in my savvy framework, but, I was just hoping this would work for extendr because this would require less changes.

Let's give up for now.

PR #627, opened 2023-09-21
Fix non-Api calls issue in R packages

Ilia-Kosenkov:
` Found non-API calls to R` issue arises because `extendr-engine` uses non-API calls in its methods. `extendr-engine` was included as a dependency even though it is used only for testing (`with_r()` and `test!{}`). Moving `extendr-engine` to `dev-dependencies` solves the problem.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -221,13 +221,13 @@ jobs:
       # Rust pseudo multi-targeting: R CMD CHECK receives an extra argument '--no-multiarch' if BUILD_TARGETS is not 'default' and does not contain any commas
       # To enable RStudio support, extendrtests should be installable from 'extendr/tests/extendrtests',
       # for this work directory should be modified and then reverted back
-      - name: Run R integration tests using {extendrtests}
+      - name: Run R intergration tests using {extendrtests}
         uses: r-lib/actions/check-r-package@v2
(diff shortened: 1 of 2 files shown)


CGMossa (approved):
Thanks for this

yutannihilation:
Oh, thanks. I didn't notice this. One quick note about this point,

> ` Found non-API calls to R` issue arises because `extendr-engine` uses non-API calls in its methods.

the current CRAN version of string2path includes `extendr-engine` but has no such a problem. Probably this matters only when `test!` is actually used?

https://github.com/yutannihilation/string2path/blob/2ffb90eb51a46ea681a0888e2f381be513a7ade6/src/rust/Cargo.lock#L37

yutannihilation:
Ok, I found the difference. Since {extendrtests} uses the debug profile, it probably contains all the symbols. So, any existing R package that passes `--release` should have no problem.

https://github.com/extendr/extendr/blob/ffc02dbc0047d1016abf0fa99293f5bfeeb06667/tests/extendrtests/src/Makevars#L2

But, anyway, this pull request looks good to me. Thanks!

Ilia-Kosenkov:
How come that people get packages without `--release`. Did we mess up with `{rextendr}` templates?

PR #630, opened 2023-10-01
Add `Option<String>` conversion to `Rstr`

CGMossa:
While helping a user with a type-conversion issue, it came up that `Rstr` doesn't cover `Option<String>` conversion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 - `use_rng` option for the `extendr` attribute macro, which enables the use of
 random number sampling methods from R, e.g. `#[extendr(use_rng = true)` [[#476]](https://github.com/extendr/extendr/pull/476)
 - `[T; N]` conversions to `Robj` [[#594]](https://github.com/extendr/extendr/pull/594/)
-- `Rstr` can be constructed from `Option<String>` now [[#630]](https://github.com/extendr/extendr/pull/630)
+- `Rstr` can now be constructed from `Option<String>`  [[#630]](https://github.com/extendr/extendr/pull/630)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - `use_rng` option for the `extendr` attribute macro, which enables the use of
 random number sampling methods from R, e.g. `#[extendr(use_rng = true)` [[#476]](https://github.com/extendr/extendr/pull/476)
 - `[T; N]` conversions to `Robj` [[#594]](https://github.com/extendr/extendr/pull/594/)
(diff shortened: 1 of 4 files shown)


extendr-api/src/wrapper/rstr.rs
```diff
@@ -61,10 +61,20 @@ impl From<&str> for Rstr {
...
+impl From<Option<String>> for Rstr {
+    fn from(value: Option<String>) -> Self {
+        if let Some(string) = value {
+            Self::from(string)
+        } else {
+            Self { robj: na_string() }
+        }
+    }
+}
+
```
JosiahParry, line 73: I see literally no downsides with this! Beautiful! Just out of curiosity since I still dont really get the use of `if let` what is the use of it here instead of 

```rust
match value {
    Some(v) => Self::from(v),
    None => na_string()
}
```

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - `use_rng` option for the `extendr` attribute macro, which enables the use of
 random number sampling methods from R, e.g. `#[extendr(use_rng = true)` [[#476]](https://github.com/extendr/extendr/pull/476)
 - `[T; N]` conversions to `Robj` [[#594]](https://github.com/extendr/extendr/pull/594/)
+- `Rstr` can be constructed from `Option<String>` now [[#630]](https://github.com/extendr/extendr/pull/630)
```
Ilia-Kosenkov, line 13:
```suggestion
- `Rstr` can now be constructed from `Option<String>`  [[#630]](https://github.com/extendr/extendr/pull/630)
```

Ilia-Kosenkov (approved):
Looks ok

extendr-api/src/wrapper/rstr.rs
CGMossa: `match` looks succinct here. I should have written it like this. But `match` looks innocent, but it is very flexible. `if let` is a match on "patterns" only. While you could have a match arm include a test of value.
So to me, when I see `if let` I see a constrained construction, and it is sort of easier for me to read.
I don't know what to say really. `match` 

See this for an example of "dangerous" uses of `match`: https://github.com/rust-lang/rust/issues/37612

CGMossa:
@Ilia-Kosenkov nightly is causing `extendrtests` to fail due to snapshot tests. Do you mind dealing with this please?

Ilia-Kosenkov:
Ouch, something has been deeply reworked in Rust. I suggest to go with `skip_if_nightly()`. If you don't know how to do that, just merge with failing pipeline & I will fix it later, today or tomorrow.

CGMossa:
Merging this as the only failing thing is snapshots on nightly which aren't part of this anyways.

PR #634, opened 2023-10-04
Use error instead of panic! in extendr-macros

yutannihilation:
Make the `#[extendr]` macro return clear error like this (See `extendr-macros/tests/cases/case-extendr-macro.stderr` for more examples).

```
error: Unexpected key
 --> tests/cases/case-extendr-macro.rs:4:11
  |
4 | #[extendr(foo = true)]
  |   
```

### DIFF
--- a/extendr-macros/tests/trybuild.rs
+++ b/extendr-macros/tests/trybuild.rs
@@ -1,0 +1,5 @@
+#[test]
+fn test_macro_failures() {
+    let t = trybuild::TestCases::new();
+    t.compile_fail("tests/cases/*.rs");
+}
--- a/extendr-macros/Cargo.toml
+++ b/extendr-macros/Cargo.toml
@@ -19,3 +19,4 @@ proc-macro2 = { version = "1.0" }
 extendr-api = { path = "../extendr-api" }
 extendr-engine = { path = "../extendr-engine" }
 libR-sys = { workspace = true }
+trybuild = "1.0"
--- a/extendr-macros/src/extendr_function.rs
+++ b/extendr-macros/src/extendr_function.rs
@@ -6,7 +6,12 @@ use syn::{meta::ParseNestedMeta, ItemFn, Lit, LitBool};
 /// Generate bindings for a single function.
 pub fn extendr_function(mut func: ItemFn, opts: &wrappers::ExtendrOptions) -> TokenStream {
     let mut wrappers: Vec<ItemFn> = Vec::new();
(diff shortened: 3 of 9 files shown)


extendr-macros/tests/trybuild.rs
```diff
@@ -0,0 +1,5 @@
+#[test]
+fn test_macro_failures() {
+    let t = trybuild::TestCases::new();
+    t.compile_fail("tests/cases/*.rs");
```
Ilia-Kosenkov, line 4: Hm, can this do macros expand? Maybe we can use it for snapshot testing macros, instead of doing it through `extendrtests`?

CGMossa: This is the rust way of making these tests. I don't think looping in the weird snapshot tests through R should replace this.

yutannihilation: In my understanding, these cannot be expanded because the macros usages are wrong and fail.

JosiahParry:
This looks like a rather cut and dry PR. Is there anything preventing this PR from being merged?

CGMossa:
I don't think there is any reason not to merge this. I liked it when I first saw it.. And I thought we had merged it..

@yutannihilation please forgive me for not approving this right away. Mind taken an extra look, and then we'll merge it?

yutannihilation:
Yeah, I'm fine and no hurry at all. I think this is low-priority as this is just a small improvement.

CGMossa:
Okay. I'll rebase this and approve soon then

CGMossa:
@yutannihilation Hello! I tried to rebase this. Have you made commits this moment that may have been overwritten by the rebase?

I'm sorry, i'm just getting a bit paranoid here.

yutannihilation:
Thanks. I think nothing is overwritten.

extendr-macros/tests/trybuild.rs
CGMossa: I now understand what you mean. This is not possible to do with testthat as the snapshot tests are expected to succeed, and not fail, like these tests. This is now resolved.

CGMossa (approved):
There's a lot of extremely nice things within this PR.
LGTM

As a follow-up, I believe this `trybuild` can be made a `dev-dependency`, to ease compilation burden. And I wished that there were less explicit type annotations on things. But we'll survive for now.

yutannihilation:
Thanks for merging!

> this `trybuild` can be made a `dev-dependency`,

`trybuild` is already a dev-dependency.

PR #635, opened 2023-10-04
Replace mechanism for single-threaded access to R-API

CGMossa:
I'd like to simplify the single-threaded manner in which we access R-API. This is mainly to understand that, but there is definitely room for improvement.

The challenge is nested calls of `single_threaded`, which entails that the chosen lock shouldn't halt when the same thread repeatedly attempts to acquire it. Otherwise we are waiting for the same thread to finish execution, ergo deadlock.

`ReentrantMutex` from `parking_lot` fulfills this promise. However, commit [c750eb0](https://github.com/extendr/extendr/pull/635/commits/c750eb089e1c4d33151396b1e507c1c46faee8bb) didn't pass. Only one thing broke there, and I don't know why.

Note that there is a

```rs
std::thread::sleep(std::time::Duration::from_millis(1))
```

in current implementation, which means there is a 1ms for each nested call of `single_threaded` call associated with R-api. This might not be of importance in practice, and might not even appear on measurements, but it is still something.

### How can you help?

I'm sure these approaches improves things somehow, but I haven't measured them. I am going to attempt to have a few commits that implements one approach, and another, and those could be tested against each other. 

A possible case could be how long it takes to run `cargo test`, but that's a little circumstantial.

### DIFF
--- a/extendr-api/src/thread_safety.rs
+++ b/extendr-api/src/thread_safety.rs
@@ -2,18 +2,39 @@
 
 use crate::*;
 use once_cell::sync::Lazy;
-use parking_lot::ReentrantMutex;
-use std::sync::Arc;
+use std::cell::RefCell;
+use std::sync::Mutex;
 
-static LOCK: Lazy<Arc<ReentrantMutex<()>>> = Lazy::new(|| Arc::new(ReentrantMutex::new(())));
+static R_API_LOCK: Lazy<Mutex<()>> = Lazy::new(|| Default::default());
+
+thread_local! {
+    static THREAD_HAS_LOCK: RefCell<bool> = RefCell::new(false);
+}
 
 pub fn single_threaded<F, R>(f: F) -> R
 where
     F: FnOnce() -> R,
 {
-    let guard = LOCK.lock();
+    // let already_has_lock = THREAD_HAS_LOCK.with(|flag| *flag.borrow());
+    let has_lock = THREAD_HAS_LOCK.with_borrow(|x| *x);
+
+    // acquire R-API lock
+    let _guard = if !has_lock {
+        Some(R_API_LOCK.lock().unwrap())
+    } else {
+        None
+    };
+
+    // this thread now has the lock
+    THREAD_HAS_LOCK.with_borrow_mut(|x| *x = true);
+
     let result = f();
-    drop(guard);
+
+    // release the R-API lock
+    if _guard.is_some() {
+        THREAD_HAS_LOCK.with_borrow_mut(|x| *x = false);
+    }
+
     result
 }
 
--- a/extendr-api/src/thread_safety.rs
+++ b/extendr-api/src/thread_safety.rs
@@ -102,7 +102,7 @@ where
         }
     }
 
-    unsafe {
+    single_threaded(|| unsafe {
         let fun_ptr = do_call::<F> as *const ();
         let clean_ptr = do_cleanup as *const ();
         let x = false;
@@ -122,5 +122,5 @@ where
         };
         Rf_unprotect(1);
         res
-    }
+    })
 }
--- a/extendr-api/src/thread_safety.rs
+++ b/extendr-api/src/thread_safety.rs
@@ -1,16 +1,15 @@
 //! Provide limited protection for multithreaded access to the R API.
 
 use crate::*;
-use once_cell::sync::Lazy;
-use std::cell::RefCell;
+use std::cell::Cell;
 use std::sync::Mutex;
 
 /// A global lock, that should represent the global lock on the R-API.
 /// It is not tied to an actual instance of R.
-static R_API_LOCK: Lazy<Mutex<()>> = Lazy::new(|| Default::default());
+static R_API_LOCK: Mutex<()> = Mutex::new(());
 
 thread_local! {
-    static THREAD_HAS_LOCK: RefCell<bool> = RefCell::new(false);
+    static THREAD_HAS_LOCK: Cell<bool> = Cell::new(false);
 }
 
 /// Run `f` while ensuring that `f` runs in a single-threaded manner.
@@ -23,8 +22,7 @@ pub fn single_threaded<F, R>(f: F) -> R
 where
     F: FnOnce() -> R,
 {
-    // let already_has_lock = THREAD_HAS_LOCK.with(|flag| *flag.borrow());
-    let has_lock = THREAD_HAS_LOCK.with(|x| *x.borrow());
+    let has_lock = THREAD_HAS_LOCK.with(|x| x.get());
 
     // acquire R-API lock
     let _guard = if !has_lock {
@@ -34,13 +32,13 @@ where
     };
 
     // this thread now has the lock
-    THREAD_HAS_LOCK.with(|x| x.replace(true));
+    THREAD_HAS_LOCK.with(|x| x.set(true));
 
     let result = f();
 
     // release the R-API lock
     if _guard.is_some() {
-        THREAD_HAS_LOCK.with(|x| x.replace(false));
+        THREAD_HAS_LOCK.with(|x| x.set(false));
     }
 
     result
--- a/extendr-api/src/thread_safety.rs
+++ b/extendr-api/src/thread_safety.rs
@@ -1,69 +1,20 @@
 //! Provide limited protection for multithreaded access to the R API.
 
 use crate::*;
-use std::sync::atomic::{AtomicU32, Ordering};
+use once_cell::sync::Lazy;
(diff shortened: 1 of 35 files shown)


CGMossa:
The reentrant mutex idea is failing.

I'm now experimenting with an idea from @wolvereness. Here the idea that each thread will have the lock, and
successive calls to `single_threaded` will first check if current thread has the lock, if not, proceed as usual, or otherwise attempt to acquire the lock!.

CGMossa:
This actually passed CI, and it looks understandable, so I guess I'd like to have it reviewed.

extendr-api/src/thread_safety.rs
```diff
@@ -1,69 +1,49 @@
...
-use std::sync::atomic::{AtomicU32, Ordering};
+use once_cell::sync::Lazy;
+use std::cell::RefCell;
+use std::sync::Mutex;
 
-static OWNER_THREAD: AtomicU32 = AtomicU32::new(0);
-static NEXT_THREAD_ID: AtomicU32 = AtomicU32::new(1);
+/// A global lock, that should represent the global lock on the R-API.
+/// It is not tied to an actual instance of R.
+static R_API_LOCK: Lazy<Mutex<()>> = Lazy::new(|| Default::default());
```
Wolvereness, line 10: This doesn't need to be lazy:
```rs
static R_API_LOCK: Mutex<()> = Mutex::new();
```

extendr-api/src/thread_safety.rs
```diff
@@ -1,69 +1,49 @@
...
 
 thread_local! {
-    static THREAD_ID: u32 = NEXT_THREAD_ID.fetch_add(1, Ordering::SeqCst);
+    static THREAD_HAS_LOCK: RefCell<bool> = RefCell::new(false);
```
Wolvereness, line 13: Only needs to be a `Cell`, which doesn't have lock-checks.
```rs
static THREAD_HAS_LOCK: Cell<bool> = Cell::new(false);
```

extendr-api/src/thread_safety.rs
```diff
@@ -1,69 +1,49 @@
...
-        // wait for OWNER_THREAD to become 0 and put us as the owner.
-        while OWNER_THREAD
-            .compare_exchange(0, id, Ordering::Acquire, Ordering::Relaxed)
-            .is_err()
-        {
-            std::thread::sleep(std::time::Duration::from_millis(1));
-        }
-    }
+    // let already_has_lock = THREAD_HAS_LOCK.with(|flag| *flag.borrow());
+    let has_lock = THREAD_HAS_LOCK.with(|x| *x.borrow());
```
Wolvereness, line 27: If it were `Cell`:
```rs
let has_lock = THREAD_HAS_LOCK.with(|x| x.get());
```

extendr-api/src/thread_safety.rs
```diff
@@ -1,69 +1,49 @@
...
+
+    // acquire R-API lock
+    let _guard = if !has_lock {
+        Some(R_API_LOCK.lock().unwrap())
+    } else {
+        None
+    };
+
+    // this thread now has the lock
+    THREAD_HAS_LOCK.with(|x| x.replace(true));
```
Wolvereness, line 37: If it were `Cell`:
```rs
let has_lock = THREAD_HAS_LOCK.with(|x| x.set(true));
```

extendr-api/src/thread_safety.rs
```diff
@@ -1,69 +1,49 @@
...
 
-    let res = f();
+    let result = f();
 
-    if old_id != id {
-        // release the lock and signal waiting threads.
-        OWNER_THREAD.store(0, Ordering::Release);
+    // release the R-API lock
+    if _guard.is_some() {
+        THREAD_HAS_LOCK.with(|x| x.replace(false));
```
Wolvereness, line 43: If it were `Cell`:
```rs
let has_lock = THREAD_HAS_LOCK.with(|x| x.set(false));
```

Wolvereness (changes requested):

CGMossa:
Thanks for review @Wolvereness !

CGMossa:
I'm now convinced that there is an issue with ExtendR, that is outside of the thread safety rs file, and it was probably masking before.
I'll try to explore what that issue is.
CI is passed from time to time.

CGMossa:
A lot of places are lacking the `single_threaded`. I tried adding it different places.

* Don't add `single_threaded` to `ownership.rs`. Why??

CGMossa:
As far as I'm concerned, the change in this PR is correct. But there are missing cases of `single_threaded` still. This is beyond impossible to track down, as issue is only appearing on CI **sometimes**, and never locally.

Notes:
```md
Deadlock on `pairlist_macro_works`

Access violation on `altrep_tests`
```

I'll slap on more `single_threaded` and hope they disappear.

I think this is revealing issues that were hidden before. Now also in the current altrep implementation.

extendr-api/src/thread_safety.rs
```diff
@@ -1,69 +1,47 @@
...
-static OWNER_THREAD: AtomicU32 = AtomicU32::new(0);
-static NEXT_THREAD_ID: AtomicU32 = AtomicU32::new(1);
+/// A global lock, that should represent the global lock on the R-API.
+/// It is not tied to an actual instance of R.
+static R_API_LOCK: Mutex<()> = Mutex::new(());
 
 thread_local! {
-    static THREAD_ID: u32 = NEXT_THREAD_ID.fetch_add(1, Ordering::SeqCst);
+    static THREAD_HAS_LOCK: Cell<bool> = Cell::new(false);
 }
```
yutannihilation, line 13: Why is this variable needed?

CGMossa: It's due to nested calls of single threaded.
A thread acquires the lock, and performs calculations. That works fine.
If the same thread acquires the lock. And then it performs one calculation, but then the next thing is that it wants to acquire the lock again, then it is waiting in itself to release the lock to acquire it again. This variable tells it, if it already has the lock, I.e if it is already the thread that's allowed to access the r api at the moment.

This is the reentrant part.

yutannihilation: Ah, thanks, makes sense.

(What I felt a bit anxious was that, since this is not an atomic operation, if it can happen `has_lock` is `true` when checked here but is unlocked during `f()`. But, probably it won't happen as long as the things called sequentially.)

``` rust
    let _guard = if !has_lock {
        Some(R_API_LOCK.lock().unwrap())
    } else {
        None
    };
```

CGMossa: Let me think about that.. I mean has lock is a property of the thread, so if won't be released by any other thread or anything, whilst f is running.

But it is incredibly difficult to reason about this.
Im still a bit unclear about the Cell part. I think that's the part that takes care of the "atomic" component...

Wolvereness: > Ah, thanks, makes sense.
> 
> (What I felt a bit anxious was that, since this is not an atomic operation, if it can happen `has_lock` is `true` when checked here but is unlocked during `f()`. But, probably it won't happen as long as the things called sequentially.)

The catch is that it's not even a "probably", it's a "can't".

```rs
    let has_lock = THREAD_HAS_LOCK.with(|x| x.get());

    // acquire R-API lock
    let _guard = if !has_lock {
        Some(R_API_LOCK.lock().unwrap())
    } else {
        None
    };

    // this thread now has the lock
    THREAD_HAS_LOCK.with(|x| x.set(true));

    let result = f();
```

The inherent-nature of a single thread of execution is that there are no code paths that can diverge into another `f()` (or anywhere at all) between the thread-local check, the mutex-lock, and the thread-local set. Other threads contending for mutex can't access the other thread-locals, only its own thread of execution's thread-local.

The worst-case-scenario is that a thread unwinds without releasing the lock. There's no UB, because the mutex will be poisoned for other threads (they wont be able to lock), and the original thread will always assume it already owns the lock.

yutannihilation: Yeah, you are right. Sorry if my wording made you feel uncomfortable.

CGMossa: No offence taken! 

I think it is very difficult with this. I'm proposing an alternative mechanism, that somehow is failing CI,
and we have no way of truly figuring out if I'm write, or if there is another problem on extendr first.

Personally, I'm questioning everything.

I think this is pretty solid, so either (1) extendr is broken somewhere, and the `sleep` just masked that (2) conceptually, there is a missing piece.

In any case, reviews are welcome.

CGMossa:
- [ ] Try again when #642 is reviewed & merged

PR #636, opened 2023-10-04
Use `ty` for types in macros instead of tokens `tt`

CGMossa:
* First, we don't need to export `Index` from `std::ops`; We never implement it for any of our types, nor do we use it as a requirement / super-trait.
* There are instances of `tt`, where it is meant to be `ty` for type.

These changes have zero impact and are purely internal.

### DIFF
--- a/extendr-api/src/prelude.rs
+++ b/extendr-api/src/prelude.rs
@@ -69,8 +69,6 @@ pub use super::iter::StrIter;
 
 pub use std::convert::{TryFrom, TryInto};
 
-pub use std::ops::Index;
-
 pub use super::scalar::*;
 
 pub use super::Nullable::*;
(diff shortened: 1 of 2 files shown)


extendr-api/src/prelude.rs
```diff
@@ -69,8 +69,6 @@ pub use super::iter::StrIter;
 
 pub use std::convert::{TryFrom, TryInto};
 
-pub use std::ops::Index;
```
Ilia-Kosenkov, line 72: I am surprised we don't use it. Shouldn't vector types like `Integers` have it implemented?

Ilia-Kosenkov (approved):

extendr-api/src/prelude.rs
CGMossa: I found out that what we really implement is a `Slice` trait somehow.. And that's what we use. So for `Integers`, would probably implement that, via the wrapped `Robj`.

PR #637, opened 2023-10-05
Release v0.6.0

yutannihilation:


yutannihilation:
Hmm, I wanted to gather all the version specifications into the workspace Cargo.toml, but it seems it requires Rust 1.64 (the current MRSV is 1.60). I bet it should be fine, but let's avoid it here for maximum safety.

eitsupi:
> I bet it should be fine, but let's avoid it here for maximum safety.

FYI, currently the lowest version of Rust on CRAN is 1.69 on Windows.
Debian testing is 1.70 and Ubuntu LTS is 1.66.

I believe Posit Package Manager has recently started building Debian stable and old stable binaries, but these Rusts are too old to consider and will probably not build anything.......

eitsupi:
> Hmm, I wanted to gather all the version specifications into the workspace Cargo.toml, but it seems it requires Rust 1.64 (the current MRSV is 1.60).

By the way, this is a workspace inheritance issue, but after the release to crates.io, the MSRV goes down because workspace inheritance is irrelevant.

CGMossa:
You shouldn't need to change the stuff with `workspace version`. It should be only the version in the root.

Also, I wanted maybe to get this done before a release, https://github.com/extendr/extendr/pull/591

CGMossa:
I see the issue is the MSRV is the one that depends on the workspace feature.... ooops.

yutannihilation:
Thanks, yeah, I too think 1.66 is the lowest version that we should consider, and this shouldn't matter. But, since it's just for convenience, I can do it later. I just want to get this done as soon as possible.

yutannihilation:
Ah, no worries. It's my fault that I forgot to ask for review. Thanks for approving!

PR #639, opened 2023-10-05
Add `.Generic` symbol

CGMossa:
I don't know what the versioning problem is.

This symbol was added to R in 2011.


CGMossa:
Do you remember what was the issue here @andy-thomason?

yutannihilation (approved):
That line was added by https://github.com/extendr/extendr/pull/107, but I found no clue about why it's commented out. I think you can proceed.

PR #640, opened 2023-10-05
Use [workspace.package] in Cargo.toml

yutannihilation:
As of [Rust 1.64](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#cargo-improvements-workspace-inheritance-and-multi-target-builds), `[workspace.package]` can be used to specify the metadata for the whole workspace. This should be easier to maintain.

https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table

A few notes:

* As described above, this requires 1.64, so we have to increase MSRV. As [eitsupi commented](https://github.com/extendr/extendr/pull/637#issuecomment-1748711501), this version requirement should be safe even on CRAN.
* I found the `authors` field differ between crates. I think it's fine to include an author to all extendr crates even if they contributed to only one of the crates.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ authors = [
 edition = "2021"
 
 [workspace.dependencies]
-# When update extendr's version, this version also needs to be updated
+# When updating extendr's version, this version also needs to be updated
 extendr-macros = { path = "./extendr-macros", version = "0.6.0" }
 
 libR-sys = "0.6.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,25 @@ resolver = "2"
 
 members = ["extendr-api", "extendr-engine", "extendr-macros"]
 
+[workspace.package]
+version = "0.6.0"
+repository = "https://github.com/extendr/extendr"
+license = "MIT"
+authors = [
+    "andy-thomason <andy@andythomason.com>",
+    "Thomas Down",
+    "Mossa Merhi Reimert <mossa@sund.ku.dk>",
+    "Claus O. Wilke <wilke@austin.utexas.edu>",
+    "Hiroaki Yutani",
+    "Ilia A. Kosenkov <ilia.kosenkov@outlook.com>",
+]
+rust-version = "1.64"
+edition = "2021"
+
 [workspace.dependencies]
+# When update extendr's version, this version also needs to be updated
+extendr-macros = { path = "./extendr-macros", version = "0.6.0" }
+
 libR-sys = "0.6.0"
 
 [patch.crates-io]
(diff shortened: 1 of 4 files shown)


eitsupi:
cargo-msrv does not support `rust-version` inheritance (foresterre/cargo-msrv#590).
So please set `metadata.msrv` to each Cargo.toml.

yutannihilation:
Thanks, reverted the `rust-version` specification.

CGMossa (approved):
LGTM

I did look into this, and I was also of the same opinion about the authors, but I just wanted things merged.. 

Good work! 

As a side note: I'd love more reviews, especially on my tiny PRs.

Cargo.toml
```diff
@@ -3,7 +3,25 @@ resolver = "2"
...
+    "Mossa Merhi Reimert <mossa@sund.ku.dk>",
+    "Claus O. Wilke <wilke@austin.utexas.edu>",
+    "Hiroaki Yutani",
+    "Ilia A. Kosenkov <ilia.kosenkov@outlook.com>",
+    "Michael Milton <michael.r.milton@gmail.com>",
+]
+edition = "2021"
+
 [workspace.dependencies]
+# When update extendr's version, this version also needs to be updated
```
Ilia-Kosenkov, line 22:
```suggestion
# When updating extendr's version, this version also needs to be updated
```

extendr-api/Cargo.toml
```diff
@@ -28,7 +22,7 @@ num-complex = { version = "0.4", optional = true }
 serde = { version = "1.0", features = ["derive"], optional = true }
 
 [dev-dependencies]
-extendr-engine = { path = "../extendr-engine", version = "0.6.0" }
+extendr-engine = { path = "../extendr-engine" }
```
Ilia-Kosenkov, line 25: Should this be workspace as well?

CGMossa: In theory, yes!, in practice this is such an annoying package, that I don't want to mess with it too much already.
If multiple packages in a workspace depends on a crate, then said crate should be `workspace = true`, in my opinion.

Cargo.toml
yutannihilation: Thanks.

extendr-api/Cargo.toml
yutannihilation: I think the pain point is that we have to rewrite the versions here and there, but It seems `dev-dependencies` doesn't require `version`, so I didn't include in this pull request. I don't have strong opinion here.

CGMossa: I think this should be left like this. If we add another package to the workspace, that needs `extendr-engine`, I'll deal with it then.

PR #641, opened 2023-10-06
Fix unused_mut warning

yutannihilation:
I see this warning:

```
warning: variable does not need to be mutable
   --> src\wrapper\strings.rs:112:13
    |
112 |         let mut robj = Strings::alloc_vector(STRSXP, len);
    |             ----^^^^
    |             |
    |             help: remove this `mut`
    |
    = note: `#[warn(unused_mut)]` on by default
```

### DIFF
--- a/extendr-api/src/wrapper/strings.rs
+++ b/extendr-api/src/wrapper/strings.rs
@@ -109,7 +109,7 @@ impl<T: AsRef<str>> FromIterator<T> for Strings {
         let iter_collect: Vec<_> = iter.into_iter().collect();
         let len = iter_collect.len();
 
-        let mut robj = Strings::alloc_vector(STRSXP, len);
+        let robj = Strings::alloc_vector(STRSXP, len);
         crate::single_threaded(|| unsafe {
             for (i, v) in iter_collect.into_iter().enumerate() {
                 SET_STRING_ELT(robj.get(), i as isize, str_to_character(v.as_ref()));


CGMossa (changes requested):
This shows that `extendr` is unsound. We need to deal with that everywhere. I'd like this warning to stay please.

yutannihilation:
Do you mean this should actually require `mut`? If so, I just close this pull request.

CGMossa:
Yes!

We create a vector here that we fill out. Usually, you need a `&mut` to do
so.
But `GetSexp` only has a `get(&self)`, while it should in fact also carry a
`get_mut(&mut self)`,
that is then fed into `SET_STRING_ELT`. This is how we unsafely pass around
a pointer,
but ensure exclusive-access is maintained properly.

I've tried a few times to add this, and now that I'm working on the
`OwnedRobj`/`PermanentRobj`/`RawRobj` PR, I might figure out how to add
this properly.

CGMossa:
Yes!

We create a vector here that we fill out. Usually, you need a `&mut` to do
so.
But `GetSexp` only has a `get(&self)`, while it should in fact also carry a
`get_mut(&mut self)`,
that is then fed into `SET_STRING_ELT`. This is how we unsafely pass around
a pointer,
but ensure exclusive-access is maintained properly.

I've tried a few times to add this, and now that I'm working on the
`OwnedRobj`/`PermanentRobj`/`RawRobj` PR, I might figure out how to add
this properly.

yutannihilation:
Okay, I don't feel it really matters whether this `mut` is left or not, but let's close now. Thanks for sharing.

Issue #583, opened 2023-07-12
Can extendr give up embedded usage?


yutannihilation:
Seriously, I believe it makes things super complicated for extendr to support both R package use and embedded use. Does it really worth to support the embedded usage?

[The recent trouble about the path to `libR.so`](https://github.com/extendr/extendr/issues/580) is the case; libR-sys sets [`rustc-link-search`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-link-search), but this isn't used for R packages because linking against libR happens AFTER `cargo build`, iiuc. libR-sys's `build.rs` can be much simpler if it gives up the embedded usage.

[This tricky implementation of `NA` for `&str`](https://github.com/extendr/extendr/pull/477#issuecomment-1423452814) is also the case. See https://github.com/extendr/extendr/issues/483#issuecomment-1435499525 for the details. In short, it handles the problems that can occur only when multiple R sessions exist, which should never happen as long as extendr is used in R packages.

Placing `single_threaded()` here and there is also for this reason. I believe most of them are unnecessary except when it's used in some embedded usage. As long as it's called from R session, most of the things should be synchronous.

But, to be fair, giving up embedded usage means you cannot have `extendr_engine()`, which means you cannot have `test!` macro, which means you cannot do any unit testing at all. You'll always need a real R session and real R package. So, I guess no one will agree with this idea, but let me express my frustration here.

andy-thomason:
Yes. The whole raison d'etre of the embedded mode is for testing.

I had thought about making R libraries available on crates.io, too.

We could provide a config to no op `single_threaded()` when run from and R
package. That would improve performance.

andy-thomason:
This would also make Robj reference counting much cheaper.

CGMossa:
I'm somewhat against this, and I'd like to elaborate soon on it.

In the mean time, please don't do this before I've given my opinion on this.

multimeric:
But has anyone answered the question of how we would do testing without embedded R?

CGMossa:
You cannot.
You cannot allocate.

yutannihilation:
> The whole raison d'etre of the embedded mode is for testing.

Oh, really? I didn't know this. Then, if only we can find some alternative way of testing, we might not need this. e.g. I was wondering if it's possible to inject fake libR for unit testing.

> In the mean time, please don't do this

Don't worry, this would require enormous amount of efforts. No one can do this overnight :)

[...]

PR #642, opened 2023-10-07
Added `single_threaded` on things

CGMossa:
I looked over all instances of `unsafe`, which must contain all access to R-API.

There are cases where a `single_threaded` is presumed, but it isn't wrapping the R-API call.

A silly example is:
```rs
pub fn current_env() -> Environment {
    unsafe { Robj::from_sexp(R_GetCurrentEnv()).try_into().unwrap() }
}
```
which translates to
```rs
    unsafe { {
        let sexp = R_GetCurrentEnv();
        single_threaded(|| {
            unsafe { ownership::protect(sexp) };
            Robj { inner: sexp }
        })
    }.try_into().unwrap() }
```
hence there is an R-API call not within a `single_threaded`.

I've only added `single_threaded` to calls that either mutates or allocates something.. Or things I think could be suspicious.

This might be a cure for #635, but that's not the objective here.


extendr-api/src/iter.rs
```diff
@@ -83,7 +83,7 @@ impl Iterator for StrIter {
     }
 
     fn next(&mut self) -> Option<Self::Item> {
-        unsafe {
```
Ilia-Kosenkov, line 86: I am not entirely convinced this should be single-threaded.

extendr-api/src/robj/mod.rs
```diff
@@ -580,7 +580,7 @@ impl Robj {
     /// }
     /// ```
     pub fn as_str<'a>(&self) -> Option<&'a str> {
-        unsafe {
```
Ilia-Kosenkov, line 583: This one also looks like read-only access, do we need `single_threaded` here?

extendr-api/src/robj/rinternals.rs
```diff
@@ -77,7 +77,7 @@ pub trait Rinternals: Types + Conversions {
 
     /// Convert to a string vector.
     fn as_character_vector(&self) -> Robj {
-        unsafe { Robj::from_sexp(Rf_asChar(self.get())) }
```
Ilia-Kosenkov, line 80: This one also does not strike me as thread-unsafe.

extendr-api/src/wrapper/list.rs
```diff
@@ -267,7 +267,9 @@ impl Iterator for ListIter {
         if i >= self.len {
             None
         } else {
-            Some(unsafe { Robj::from_sexp(VECTOR_ELT(self.robj.get(), i as isize)) })
+            Some(unsafe {
+                Robj::from_sexp(single_threaded(|| VECTOR_ELT(self.robj.get(), i as isize)))
```
Ilia-Kosenkov, line 271: Again, getting an element should be thread-safe given CoW behavior.

extendr-api/src/wrapper/macros.rs
```diff
@@ -96,7 +96,7 @@ macro_rules! gen_vector_wrapper_impl {
                     if(index >= self.len()) {
                         <$scalar_type>::na()
                     } else {
-                        unsafe { [<$r_prefix _ELT>](self.get(), index as R_xlen_t).into() }
+                        single_threaded(||unsafe { [<$r_prefix _ELT>](self.get(), index as R_xlen_t).into() })
```
Ilia-Kosenkov, line 99: Here also.

Ilia-Kosenkov (commented):
We need to be careful. Every `single_threaded()` is a performance hit, not every `unsafe` call into `libR-sys` is thread-unsafe.

CGMossa:
> Member

Sure. But I'd like to remind that #635 is an attempt to remove the performance penalty, and it has uncovered that there is *something* going on here.

Besides, I'm not sure, that read-only behaviour shouldn't be `single_threaded`. But I won't debate this here.

Ilia-Kosenkov:
All I am saying that maybe we don't need it there, I do not have all of the intricacies of extendr in my head all the time, but it is a guess as to why there was no `single_threaded` there before.

Ilia-Kosenkov (approved):
Let's roll, but with care. I hope we have extensive enough test coverage.

yutannihilation:
Let me leave a comment, not to block, but to shamelessly advertise my take on this, https://github.com/extendr/extendr/issues/583, again.

Is this a problem that will disappear if you add `--test-threads=1`? If so, I personally don't feel it's a good idea to make the code complicated just to fix the error that won't happen on a real R session.

CGMossa:
@yutannihilation thanks for the suggestion. I haven't thought about removing all embedded usage of R.
I need a longer discussion on this, and with answers on what limitations this puts on extendr.
I thought the reason for this, is that someone can use R-constructs in situations like `rayon`, or
if someone wants to make intricate concurrent stuff with R & Rust. 

As it stands, I simply see a construct in `extendr`, that isn't 100% fulfilling its purpose, then I figure out how it may be improved, and I propose a PR. 
I do not have the global overview like you have, to propose to eliminate something all together.
Personally, I would have no problem "fulfilling" extendr's API first, so that it works, and then we can later propose to remove these things all together. Like the type-checking on `#[extendr]`-impl blocks, or this, or even `extendr-engine`. But for the sake of discussing accurately, I think these things should be working 100%.

For instance, we have a `sleep` in the single-threaded stuff, this has made @Ilia-Kosenkov say that there is a performance penalty associated with `single_threaded`. That is true, regardless, but a `sleep` is a much more severe issue, then just the stuff I've proposed in #635.

Anyways, to summarise: I don't mind if you make issue/PR that suggests removing *something*, even if I just worked on improving it. I'm just improving it, because it is part of the package at the moment.

yutannihilation:
> I haven't thought about removing all embedded usage of R.

To be clear, it's not my suggestion here. I just suggested `--test-threads=1`.

> Personally, I would have no problem "fulfilling" extendr's API first

Yeah, it's fine to focus on the API. But, this isn't about API. You made this pull request just to pass tests that is unlikely to fail on a real R session. That's what I meant.

> I thought the reason for this, is that someone can use R-constructs in situations like `rayon`, or
> if someone wants to make intricate concurrent stuff with R & Rust.

I might be wrong, but I haven't heard it's in the current scope of extendr (I know it's listed in the future goal in the README). If you really think so, you should explicitly add tests about concurrency-safety instead of relying on the side effect of not specifying `--test-threads=1`.

CGMossa:
Great points.
Yes, I should make concurrency tests... I just don't have a clue how to do that. I'll think about!

yutannihilation:
No, I'm not primarily suggesting adding concurrency tests. Before anything, you should discuss if extendr should support concurrency by default or not.

PR #643, opened 2023-10-07
Refactor passing of `&str` to r

CGMossa:
* Instead of constructing a `CString` by hand, we should just use `CString`.
*`thread_safety` is passing a string
without ensuring it is a null-terminated,
c-string.
* `make_symbol` should just use `Rf_installTrChar`.

### DIFF
--- a/extendr-api/src/lang_macros.rs
+++ b/extendr-api/src/lang_macros.rs
@@ -45,13 +45,8 @@ macro_rules! args {
 #[doc(hidden)]
 pub unsafe fn append_with_name(tail: SEXP, obj: Robj, name: &str) -> SEXP {
     single_threaded(|| {
-        let mut name = Vec::from(name.as_bytes());
-        name.push(0);
         let cons = Rf_cons(obj.get(), R_NilValue);
-        SET_TAG(
-            cons,
-            Rf_install(name.as_ptr() as *const std::os::raw::c_char),
-        );
+        SET_TAG(cons, crate::make_symbol(name));
         SETCDR(tail, cons);
         cons
     })
@@ -66,11 +61,7 @@ pub unsafe fn append(tail: SEXP, obj: Robj) -> SEXP {
 
 #[doc(hidden)]
 pub unsafe fn make_lang(sym: &str) -> Robj {
-    let mut name = Vec::from(sym.as_bytes());
-    name.push(0);
-    let sexp =
-        single_threaded(|| Rf_lang1(Rf_install(name.as_ptr() as *const std::os::raw::c_char)));
-    Robj::from_sexp(sexp)
+    Robj::from_sexp(single_threaded(|| Rf_lang1(crate::make_symbol(sym))))
 }
 
 /// Convert a list of tokens to an array of tuples.
--- a/extendr-api/src/thread_safety.rs
+++ b/extendr-api/src/thread_safety.rs
@@ -77,7 +77,10 @@ where
 {
     match std::panic::catch_unwind(f) {
         Ok(res) => res,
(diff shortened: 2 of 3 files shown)


CGMossa:
@yutannihilation can I please ask you diagnose something for me? I seem unable to actually **use** `Rf_installNoTrChar`. It is present in all the `bindings.rs` files. But it simply isn't in `libR.so` (I'm on linux right now).
I have no clue how to diagnose this otherwise.

yutannihilation:
Oh, it seems it's intentionally hidden. If possible, libR-sys should filter out such functions.

https://github.com/wch/r-source/blob/ad9c596f97904f330c03cfaed9f8ff21200ed9de/src/main/names.c#L1281
https://github.com/wch/r-source/blob/ad9c596f97904f330c03cfaed9f8ff21200ed9de/src/include/Defn.h#L49

One more (slightly) bad news is that, it seems `Rf_installTrChar` is a "non-API". The release note of R 4.0.0 says:

> installChar is now remapped in ‘Rinternals.h’ to installTrChar, of which it has been a wrapper since R 3.6.0. **Neither are part of the API**, but ...
(https://cran.r-project.org/doc/manuals/r-devel/NEWS.html)

I feel the nuance is not that they prohibit the use of it as it's already used in several packages, so I think this can be a choice.

CGMossa:
That's very unfortunate. I do not understand why all these useful things has to be non-API.

+ I'm unable to find these functions mentioned in `nonAPI.txt`, even if I run the command to regenerate it.

This PR contains a fix to a potential issue (definite issue) with `handle_panic`, that wasn't passing a null-terminated string to `Rf_error`.  

Right now, in `append_with_name` and `make_lang`:
1. A str is made into bytes and these are stored in a `Vec` (allocation)
2. A null is added, and this is why this is done
3. A pointer to that is sent off...

I wanted to side-step an allocation by using those functions, which wouldn't necessarily require a null terminated string.

Anyways, the **real issue** is also that we are returning a pointer to a locally allocated thing. 
```rs
fn foo(a: &str)-> &u8 {
    let vec = Vec::from(a.as_bytes());
    return &vec[0]
}
```
```
   Compiling playground v0.0.1 (/playground)
error[E0515]: cannot return value referencing local variable `vec`
 --> src/lib.rs:3:12
  |
3 |     return &vec[0]
  |            ^---^^^
  |            ||
  |            |`vec` is borrowed here
  |            returns a value referencing data owned by the current function

For more information about this error, try `rustc --explain E0515`.
```

It only works, because we are escaping from Rust, by passing a pointer.
But unless `Rf_install` does its thing completely in that call, and the pointer is no longer used,
there is an issue.

But it has worked fine until now! I have revised it now, and I hope we can review & merge this for clarity, and also the issue with `handle_panic`.

yutannihilation:
> I'm unable to find these functions mentioned in `nonAPI.txt`, even if I run the command to regenerate it.

`nonAPI.txt` collects only the functions explicitly marked as non-API. In R, the definition of "API" is so crazy as

> Entry points which are documented in this manual and declared in an installed header file.
(https://cran.r-project.org/doc/manuals/r-release/R-exts.html#The-R-API)

`Rf_installTrChar()` is not explained in WRE, so it's a "non-API". Good news is that it's not the type of non-API that gets the warning, so it's safe in the sense of whether or not it passes CRAN check, but unsafe in the sense of whether or not the function can be changed or deleted without notice. In actual, it's unlikely that the R core casually deletes a function that is already used in several packages on CRAN.

CGMossa:
@Ilia-Kosenkov I think it is ready to be merged.

Ilia-Kosenkov (approved):
Ok, let's roll

PR #644, opened 2023-10-07
Add `get_mut` to deal with unsoundness of `Robj`

CGMossa:
In `GetSexp` we have a `get`, but not a `get_mut`. While these functions return pointers, it doesn't mean the semantics of exclusive (`mut`) and shared (`ref`) should be completely neglected.

I believe this is referred to as unsoundness.

EDIT:

- [ ] It is important that the unsoundness from `Clone for &Robj` is dealt with as well.

### DIFF
--- a/extendr-api/src/robj/mod.rs
+++ b/extendr-api/src/robj/mod.rs
@@ -132,6 +132,8 @@ pub trait GetSexp {
     /// Access to a raw SEXP pointer can cause undefined behaviour and is not thread safe.
     unsafe fn get(&self) -> SEXP;
 
+    unsafe fn get_mut(&mut self) -> SEXP;
+
     /// Get a reference to a Robj for this type.
     fn as_robj(&self) -> &Robj;
 
@@ -141,6 +143,10 @@ pub trait GetSexp {
 
 impl GetSexp for Robj {
     unsafe fn get(&self) -> SEXP {
+        self.inner
+    }
+
+    unsafe fn get_mut(&mut self) -> SEXP {
         self.inner
     }
 
@@ -175,7 +181,7 @@ pub trait Slices: GetSexp {
     /// Not all obejects (especially not list and strings) support this.
     unsafe fn as_typed_slice_raw_mut<T>(&mut self) -> &mut [T] {
         let len = XLENGTH(self.get()) as usize;
-        let data = DATAPTR(self.get()) as *mut T;
+        let data = DATAPTR(self.get_mut()) as *mut T;
         std::slice::from_raw_parts_mut(data, len)
     }
 }
@@ -849,7 +855,7 @@ pub trait Attributes: Types + Length {
     ///    assert_eq!(robj.get_attrib(sym!(xyz)), Some(r!(1)));
     /// }
     /// ```
-    fn set_attrib<N, V>(&self, name: N, value: V) -> Result<Robj>
+    fn set_attrib<N, V>(&mut self, name: N, value: V) -> Result<Robj>
     where
         N: Into<Robj>,
         V: Into<Robj>,
@@ -857,7 +863,7 @@ pub trait Attributes: Types + Length {
         let name = name.into();
         let value = value.into();
         unsafe {
-            let sexp = self.get();
+            let sexp = self.get_mut();
             single_threaded(|| {
                 catch_r_error(|| Rf_setAttrib(sexp, name.get(), value.get()))
                     .map(|_| Robj::from_sexp(sexp))
@@ -981,7 +987,7 @@ pub trait Attributes: Types + Length {
     ///     assert_eq!(obj.inherits("a"), true);
     /// }
     /// ```
-    fn set_class<T>(&self, class: T) -> Result<Robj>
+    fn set_class<T>(&mut self, class: T) -> Result<Robj>
     where
         T: IntoIterator,
         T::IntoIter: ExactSizeIterator,
--- a/extendr-api/src/robj/into_robj.rs
+++ b/extendr-api/src/robj/into_robj.rs
@@ -631,7 +631,7 @@ pub trait RobjItertools: Iterator {
         Robj: AsTypedSlice<'a, Self::Item>,
         Self::Item: 'a,
     {
-        let vector = self.collect_robj();
+        let mut vector = self.collect_robj();
         let prod = dims.iter().product::<usize>();
         if prod != vector.len() {
             return Err(Error::Other(format!(
--- a/extendr-api/src/wrapper/complexes.rs
+++ b/extendr-api/src/wrapper/complexes.rs
@@ -63,7 +63,7 @@ impl DerefMut for Complexes {
     /// Treat Complexes as if it is a mutable slice, like `Vec<Rcplx>`
     fn deref_mut(&mut self) -> &mut Self::Target {
         unsafe {
-            let ptr = DATAPTR(self.get()) as *mut Rcplx;
+            let ptr = DATAPTR(self.get_mut()) as *mut Rcplx;
             std::slice::from_raw_parts_mut(ptr, self.len())
         }
     }
--- a/extendr-api/src/wrapper/doubles.rs
+++ b/extendr-api/src/wrapper/doubles.rs
@@ -54,7 +54,7 @@ impl Doubles {
 impl Doubles {
     pub fn set_elt(&mut self, index: usize, val: Rfloat) {
         unsafe {
-            SET_REAL_ELT(self.get(), index as R_xlen_t, val.inner());
+            SET_REAL_ELT(self.get_mut(), index as R_xlen_t, val.inner());
         }
     }
 }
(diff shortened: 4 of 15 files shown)


CGMossa:
This deals with the warning in #641.

CGMossa:
I've requested a review of this to you @yutannihilation, because we have already somewhat initiated a conversation about this.

CGMossa:
@Ilia-Kosenkov I need to make snapshot tests pass, as I did indeed change macros.

Can you please provide me a way to do this? Note that you have to be a bit explicit, as the whole `extendrtests` thing is very sensitive. Thx!

Ilia-Kosenkov:
So you changed some code that is generated as part of macros expansion. The easiest way is to run tests locally, observe the diff and then accept snapshot using `{testthat}` command (you will see a prompt).

yutannihilation:
In theory, as the definition of `SEXP` is `*mut`, shouldn't all API call require `mut` then?

``` rust
pub type SEXP = *mut SEXPREC;
```
(https://extendr.github.io/libR-sys/libR_sys/type.SEXP.html)

Honestly, I'm not sure if there are any API call on SEXP that is guaranteed NOT to mutate the R object. For example, even `*_ELT()` might change the state of the underlying data if it's `ALTREP`. If it's always with `mut`, then isn't it effectively the same as always without `mut`?

The example I picked might be just a corner case, and I don't think the direction of this pull request is all wrong, but at the moment I have no idea what is the supposed interface.

CGMossa:
> In theory, as the definition of `SEXP` is `*mut`, shouldn't all API call require `mut` then?

It is a little complex I think. A pointer isn't a `&mut` or `&`, but we are treating it like that.
At the very least, the current change makes the rust code make sense. If you allocate a `Vec`,
and then insert elements into it, you'd need to set that `Vec` to `mut`.

> Honestly, I'm not sure if there are any API call on SEXP that is guaranteed NOT to mutate the R object. For example, even `*_ELT()` might change the state of the underlying data if it's `ALTREP`. If it's always with `mut`, then isn't it effectively the same as always without `mut`?

I think you're 100% correct here. Regardless of this, I think we want to maintain "ownership" / soundness / whatever it is called, when we are within the Rust part of the interface. I am quite sure that `SEXP`s change
a lot. I believe they point to themselves and even another object, and they contain GC information, and all sorts of nonsense. It is more like that `Robj`, encompass a pointer, and the interior mutability of the underlying data, not pointer, we can at least pretend we are upholding properly.

Sorry, I'm not a rust-systems guy enough to actually have the right, succinct explanation for this.

> The example I picked might be just a corner case, and I don't think the direction of this pull request is all wrong, but at the moment I have no idea what is the supposed interface.

Should I ask a Rust person to review this? @jojojet has been kind enough to give some of our PRs a good luck, and they are very much involved with projects that does.. somewhat similar things.

yutannihilation:
My question is, if the API call should require `mut`, isn't it what should be enforced on libR-sys (or something intermediate between)'s side? I'm afraid it's not significantly different to carefully choose `.get()` or `.get_mut()` than to carefully use `.get()` if it still depends on our carefulness. You still can mistakenly use `.get()` on where it should be `.get_mut()`, right?

CGMossa:
Good points! And I'd totally be up for writing an intermediary layer between extendr and libR-sys that was more.. minimal. For now, I think it is fine these things are intertwined with extendr, until we figure them out.

I'd like to point out, that the meaning of `unsafe`, is that you "carefully" write code, and that this code
is reviewed, and constantly up for discussion, because it isn't obvious why it is okay, because the compiler
aren't enforcing anything, but that it is okay because we said so (through stating `unsafe`).

Although, @jojojet relayed to me, that this is okay for now, but the fact that `Clone` is implemented for `&Robj` makes this attempt a little short. I asked @jojojet to put his comment here, so I won't put anymore words in his mouth.

Awesome stuff all around folks!

CGMossa:
I do not know this is failing.
```
rror: test failed, to rerun pass `--test serializer_tests`
  
  Caused by:
    process didn't exit successfully: `D:\a\extendr\extendr\target\x86_64-pc-windows-gnu\debug\deps\serializer_tests-c8bc75661f3b52d6.exe --nocapture` (exit code: 0xc0000028)
  Error: 'Test extendr-api \w tests for x86_64-pc-windows-gnu target': cargo failed with code -1073741784 (args: +stable-msvc test --manifest-path extendr-api/Cargo.toml --features tests --target=x86_64-pc-windows-gnu  -- --nocapture)
ci-cargo: D:\a\_temp\36b2f20c-9c14-49dc-b32d-b4e9d3a84c70.ps1:16
Line |
  16 |    ci-cargo +$toolchain test --manifest-path extendr-api/Cargo.toml -- …
     |    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | 'Test extendr-api \w tests for x86_64-pc-windows-gnu target': cargo failed with code -1073741784 (args:
     | +stable-msvc test --manifest-path extendr-api/Cargo.toml --features tests --target=x86_64-pc-windows-gnu  --
     | --nocapture)
```

I've run this test locally, and it works. You have to run `cargo test --test serializer_tests --features serde -- --nocapture`, to see it. 

When I search, the only thing of note that I find, is this (I searched on the exact exit code from the test).

<https://github.com/bytecodealliance/wasmtime/issues/1967>

Please hellp.

extendr-api/src/robj/mod.rs
```diff
@@ -849,15 +855,15 @@ pub trait Attributes: Types + Length {
     ///    assert_eq!(robj.get_attrib(sym!(xyz)), Some(r!(1)));
     /// }
     /// ```
-    fn set_attrib<N, V>(&self, name: N, value: V) -> Result<Robj>
+    fn set_attrib<N, V>(&mut self, name: N, value: V) -> Result<Robj>
```
Ilia-Kosenkov, line 858: But why do you need a `&mut` here if setting attribute (or class) results in the construction of a new `Robj`?

Ilia-Kosenkov (commented):
So how does it work if both methods return `SEXP`? `SEXP` is a pointer to `SEXPREC` structure, so if we returned something like `*const SEXPREC` from `.get()` and `*mut SEXPREC` from `get_mut()`, it would've been better, but I guess it is incompatible with the rest of the API in libR-sys.

extendr-api/src/robj/mod.rs
CGMossa: If it does indeed return a whole new object, then it shouldn't be `mut`.
But I'm not sure that's the case.
I dug, and found that this is
```C

/* 'name' should be 1-element STRSXP or SYMSXP */
SEXP setAttrib(SEXP vec, SEXP name, SEXP val)
{
    PROTECT(vec);
    PROTECT(name);

    if (isString(name)) {
	PROTECT(val);
	name = installTrChar(STRING_ELT(name, 0));
	UNPROTECT(1);
    }
    if (val == R_NilValue) {
	/* FIXME: see do_namesgets().
	if (name == R_NamesSymbol && isOneDimensionalArray(vec)) {
	    UNPROTECT(2);
	    return removeAttrib(vec, R_DimNamesSymbol);
	}
	*/
	UNPROTECT(2);
	return removeAttrib(vec, name);
    }

    /* We allow attempting to remove names from NULL */
    if (vec == R_NilValue)
	error(_("attempt to set an attribute on NULL"));

    UNPROTECT(2);

    if (name == R_NamesSymbol)
	return namesgets(vec, val);
    else if (name == R_DimSymbol)
	return dimgets(vec, val);
    else if (name == R_DimNamesSymbol)
	return dimnamesgets(vec, val);
    else if (name == R_ClassSymbol)
	return classgets(vec, val);
    else if (name == R_TspSymbol)
	return tspgets(vec, val);
    else if (name == R_CommentSymbol)
	return commentgets(vec, val);
    else if (name == R_RowNamesSymbol) // "row.names" -> care for data frames
	return row_names_gets(vec, val);
    else
	return installAttrib(vec, name, val);
}
```

Take `classgets`, it says that it is more `attr(x, "class") <- value`. 

I think what this returns is the symbol and not the class... Which might be a bigger problem haha..
Especially if that's true, then this doesn't need protection either..

CGMossa:
> So how does it work if both methods return `SEXP`? `SEXP` is a pointer to `SEXPREC` structure, so if we returned something like `*const SEXPREC` from `.get()` and `*mut SEXPREC` from `get_mut()`, it would've been better, but I guess it is incompatible with the rest of the API in libR-sys.

I think it is important to make sure that the rust api has the proper semantics. And also this is a little confined change that I think we can agree on and approve.

Part of my rework of `Robj`, that I've described in #608, etc., is to make `Robj` work more like a `Box`, then what it is right now, which is an `Rc`.

We cannot really return `*const SEXPREC`, because nowhere does the R-API really use that...

Ilia-Kosenkov (approved):

PR #645, opened 2023-10-08
Add `clap` to parse cmd args of `xtask`

Ilia-Kosenkov:
- Use `clap` to parse cmd args
- Split tasks into separate modules
- Task implementation not yet fixed

### DIFF
--- a/xtask/src/commands/r_cmd_check.rs
+++ b/xtask/src/commands/r_cmd_check.rs
@@ -1,0 +1,17 @@
+use xshell::{Error, Shell};
+
+pub(crate) fn run(shell: &Shell, no_build_vignettes: bool) -> Result<(), Error> {
+    let mut cmd = shell
+        .cmd("R")
+        .arg("CMD")
+        .arg("check")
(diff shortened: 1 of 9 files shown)


xtask/src/commands/r_cmd_check.rs
```diff
@@ -0,0 +1,17 @@
+use xshell::{Error, Shell};
+
+pub(crate) fn run(shell: &Shell, no_build_vignettes: bool) -> Result<(), Error> {
+    let mut cmd = shell
```
CGMossa, line 4: why not use the `cmd!` macro here?

CGMossa (approved):
Merging anyways, but please answer the pending comment for my sanity.

xtask/src/commands/r_cmd_check.rs
Ilia-Kosenkov: I did not want to duplicate the string to avoid mistakes when you fix one branch of if but forget about another. And macro does not accept non-literals

PR #646, opened 2023-10-09
Use `CStr` to interface with R

CGMossa:


yutannihilation:
I didn't notice extendr doesn't use `CStr` here. While I agree using `CStr` is a good idea, you should be aware this change have some performance impact. In my understanding, the main points of using `CStr` are:

1. properly determine the length by parsing the bytes until hitting NUL.
2. reject invalid UTF-8 characters.

I think the length check is more significant as you have to parse all the bytes. To avoid this, `CStr::from_bytes_with_nul_unchecked()` might be an option. But, I'm not sure in what case the result of `Rf_xlength()` and the actual length doesn't match.

https://github.com/rust-lang/rust/blob/093b9d5b295d85e144d0ee7da65ea03987214c06/library/core/src/ffi/c_str.rs#L697-L722

Regarding the UTF-8 validity check, it's probably necessary for safety; considering the complexity of UTF-8, it's possible that the definitions of "valid UTF-8" can be different between R and Rust.

(Note that, regarding UTF-8 validity, it's safer to call `Rf_translateCharUTF8()` to handle the case when the user's encoding is non-UTF-8 multi-byte. But, this happen only on very old Windows, so I think it's reasonable to ignore this.)

yutannihilation:
Two updates (mainly reply to self).

> this change have some performance impact.

I might be wrong here, sorry. The reason I thought so was that I was wondering about why my framework is slower than extendr to handle strings ([benchmark](https://github.com/yutannihilation/savvy-benchmark#string-input)), and I came up with the idea that this difference matters. But, even if I switch to `std::str::from_utf8_unchecked()`, it didn't get improved much. So, I guess this doesn't have a significant impact.

> `Rf_translateCharUTF8()`

`R_CHAR()` returns the string stored in [the global `CHARSXP` cache](https://cran.r-project.org/doc/manuals/r-devel/R-ints.html#The-CHARSXP-cache), but the result of `Rf_translateCharUTF8()` is placed on a `R_alloc()`-ed memory. Probably the former will never get GC-ed, while the latter can. So, probably it's better to avoid using `Rf_translateCharUTF8()` for simplicity.

CGMossa:
Alright. @yutannihilation do you mind reviewing this? You know this the most anyways. I think a simpler and more complete API, plus something that fully supports utf8, is preferred anyways..

yutannihilation:
Sorry, could you explain what's the strategy of this PR? By `unwrap()`, this implementation assumes the input is the valid UTF-8. How is it different from the status quo in terms of UTF-8 support?

CGMossa:
> Sorry, could you explain what's the strategy of this PR? By `unwrap()`, this implementation assumes the input is a valid UTF-8. How is it different from the status quo in terms of UTF-8 support?

Hello! Sorry for not answering this. I've finally figured out _things_ here. We do the length check already. We could avoid it by relaying on the behavior, that is, Rf_xlength returns the number of bytes for `CHARSXP`s, and that seems to work very reliably, even with emojis and the like. I've implemented this in #761.

I believe once that is done, we probably should provide Cstr as another data-type to be used, and we can do that without the null-check. But I'll have to think about that. This would have been useful, if there were more C-APIs that require C-strings.

For this PR, I think the idea was, if we are already checking length, why don't we let the Rust compiler do the checking for us. But with #761, we omit this entirely.

PR #649, opened 2023-10-28
[CI] Use `cargo extendr r-cmd-check` to run `{extendrtests}`

Ilia-Kosenkov:
- Switch to using `cargo extendr r-cmd-check` to run `{extendrtests}` on CI (and also replace local paths)
- Add support for `--error-on` option to control when `R CMD check` fails
- Add support for `--check-dir`

### DIFF
--- a/xtask/src/cli/r_cmd_check.rs
+++ b/xtask/src/cli/r_cmd_check.rs
@@ -1,3 +1,4 @@
+use crate::commands::r_cmd_check::RCmdCheckErrorOn;
 use clap::{Args, ValueEnum};
 
 #[derive(Args, Debug)]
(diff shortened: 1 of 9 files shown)


CGMossa:
Dear @yutannihilation , what are your thoughts on this?
We introduced xtask, very uncontroversial feature to enhance developer / contributor experience.
But now, maybe we can replace some of the complicated CI procedure with them.

Since you seem to be the CI expert, along with Ilia, I would be more comfortable, if you gave your thoughts about this.

yutannihilation:
I see no major concern on this change. One thing is that this loses `check-dir`, but it seems this action doesn't use the check result anyway at the moment, so I think it's fine.

CGMossa (approved):
LGTM

Ilia-Kosenkov:
I believe `check_dir`, while insignificant in this particular case, is valuable feedback. I added support for `--check-dir` which respects the directory from which it is executed when a relative path is used.

xtask/src/cli/r_cmd_check.rs
```diff
@@ -0,0 +1,36 @@
...
+pub(crate) struct RCmdCheckArg {
+    #[arg(long, default_value = "false", help = "Passed to R CMD check")]
+    pub(crate) no_build_vignettes: bool,
+    #[arg(
+        long,
+        default_value = "warning",
+        help = "Determines which R CMD check errors to fail on"
+    )]
+    pub(crate) error_on: ErrorOn,
+    #[arg(long, help = "If not specified, temp dir is used")]
```
CGMossa, line 15:
```suggestion
    #[arg(long, help = "Defaults to a temporary directory")]
```

CGMossa (approved):
LGTM

PR #651, opened 2023-10-29
Fix notes on `extendrtests/README.md`

CGMossa:
Now that I can run `R CMD check` on `extendrtests`, I can see that some links are broken, and things needs to be updated.

### DIFF
--- a/tests/extendrtests/README.md
+++ b/tests/extendrtests/README.md
@@ -1,26 +1,34 @@
 # Integration Tests for Calling Extendr from an R Package
 
 [![R build status](https://github.com/extendr/extendr/workflows/Tests/badge.svg)](https://github.com/extendr/extendr/actions)
-[![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
+[![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/license/mit/)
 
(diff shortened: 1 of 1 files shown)


tests/extendrtests/README.md
```diff
@@ -1,26 +1,34 @@
...
+This is necessary so the relative paths in `Cargo.toml` points to the correct location.
 
 Clicking "Check" in RStudio will not work because it runs `devtools::check()` (that resolves to `rcmdcheck::rcmdcheck()`) with no arguments and the checks are then run in a temporary directory that doesn't have the same relative path.
 
+## Running tests via `cargo`
+
+It is also possible to run `R CMD check` in the root directory of `extendr`,
+by invoking `cargo extendr`:
+
+```shell
```
eitsupi, line 32: As an aside, Pandoc, which is familiar to R users, does not consider `shell` to be a valid language on code blocks on Markdown.
I think it's common to use `sh`.

CGMossa: Didn't know that! Thanks! I have seen people use `default`, which I thought was obnoxious. `sh` it is from now on..

yutannihilation: I think it depends on the main purpose of the README. If it's mainly for being converted by pandoc, that's true. If it's mainly for displaying on the GitHub repo, you should rather care about the GitHub's spec. For example, I often use `console`, which is not very common but is supported on GitHub.

Ilia-Kosenkov (approved):

PR #653, opened 2023-10-29
`set_class` returns the same object!

CGMossa:
While working on #644, we questioned whether `set_class` should take `&mut self`.

First, `set_class` returns the object itself back, not the class-label as I erroneously hinted at before.
Secondly, `set_class` modifies the object in-place. It does not return a new object.  

## Appendix

```rs
use extendr_api::{Attributes, GetSexp};
use extendr_engine::with_r;
use extendr_macros::list;
use libR_sys::{R_compute_identical, Rboolean_TRUE, Rf_PrintValue};

#[test]
fn test_what_is_returned_from_set_class() {
    with_r(|| {
        let mut a = list!(a = 42);
        let a_class = a.set_class(&["class"]).unwrap();

        unsafe {
            Rf_PrintValue(a.get());
            Rf_PrintValue(a_class.get());

            /* R_compute_identical:  C version of identical() function
            The third arg to R_compute_identical() consists of bitmapped flags for non-default options:
            currently the first 4 and the 6th default to TRUE, so the flag is set for FALSE values:
            1 = !NUM_EQ
            2 = !SINGLE_NA
            4 = !ATTR_AS_SET
            8 = !IGNORE_BYTECODE
            16 = !IGNORE_ENV
            32 = !IGNORE_SRCREF
            Default from R's default: 16 = (0 + 0 + 0 + 0 + 16 + 0)
            */
            assert!(R_compute_identical(a.get(), a_class.get(), 0) == Rboolean_TRUE);
            // assert!(R_compute_identical(a.get(), a_class.get(), 1) == Rboolean_TRUE);
            // assert!(R_compute_identical(a.get(), a_class.get(), 2) == Rboolean_TRUE);
            // assert!(R_compute_identical(a.get(), a_class.get(), 4) == Rboolean_TRUE);
            // assert!(R_compute_identical(a.get(), a_class.get(), 8) == Rboolean_TRUE);
            // // R default flag is 16
            // assert!(R_compute_identical(a.get(), a_class.get(), 16) == Rboolean_TRUE);
            // assert!(R_compute_identical(a.get(), a_class.get(), 32) == Rboolean_TRUE);
        }
    })
}
```

Output:

```
running 1 test
$a
[1] 42

attr(,"class")
[1] "class"
$a
[1] 42

attr(,"class")
[1] "class"
test test_what_is_returned_from_set_class ... ok
```

### DIFF
--- a/extendr-api/src/robj/mod.rs
+++ b/extendr-api/src/robj/mod.rs
@@ -802,7 +802,7 @@ make_typed_slice!(Rcomplex, COMPLEX, CPLXSXP);
 /// These are helper functions which give access to common properties of R objects.
 #[allow(non_snake_case)]
 pub trait Attributes: Types + Length {
-    /// Get a specific attribute as a borrowed robj if it exists.
+    /// Get a specific attribute as a borrowed `Robj` if it exists.
     /// ```
     /// use extendr_api::prelude::*;
     /// test! {
@@ -821,6 +821,7 @@ pub trait Attributes: Types + Length {
         if self.sexptype() == CHARSXP {
             None
         } else {
+            // FIXME: this attribute does not need protection
             let res = unsafe { Robj::from_sexp(Rf_getAttrib(self.get(), name.get())) };
             if res.is_null() {
                 None
@@ -844,7 +845,7 @@ pub trait Attributes: Types + Length {
         }
     }
 
-    /// Set a specific attribute and return the object.
+    /// Set a specific attribute in-place and return the object.
     ///
     /// Note that some combinations of attributes are illegal and this will
     /// return an error.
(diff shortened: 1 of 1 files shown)


Ilia-Kosenkov:
Can you call `a.set_class()` and then observe this class in the original `a`? If yes, maybe we should return nothing from `set_class()`?

CGMossa:
> Can you call `a.set_class()` and then observe this class in the original `a`? If yes, maybe we should return nothing from `set_class()`?

The answer is yes! See the output above...

We are to return something.. I'd honestly return `&mut self` for like chaining reason.

Ilia-Kosenkov:
It is a sketchy approach because it is Rust. We could have consumed it completely and return value, but I suspect it does not work with referecnes.

CGMossa:
Alright. I want to just merge this as is. I don't want to introduce a potential breaking change just yet. It isn't worth it. Clarity was needed so here it is.

Ilia-Kosenkov (approved):

Issue #659, opened 2023-11-01
Should `extendr-api` be thread safe?


CGMossa:
Dear all, I think it is time, that we discussed this topic at length. 

Should `extendr-api` be thread safe? Meaning, should it be possible to use `Robj` in threads?

1. ExtendR should not support multithreaded use, as R is single-threaded anyway, and in order to
use parallelism, we have to stick to accessing the R-API in a single-threaded way, which is currently accomplished
via `single_threaded`-guard everywhere that functions from `libR-sys` are invoked, that seemingly allocates or mutates.
2. Make `extendr-api` completely thread-safe.
3. Choose option (1), but provide tools for people to accomplish multithreaded use by hand.

There are many reasons that we are discussing this. `single-threaded` was developed due to issues with testing, but
it does provide guarantees that translate to multithreaded code.
The maintenance burden is growing ever more, as we constantly have issue with multiple threads, etc.
The current implementation of `single-threaded` has a sleep in it, thus we are essentially slowing down even 1-thread usage of ExtendR.

_My personal opinion_ on this matter is twofold: Currently, we have `single-threaded` in our API, and thus I believe we should
at least make sure that this works first. I would like to consider the option of not promising multithreaded in extendr, but give tools
for users to write these themselves, mainly a functioning `single_threaded`-guard, and a wrapper for `Robj` that is `Send` (and maybe `Sync`?). As of right now, I believe that we're fairly close to the first goal. The second goal is being discussed in this thread.

Therefore, no matter what we decide here, we have to discuss, review and merge the following:

## Previous discussions

This issue is mainly raised due to https://github.com/extendr/extendr/issues/583#issuecomment-1752184772,
however after carefully trying to deal with this, I do clearly see the ever growing maintenance burden due to this.
It isn't fun. It isn't possible to spend `n` hours and solve any given issue. It is always many people required to even
put a dent in the issues that arise from this.

Another commentary from me is here: https://github.com/extendr/extendr/issues/583#issuecomment-1752697578.
This references having embedded R in extendr as well, that is also a contentious issue, that need similar discussions like this one.
But @yutannihilation have raised this in #583. 

### Tracking issue for `extendr` with multithreaded

- [ ] #635 that is now superseded by #658. I have removed the "thread sleep" from the implementation,
and I have had an external Rust expert review the PR (see #635).
- [ ] Make a PR where `Robj` implements `Send`, as currently it does **not**, but we have already made that
promise that it should if it is `Send`.
[...]

PR #666, opened 2023-11-05
Added unsafe `SendSEXP` to `ownership`-module

CGMossa:
Currently, the `Ownership` module converts all `SEXP` to `usize` (?), such that they become `Send`, so that the ownership module can be used from multiple threads.

- Wrap `SEXP` in an explicit `Send` type. Accessing the pointer would require safety comments, as it can only be accessed through `inner` and `set_inner`. 
- No pointer conversion to `usize` and back.
- Prototype for an eventual support for an unsafe `SendRobj`. 

## Background

This came up when looking at why tests started to fail sporadically. This isn't meant to fix anything in particular,
but it adds an explicit unsafe layer that is involved with what the `ownership`-module currently does. If we get rid of multithreaded support, we'd still provide users with something similar to this, or they would write something themselves, see #659 for discussion.

### DIFF
--- a/extendr-api/src/ownership.rs
+++ b/extendr-api/src/ownership.rs
@@ -110,7 +110,6 @@ impl Ownership {
 
     // Garbage collect the tracking structures.
     unsafe fn garbage_collect(&mut self) {
-        // println!("garbage_collect {} {}", self.cur_index, self.max_index);
         let new_size = self.cur_index * 2 + EXTRA_PRESERVATION_SIZE;
         let new_sexp = Rf_allocVector(VECSXP, new_size as R_xlen_t);
         R_PreserveObject(new_sexp);
@@ -135,7 +134,7 @@ impl Ownership {
         }
 
         R_ReleaseObject(old_sexp);
-        self.preservation.set_inner(new_sexp);
+        self.preservation = (new_sexp).into();
         self.cur_index = j;
         self.max_index = new_size;
         self.objects = new_objects;
@@ -188,7 +187,7 @@ impl Ownership {
     }
 
     pub unsafe fn unprotect(&mut self, sexp: SEXP) {
-        let sexp_usize = sexp.into();
+        let send_sexp = sexp.into();
         let Ownership {
             preservation,
             cur_index: _,
@@ -196,7 +195,7 @@ impl Ownership {
             ref mut objects,
         } = self;
 
-        let mut entry = objects.entry(sexp_usize);
+        let mut entry = objects.entry(send_sexp);
         match entry {
             Entry::Occupied(ref mut occupied) => {
                 let object = occupied.get_mut();
@@ -228,8 +227,7 @@ impl Ownership {
             ref mut objects,
         } = *self;
 
-        let sexp_usize = sexp.into();
-        let mut entry = objects.entry(sexp_usize);
+        let mut entry = objects.entry(sexp.into());
         match entry {
             Entry::Occupied(ref mut occupied) => occupied.get().refcount,
             Entry::Vacant(_) => 0,
--- a/extendr-api/src/ownership.rs
+++ b/extendr-api/src/ownership.rs
@@ -17,6 +17,52 @@ use libR_sys::{
     Rf_unprotect, LENGTH, SET_VECTOR_ELT, SEXP, VECSXP, VECTOR_ELT,
 };
 
+mod send_sexp {
+    //! Provide a wrapper around R's pointer type `SEXP` that is `Send`.
+    //!
+    //! This can lead to soundness issues, therefore accessing the `SEXP` has
+    //! to happen through the unsafe methods [`SendSEXP::inner`] and [`SendSEXP::set_inner`].
+    //!
+    use libR_sys::SEXP;
+
+    #[derive(Debug, Clone, PartialEq, Eq, Hash)]
+    pub struct SendSEXP(SEXP);
+
+    impl From<SEXP> for SendSEXP {
+        fn from(value: SEXP) -> Self {
+            Self(value)
+        }
+    }
+
+    unsafe impl Send for SendSEXP {}
+
+    impl SendSEXP {
+        pub unsafe fn inner(&self) -> SEXP {
+            self.0
+        }
+
+        pub unsafe fn set_inner(&mut self, value: SEXP) {
+            self.0 = value;
+        }
+    }
+
+    // impl std::ops::Deref for SendSEXP {
+    //     type Target = SEXP;
+
+    //     fn deref(&self) -> &Self::Target {
+    //         &self.0
+    //     }
+    // }
+
+    // impl std::ops::DerefMut for SendSEXP {
+    //     fn deref_mut(&mut self) -> &mut Self::Target {
+    //         &mut self.0
+    //     }
+    // }
+}
+
+use self::send_sexp::SendSEXP;
+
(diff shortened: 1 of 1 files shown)


sorhawell:
I'm not familiar with the ownership.rs module.

As I read this PR, it might be compiled down to the very same binary.
I think the aim of this PR makes sense to not hide pointers in plain usize, whenever possible.

CGMossa:
> I'm not familiar with the ownership.rs module.
> 
> As I read this PR, it might be compiled down to the very same binary before and after. I think the aim of this PR makes sense to not hide pointers in plain usize, whenever possible.

Thanks for your comment. It is indeed my intention not to convert pointers to `usize` and back again. I can't find a reason why this is definitively bad, but it is unnecessary and the intent is unclear.

CGMossa:
I've rebased and removed commented code.

Ilia-Kosenkov:
This looks scary
>  Error: Error: SET_STRING_ELT() can only be applied to a 'character vector', not a 'integer'
>  *** longjmp causes uninitialized stack frame ***: terminated

CGMossa:
> This looks scary
> 
> > Error: Error: SET_STRING_ELT() can only be applied to a 'character vector', not a 'integer'
> > *** longjmp causes uninitialized stack frame ***: terminated

Indeed. It is caused by the multithreaded issues we've been having. Restarting CI, will give us a success, by the luck of the draw. We can discuss this particular issue in #674, or other places.

extendr-api/src/ownership.rs
```diff
@@ -58,22 +100,60 @@ impl Ownership {
...
+        self.objects = new_objects;
+    }
+
     unsafe fn protect(&mut self, sexp: SEXP) {
+        // This protects the SEXP. Is this necessary?
+        // Because the Ownership object already protects an SEXP in the `preservation` field.
+        // The new `sexp` is inserted into the preservation list via `SET_VECTOR_ELT` below.
+        // If list is protected then so are all of its elements.
+        // "Protecting an R object automatically protects all the R objects pointed to in the corresponding SEXPREC, for example all elements of a protected list are automatically protected." 5.9.1
         Rf_protect(sexp);
```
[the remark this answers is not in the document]
CGMossa: Exactly. But as it enters `fn protect` is is not protected, and we might incur some dubious operations, when we want to expand the preservation struct.
I'm going to resolve it for now, as it is fine as is.

JosiahParry (approved):
I tested this PR on {rsgeo} it builds just fine and works as anticipated. 

 R CMD check output: 
image

`Cargo.lock` illustrating it is using this branch. 

image

I feel confident in merging this. But since this is my first time getting familiar with the ownership module I'd like to ensure another set of eyes approve it. Would either @Ilia-Kosenkov or @sorhawell mind giving this another approval?

JosiahParry:
@Ilia-Kosenkov do you have some time this week to review this PR?

Ilia-Kosenkov:
I saw it, I will check it this weekend.

extendr-api/src/ownership.rs
```diff
@@ -58,22 +100,63 @@ impl Ownership {
...
                 cur_index: 0,
                 max_index: INITIAL_PRESERVATION_SIZE,
                 objects: HashMap::with_capacity(INITIAL_PRESERVATION_SIZE),
             }
         }
     }
 
+    // Garbage collect the tracking structures.
+    unsafe fn garbage_collect(&mut self) {
+        // println!("garbage_collect {} {}", self.cur_index, self.max_index);
```
Ilia-Kosenkov, line 113: Remove commented out code

extendr-api/src/ownership.rs
```diff
@@ -58,22 +100,63 @@ impl Ownership {
...
+                        refcount: object.refcount,
+                        index: j,
+                    },
+                );
+                j += 1;
+            }
+        }
+
+        R_ReleaseObject(old_sexp);
+        self.preservation.set_inner(new_sexp);
```
Ilia-Kosenkov, line 138: I would suggest to instead create new `SendSEXP` and assign it to `.preservation`. This will allow discarding `set_inner()` function in favor of some constructor.

extendr-api/src/ownership.rs
```diff
@@ -105,13 +188,13 @@ impl Ownership {
     }
 
     pub unsafe fn unprotect(&mut self, sexp: SEXP) {
-        let sexp_usize = sexp as usize;
+        let sexp_usize = sexp.into();
```
Ilia-Kosenkov, line 191: This is no longer `sexp_usize`

extendr-api/src/ownership.rs
```diff
@@ -145,52 +228,18 @@ impl Ownership {
             ref mut objects,
         } = *self;
 
-        let sexp_usize = sexp as usize;
+        let sexp_usize = sexp.into();
```
Ilia-Kosenkov, line 231: I think you can delay this, and just use `.entry(sexp.into())`, no?

CGMossa: Oh.. can you explain a little bit?

extendr-api/src/ownership.rs
```diff
@@ -58,22 +100,62 @@ impl Ownership {
...
+        // > pointed to in the corresponding SEXPREC, for example all elements
+        // > of a protected list are automatically protected." 5.9.1
         Rf_protect(sexp);
 
         if self.cur_index == self.max_index {
             self.garbage_collect();
         }
 
-        let sexp_usize = sexp as usize;
+        let sexp_usize = sexp.into();
```
Ilia-Kosenkov, line 158: Not a `sexp_usize`

extendr-api/src/ownership.rs
```diff
@@ -17,6 +17,43 @@ use libR_sys::{
...
+    unsafe impl Send for SendSEXP {}
+
+    impl SendSEXP {
+        /// Get the inner `SEXP`
+        pub unsafe fn inner(&self) -> SEXP {
+            self.0
+        }
+
+        /// Set the inner `SEXP` as another `SEXP`
+        pub unsafe fn set_inner(&mut self, value: SEXP) {
```
Ilia-Kosenkov, line 49: This is not used I think.

extendr-api/src/ownership.rs
```diff
@@ -17,6 +17,43 @@ use libR_sys::{
     Rf_unprotect, LENGTH, SET_VECTOR_ELT, SEXP, VECSXP, VECTOR_ELT,
 };
 
+mod send_sexp {
+    //! Provide a wrapper around R's pointer type `SEXP` that is `Send`.
+    //!
+    //! This can lead to soundness issues, therefore accessing the `SEXP` has
+    //! to happen through the unsafe methods [`SendSEXP::inner`] and [`SendSEXP::set_inner`].
```
Ilia-Kosenkov, line 24: Update docs since `set_inner()` is not needed.

extendr-api/src/ownership.rs
Ilia-Kosenkov: I meant do not declare separate variable and inline the cast. But it is unimportant.

CGMossa:
@Ilia-Kosenkov where is your comment on `.entry`? I still don't know what to do there :(

Ilia-Kosenkov:
Disregard my comment on`.entry`, I was too meticulous :)

JosiahParry:
All comments have been resolved, green checks abound. I'm going to merge! Thanks for your patience @CGMossa

PR #668, opened 2023-11-05
Don't protect R's `NULL`

CGMossa:
Here, we explicitly choose not to protect `R_NilValue`. This is to deal with #667, but also
it will relieve the pressure on the protection-list.

It is possible in a follow-up PR, to include all the symbols, and environment-values that R provides,
in this list as well; For now, let's stick with `R_NilValue`.

See discussion in #667. We might still want to merge #667, because the assumption is that the ownership module,
are testing a `static`, and that will inevitably be interfered with, with other tests running concurrently.

### DIFF
--- a/extendr-api/src/robj/mod.rs
+++ b/extendr-api/src/robj/mod.rs
@@ -39,6 +39,10 @@ pub use operators::Operators;
 pub use operators::*;
 pub use rinternals::Rinternals;
 
+/// These are permanent r-objects, thus they do not need protection.
+const PERMANENT: once_cell::sync::Lazy<[SEXP; 1]> =
+    once_cell::sync::Lazy::new(|| unsafe { [R_NilValue] });
+
 /// Wrapper for an R S-expression pointer (SEXP).
 ///
 /// Create R objects from rust types and iterators:
@@ -221,7 +225,9 @@ impl Length for Robj {}
 impl Robj {
     pub fn from_sexp(sexp: SEXP) -> Self {
         single_threaded(|| {
-            unsafe { ownership::protect(sexp) };
+            if !PERMANENT.contains(&sexp) {
+                unsafe { ownership::protect(sexp) };
+            }
             Robj { inner: sexp }
         })
     }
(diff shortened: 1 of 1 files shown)


extendr-api/src/robj/mod.rs
```diff
@@ -221,7 +225,9 @@ impl Length for Robj {}
 impl Robj {
     pub fn from_sexp(sexp: SEXP) -> Self {
         single_threaded(|| {
-            unsafe { ownership::protect(sexp) };
+            if !PERMANENT.contains(&sexp) {
+                unsafe { ownership::protect(sexp) };
+            }
```
Ilia-Kosenkov, line 230: This looks like a significant performance hit, performing a lookup every time we construct something from SEXP.
Alternatively, shouldn't this be part of `ownership::protect()`?

CGMossa: It is still a performance hit if I put it in `ownership::protect`.
I don't know, I'm exploring options and trying to make my discoveries clear. 🤷

Ilia-Kosenkov: I mean, this logic is part of the protection mechanism, not of the `Robj` creation.

yutannihilation: I agree with Ilia. cpp11 also does the `NULL`-check on the protection mechanism's side.

https://github.com/r-lib/cpp11/blob/500f642b4ea132ec8c168fc70a28e81e9510ece3/inst/include/cpp11/protect.hpp#L252-L254

Regarding the performance, I'm not confident, but I guess this won't matter much. The protection mechanism anyway need to check the entry in the `HashMap` if not rejecting `NULL`, and I don't think it's significantly slower or faster than comparing the pointer addresses. I might be wrong.

Ilia-Kosenkov: In the limit, there is definitely a moment when comparing 4-5 pointers becomes as slow as looking up a value in a hash map of 4-5 elements, but for now we can ignore that.

PR #672, opened 2023-11-12
Don't protect permanents

CGMossa:
I extracted all symbols that `libR-sys` provides access to, and set them so they'll never be protected.

I want to be able to work with `SEXP`'s directly, so I've made a version of `libR-sys` that wraps `SEXP`, and asserts
that they are `Send` and `Sync`. 

Depends on

- [ ] https://github.com/extendr/libR-sys/pull/201

### DIFF
--- a/extendr-api/src/ownership.rs
+++ b/extendr-api/src/ownership.rs
@@ -1,12 +1,12 @@
 //! Maintain ownership of R objects.
 //!
-//! This provides the functions protect() and unprotect()
+//! This provides the functions [protect] and [unprotect].
 //! A single preserved vector holds ownership of all protected objects.
 //!
 //! Objects are reference counted, so multiple calls are possible,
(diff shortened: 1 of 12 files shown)


extendr-api/src/ownership.rs
```diff
@@ -17,30 +17,110 @@ use libR_sys::{
...
     refcount: usize,
     index: usize,
 }
 
 // A reference counted object with an index in the preservation vector.
+#[derive(Debug)]
 struct Ownership {
     // A growable vector containing all owned objects.
-    preservation: usize,
+    preservation: SEXP,
```
Ilia-Kosenkov, line 123: Wait, were we storing `SEXP` as `usize`?

CGMossa: Yes

PR #674, opened 2023-11-12
Make extendr not thread-safe

CGMossa:
- [x] Remove all usage of `single_threaded`
- [x] Set `--test-threads=1` always
- [ ] No longer rely on `ownership` tracking across threads, i.e. use `thread_local!` for `Ownership`.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,4 @@
+[env]
+RUST_TEST_THREADS = "1"
 [alias]
 extendr = "run --package xtask --"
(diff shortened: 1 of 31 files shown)


CGMossa:
My issue on Windows, is that the CI already inserts an `[env]` in the `.cargo/config.toml`.

CGMossa:
Dear @yutannihilation. This isn't ready for final review yet, but I do want your guidance on this PR.

Basically, I got fed up chasing my own tail, and really embedding unnecessary bottlenecks to extendr, without
really paying it off properly. So I thought a branch where we see how it would look without upfront promise
of thread-safety.

CGMossa:
CI in this PR is failing due to new r-devel warning on r-cmd-check that is will be addressed here #809.

PR #675, opened 2023-11-13
Remove `MSYS2` from CI

CGMossa:
We don't need `MSYS2` anymore, so I've removed it from CI.


yutannihilation (approved):
I couldn't find the history when this installation of clang was added, but I think it's fine to remove this, considering the image is shipped with LLVM today.

https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md#language-and-runtime

PR #676, opened 2023-11-13
Bump `extendrtests` version

CGMossa:
- Set version equal to `extendr`
- Mark not to publish.
- Set the `resolver` as that isn't
  done for workspaces.

### DIFF
--- a/tests/extendrtests/src/rust/Cargo.toml
+++ b/tests/extendrtests/src/rust/Cargo.toml
@@ -1,14 +1,16 @@
 [workspace]
+resolver = "2"
 
 [package]
 name = "extendrtests"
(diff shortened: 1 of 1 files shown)


yutannihilation (approved):

yutannihilation:
Btw, why does the version need to be bumped? Isn't it fine to leave it some random version like 0.0.0?

CGMossa:
No reason. I was having some very strange behavior happening all over the place, so I tried everything.
If these other two properties didn't need to be added (in my opinion), then I wouldn't have bumped the version honestly.

yutannihilation:
I hope it would work just by clearing the cache, but thanks for the explanation. I see.

PR #680, opened 2023-12-12
Add user-documentation to S4

CGMossa:
Explain what `set_class` is, and give guidance to r-package authors.

### DIFF
--- a/extendr-api/src/wrapper/s4.rs
+++ b/extendr-api/src/wrapper/s4.rs
@@ -1,4 +1,16 @@
 //! S4 class support.
+//! 
+//! It is not possible to create an S4 class from R's C-API, and thus it is
+//! not possible to do so in Rust. But an S4 class can be instantiated.
+//! 
+//! Thus, the S4 class definition must be evaluated prior to using [`S4::new`].
+//! Conveniently, to inline the defintion of an S4 class with R, one can
+//! use [`S4::set_class`].
+//! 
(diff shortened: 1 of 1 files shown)


JosiahParry:
Can we add an example here? Perhaps recreate something super simple like:
```r
person_class <- setClass(
  "person",
  slots = c(name = "character", age = "integer")
)

person_class(name = "Lubo", age = 74L)
#> An object of class "person"
#> Slot "name":
#> [1] "Lubo"
#> 
#> Slot "age":
#> [1] 74
```

<sup>Created on 2023-12-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

JosiahParry:
This looks great. @CGMossa do we need to include an integer for Fred in the example since the s4 class requires it?

JosiahParry:
Ah, I think the doc examples are what are tripping me up. I think that they can be simplified. For example something like this would be handy to illustrate it. 

I want to give this more thorough docs. But I think the addition alone is sufficient. 

```r
library(rextendr)

rust_function(r"{
fn tst() -> S4 {
    // instantiate the s4 class 
    let class = S4::set_class("fred", pairlist!(x="numeric"), r!(())).unwrap();
    let mut fred = S4::new("fred").unwrap();
    fred.set_slot("x", 0.1);
    fred
}
}")
```

JosiahParry (approved):

PR #683, opened 2024-01-11
Remove `Debug` from `ExternalPtr`

CGMossa:
Removed `'static`', `Debug` from constraints on `T`, as it seems
very unnecessary.

The advantage here is that `ExternalPtr` can be used in `no_std` contexts.


JosiahParry:
I had a misunderstanding with this. Debug does work with no_std environments but is often removed as a requirement for no_std environments. I don't think theres a usecase for no_std environments and R?

PR #685, opened 2024-01-14
fix: should not use `@docType "package"`

### LINKED ISSUE #684: `@docType "package"` is deprecated in roxygen2 7.3.0
https://www.tidyverse.org/blog/2024/01/roxygen2-7-3-0/

eitsupi:
Fix #684

The existence of this makes it impossible to use `"_PACKAGE"` in packages where extendr-wrapper.R is located, like pola-rs/r-polars#705

### DIFF
--- a/extendr-api/src/metadata.rs
+++ b/extendr-api/src/metadata.rs
@@ -352,7 +352,6 @@ impl Metadata {
         )?;
 
         if use_symbols {
-            writeln!(w, "#' @docType package")?;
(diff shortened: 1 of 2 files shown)


JosiahParry:
Just tested removing `@docType "packagename"` and `"_PACKAGE"` from my R package and there is no change in behavior. It seems that this is extra and not needed.

I don't see any downsides?

CGMossa (approved):
I defer to you on this. Better merge and move on.

Issue #68, opened 2020-12-10
Release version 0.2.0


clauswilke:
I'd like to release a version 0.2.0 that provides a minimally viable product on all platforms. Below is the checklist for the release. If I'm forgetting anything, please let me know in the comments.

* [x] Address all issues/PRs in the 0.2.0 milestone
* [x] Finalize and release libR-sys 0.2.0 (https://github.com/extendr/libR-sys/issues/20)
* [x] ~~Fix `bump.sh` and `publish.sh` (#70)~~ (see #82)
* [x] Bump libR-sys dependency to latest version
* [x] Remove any patch_crates.io statements from top-level `Cargo.toml`
* [x] Bump version to 0.2.0
* [x] Bump version number in changelog 
* [x] ~~Verify that all crates in the workspace have the same version and the same extendr version dependencies~~ (see #82)
* [x] Tag release on github
* [x] Release to crates.io

clauswilke:
@andy-thomason Given the amount of changes we're making right now, and considering that the resulting crate will likely be the first minimally usable product, should the next release be 0.2.0?

andy-thomason:
I would like to get the updated docs and cleaner API into the release.

There are some minor breaking changes as I've dropped rust type names in favour
of "integer", "logical", "numeric", "character" where possible as R users will
be more familiar with these.

The docs are looking much better now, although not visible in docs.rs until we publish.

clauswilke:
Agreed. In addition, I'd like to merge #93, since it allows us to use the module macro on Windows.

clauswilke:
@andy-thomason I have gone over the open issues and added relevant ones to the 0.2 milestone:
https://github.com/extendr/extendr/issues?q=is%3Aopen+is%3Aissue+milestone%3A0.2.0
Could you add any issues I haven't added that you think we should address before a release? In this way we can keep track of where we are.

Also, I think for 0.2.0 we should focus on API design, overall architecture, and basic functionality, since it'll probably be around for a while. We can add documentation and convenience functions later.

andy-thomason:
I think we are good to go on 0.2.0

Feel free to bump and publish, or I will go ahead on Monday.

clauswilke:
Ok, I've made a release. Let's hope it's all correct. Currently testing with rextendr.

clauswilke:
Things seem to be working fine.

andy-thomason:
Tada!

PR #689, opened 2024-01-17
Clarify .inherits() behavior

JosiahParry:
This PR improves the documentation for the `.inherits()` method to clarify that the method does not support implicit classes.

Closes https://github.com/extendr/extendr/issues/68


JosiahParry:
CI is passing for this now.

JosiahParry:
That's fine with me as well. The current behavior isn't terrible at all. It is just the weird edge case of implicit classes i've encountered.

PR #690, opened 2024-01-17
Use public RSPM in both setup-r workflows

JosiahParry:
This PR changes the use-public-rspm flag to true where it was previously false. 

RSPM will serve binaries when available. Otherwise it will default to source installations. 

Was there a reason why this was set false?


JosiahParry:
CI now passing

CGMossa (approved):

PR #699, opened 2024-02-03
Fix a likely regression in new `pwsh` version

Ilia-Kosenkov:
Partial conditions (when `else` clause is dropped) passed `''` among parameters instead of nothing, which subsequently led to broken invocations of `cargo`.


CGMossa (approved):
This needs to be pushed to libR sys also.

yutannihilation:
> This needs to be pushed to libR sys also.

Here it is.

https://github.com/extendr/libR-sys/pull/217

PR #706, opened 2024-03-03
Introduce optional faer support 

0urobor0s:
This PR introduces optional support for `faer::Mat<f64>` into and from a `Robj` in order to integrate with the [faer](https://github.com/sarah-ek/faer-rs) crate.


CGMossa:
Alright! I have been waiting for this to happen!

`faer` is a linear algebra library written in Rust, with excellent performance, comparable to BLAS / LAPACK / etc. See https://github.com/sarah-ek/faer-rs
It has been on-going for years now. There are regular releases, and many use-ful algorithms has been added.

This could be to extendr, what Armadillo is to Rcpp. Although the focus is slightly more scoped.

Feedback: I've since we spoke about this, learned a few things about our code-base. We are meant to provide both FromRobj and TryFrom. The "main" implementation should be on `&Robj`, and then for `TryFrom<Robj>` should use the `&Robj`-implementation.

We also have a section for optional crate integrations. if it is possible to move it to there, and then have the `cfg(feature = "faer")` attached to the module itself, that would be great. 

faer implements sparse matrices, that is implemented in R by the `{Matrix}` package. Making PR that helps with these conversions could be a good next step.

Also, I think we need a test or two, but we can defer that to a bit later. 


I'm very happy that you've embarked on this PR.

extendr-api/src/robj/from_robj.rs
```diff
@@ -289,3 +289,59 @@ where
...
+                if dim.len() != 2 {
+                    return Err("could not convert to matrix");
+                }
+
+                if let Some(slice) = robj.as_real_slice() {
+                    let fmat = faer::mat::from_column_major_slice::<f64>(&slice, dim[0], dim[1]);
+                    Ok(fmat.to_owned())
+                } else if let Some(slice) = robj.as_integer_slice() {
+                    let fmat = faer::Mat::<f64>::from_fn(dim[0], dim[1], |i, j| {
+                        slice[i + j * dim[0]] as f64
```
jonlachmann, line 310: Does this imply a copy?

extendr-api/src/robj/into_robj.rs
```diff
@@ -710,6 +710,16 @@ impl From<Vec<Rstr>> for Robj {
...
+#[cfg(feature = "faer")]
+impl From<faer::Mat<f64>> for Robj {
+    /// Convert a faer Mat<f64> into Robj.
+    fn from(mat: faer::Mat<f64>) -> Self {
+        mat.col_chunks(1)
+            .flat_map(|c| c.row_chunks(1).map(|r| r.read(0, 0)))
+            .collect_robj()
+    }
+}
+
```
JosiahParry, line 722: Should we also have an `impl From<RMAtrix> for faer::Mat<F64>>`?

extendr-api/src/robj/from_robj.rs
0urobor0s: From my understanding of [faer::mat::from_column_major_slice](https://docs.rs/faer/0.17.1/faer/mat/fn.from_column_major_slice.html), it uses the same values. So the current flow uses that if the values are already f64 and if not, does copy with the cast to f64.

extendr-api/src/robj/into_robj.rs
0urobor0s: I think this question is more of the API design and what to support, so I'm not sure how to answer. Currently I think this provides the most flexibility and it's easy to go from an `Robj` to `RMAtrix`, or even make the conversion on the R side.

extendr-api/src/robj/from_robj.rs
```diff
@@ -289,3 +289,59 @@ where
         }
     }
 }
+
+#[cfg(feature = "faer")]
+impl<'a> FromRobj<'a> for faer::Mat<f64> {
+    fn from_robj(robj: &'a Robj) -> std::result::Result<Self, &'static str> {
+        use crate::scalar::Scalar;
+        if robj.is_matrix() {
```
JosiahParry, line 297: its worth noting that an R matrix is not always numeric. It can be character as well

extendr-api/src/robj/from_robj.rs
jonlachmann: Okay, would the .to_owned() just copy the reference to the structure? Sorry if this is a very basic question as rust is quite new to me.

extendr-api/src/robj/into_robj.rs
```diff
@@ -710,6 +710,16 @@ impl From<Vec<Rstr>> for Robj {
     }
 }
 
+#[cfg(feature = "faer")]
+impl From<faer::Mat<f64>> for Robj {
+    /// Convert a faer Mat<f64> into Robj.
+    fn from(mat: faer::Mat<f64>) -> Self {
+        mat.col_chunks(1)
+            .flat_map(|c| c.row_chunks(1).map(|r| r.read(0, 0)))
+            .collect_robj()
```
0urobor0s, line 719: I'm not sure if this is the best/fastest way to iterate, but it works and can always be revisited later.

0urobor0s:
> Feedback: I've since we spoke about this, learned a few things about our code-base. We are meant to provide both FromRobj and TryFrom. The "main" implementation should be on `&Robj`, and then for `TryFrom<Robj>` should use the `&Robj`-implementation.

Not sure I understood this, should the traits be implemented for `&Robj` instead of `Robj`? If the implementation doesn't take ownership of the `Robj` shouldn't it always have to copy the values?

> faer implements sparse matrices, that is implemented in R by the `{Matrix}` package. Making PR that helps with these conversions could be a good next step.

Currently I haven't explored sparse matrix from R's side, so I would say for this PR, only Dense would be provided.

> By _test_ I meant we might want this https://github.com/extendr/extendr/blob/master/tests/extendrtests/tests/testthat/test-ndarray.R but for `test-faer.R`.

Not sure I understood the tests, are these tests "integration"-like tests? It seems they were testing the underlying library instead and the code indirectly more than the code itself. In theory shouldn't the tests be constrained to the affected structs and the upper abstractions already be tested?

JosiahParry:
This is exciting! I'm not familiar with `faer` but there's a few things to address and document from my perspective: 

## Supported Types 

In R, every vector type can be stored in a matrix but not every type can likely be converted in to a faer `Mat`. 

I believe the ones that _can be_ converted are:

- double
- integer
- factor
- complex

The ones that _cannot be_ converted are 

- character
- list

For example in R this is a valid list:

```r
matrix(
  list(R6::R6Class("tst"), 1L, letters, function(){}),
  ncol = 2
)
```

## Conversion

### From RMatrix

We can support conversion from RMatrix with `TryFrom` for supported types. These should fail (IIUC) if NA is present or the types are not supported (list, character)

### From faer

We can support From for faer::Mat<T> where T is i32, u32, f64, and Complex I believe.

### FromRobj

I think FromRobj should respect multiple types like described above and should not be prescribed to only f64.

0urobor0s:
> We also have a section for optional crate integrations. if it is possible to move it to there, and then have the `cfg(feature = "faer")` attached to the module itself, that would be great.

I think this is now addressed by commit https://github.com/extendr/extendr/pull/706/commits/1a8bdd52d994761306e5b49a8a9f02dbc8f99a32.

0urobor0s:
Regarding [this comment](https://github.com/extendr/extendr/pull/706#issuecomment-1975239242), [faer currently only supports](https://docs.rs/faer-core/0.17.1/faer_core/trait.Entity.html) `f32`, `f64`, `c32` and `c64`. The reason for only f64 being supported is because R doesn't support single precision, as for integer types, I don't even know if those should even be supported but chose to include them because LAPACK seems to also convert to double precision and wanted that interoperability, however for a library like `extendr` that is also an API design decision. As for complex types, I'm not familiar with their behaviour in other libraries so I cannot say. For factors, I don't know if they are a good fit for this kind of purpose.

extendr-api/src/robj/from_robj.rs
0urobor0s: Won't the `as_real_slice`/`as_integer_slice` in that case return None?

extendr-api/src/robj/from_robj.rs
0urobor0s: Ah, I see what you mean. I've asked and it does indeed copy, the alternative would be to return instead of `Mat` returning a `MatRef`. That however would not work for the integer branch since the returned value referencing `Mat` would be data owned by the current function. Maybe it would be worth it to also have a `impl<'a> FromRobj<'a> for MatRef<'a, f64>` that would only work for floats from R.

jonlachmann: That sounds like a great idea, being able to have a reference if the matrix is huge will be good both for performance and memory usage.

0urobor0s: I agree, addressed in commit https://github.com/extendr/extendr/pull/706/commits/6e14fe3078a4654e052fe1fd7a4a845aa48d44b5.

extendr-api/Cargo.toml
```diff
@@ -20,6 +20,7 @@ libc = { version = "0.2", optional = true }
 ndarray = { version = "0.15.3", optional = true }
 num-complex = { version = "0.4", optional = true }
 serde = { version = "1.0", features = ["derive"], optional = true }
+faer = { version = "0.18.0", optional = true }
```
JosiahParry, line 23: I'm worried that if we have a hard fix to a version we will make it incompatible with other libraries. I run into this issue a lot when I work with arrow

0urobor0s: I cannot say I have much experience with Rust libraries, but reading [Dependency Resolution](https://doc.rust-lang.org/cargo/reference/resolver.html#dependency-resolution) it would seem it's ok for this use case, as I don't think it's expected to depend directly on this, and a downstream project that uses `faer` directly would build the two different versions. `ndarray` also seems to be the same in the file.

CGMossa:
This is how we could add support for complex numbers: https://discord.com/channels/751419551132155925/751419551132155928/1214741810857902100

Answer provided by author of faer.

JosiahParry:
I've had a chance to test this locally. There are a couple of issues. There is a namespace conflict (present in ndarray feature too). So we need to change the `pub mod` to `mod` so we do not have a namespace conflict—this allows us to also import `faer` in our `Cargo.toml`

```rust
#[cfg(feature = "faer")]
mod faer;
```

Additionally, the conversion from `faer` to Robj does not return a matrix but rather a single vector:

```rust
#[extendr]
fn faer_mat(x: Robj) -> Robj {
    let m = Mat::<f64>::from_robj(&x).unwrap();
    m.into_robj()
}
```

The `dim` attribute needs to be added so that it gets turned into a matrix additionally, integer's `NA` are not respected. 

``` r
library(tst2)
m <- matrix(1:9, ncol = 3)
m[1,1] <- NA
m
#>      [,1] [,2] [,3]
#> [1,]   NA    4    7
#> [2,]    2    5    8
#> [3,]    3    6    9

faer_mat(m)
#> [1] -2147483648           2           3           4           5           6
#> [7]           7           8           9
```

There's also a segfault when I run 

```r
faer_mat(as.double(m))
```

0urobor0s:
The comments from https://github.com/extendr/extendr/pull/706#issuecomment-1984120974 are mostly address by https://github.com/extendr/extendr/pull/706/commits/fb24d037ad5e053aa00dd06448163d0ce411cf58, and in addition it also adds a way for `cargo extendr devtools-test` to compile the R test project, for better development experience. This addition is probably out of the scope of this PR or can introduce some issues, so maybe it's best to open it in a new PR, let me know.

> The `dim` attribute needs to be added so that it gets turned into a matrix additionally, integer's `NA` are not respected.

Should a `RArray`/`RMatrix` be returned when the trait is implemented for `Robj`? Wouldn't it be up to the caller to convert from `Robj` to the desired output? Or maybe a new trait implementation would be necessary for `RArray`/`RMatrix`. This could probably be done with something like:
```rust
mat.col_chunks(1).collect_rarray([mat.nrows(); 2])
```

> There's also a segfault when I run
> 
> ```r
> faer_mat(as.double(m))
> ```
> 

I think this is because of the `unwrap()` in your function definition, as the R `as_double()` function doesn't return a matrix, https://github.com/extendr/extendr/pull/706/commits/fb24d037ad5e053aa00dd06448163d0ce411cf58#diff-78d342e81d701ac77edd01fe11cab6da7cf941774e2ea660bfefc8076997a7da shows a possible way to check if the call was succeful or not and return an appropriate response. Does `extendr` have any implicit way of handling these cases?

xtask/src/commands/devtools_test.rs
```diff
@@ -14,6 +14,7 @@ pub(crate) fn run(shell: &Shell) -> Result<(), Box<dyn Error>> {
 
 fn run_tests(shell: &Shell) -> Result<(), Box<dyn Error>> {
     let _r_path = shell.push_dir(R_FOLDER_PATH);
+    cmd!(shell, "Rscript -e rextendr::document()").run()?;
```
CGMossa, line 17: I ask you kindly to remove this from this PR, and know that we will address this soon, in a separate PR.

extendr-api/src/optional/faer.rs
```diff
@@ -0,0 +1,275 @@
...
+            Err(Error::ExpectedMatrix(robj.clone()))
+        }
+    }
+}
+
+impl<'a> TryFrom<&'_ Robj> for MatRef<'a, f64> {
+    type Error = Error;
+
+    fn try_from(robj: &Robj) -> Result<Self> {
+        if robj.is_matrix() {
```
CGMossa, line 114: I've recently checked the behavior of `isMatrix` and it checks a lot of these things at once:
```c
INLINE_FUN Rboolean isMatrix(SEXP s)
{
    SEXP t;
    if (isVector(s)) {
	t = getAttrib(s, R_DimSymbol);
	/* You are not supposed to be able to assign a non-integer dim,
	   although this might be possible by misuse of ATTRIB. */
	if (TYPEOF(t) == INTSXP && LENGTH(t) == 2)
	    return TRUE;
    }
    return FALSE;
}
```

CGMossa (commented):
I don't think you've been able to see any of my comments. Let me know if you were able to. I'm clicking comment now to see if makes a change.

extendr-api/Cargo.toml
0urobor0s: I don't think it's safe to do so, as by following [SemVer](https://semver.org/#spec-item-4) it seems it could break at any time.

extendr-api/src/optional/faer.rs
0urobor0s: Nice, I think https://github.com/extendr/extendr/pull/706/commits/4995ebbdc699182840767eb3e98137f0d86f8f62 addresses the unnecessary check. That said would it be possible to change this line `let dim: Vec<_> = dim.iter().map(|d| d.inner() as usize).collect();`  and use the dimension directly? Maybe with [as_array_view](https://docs.rs/extendr-api/0.6.0/extendr_api/prelude/trait.Dimension.html#method.as_array_view) ?

CGMossa: I recently went through this, you can see it here https://github.com/CGMossa/extendr/blob/master/extendr-api/src/optional/ndarray.rs

CGMossa: Sorry that was the wrong reference..
```rs

impl<T> TryFrom<&Robj> for RMatrix<T>
where
    Robj: for<'a> AsTypedSlice<'a, T>,
{
    type Error = Error;

    fn try_from(robj: &Robj) -> Result<Self> {
        // isMatrix checks vector-type, presence of `dim`, and length(dim) == 2
        if !robj.is_matrix() {
            return Err(Error::ExpectedMatrix(robj.clone()));
        }
        let dim = robj.dim().unwrap();
        let dim = dim.as_slice();
        Ok(RArray::from_parts(
            robj.clone(),
            [dim[0].inner() as _, dim[1].inner() as _],
        ))
    }
}
```

CGMossa: Turns out, `.as_slice()` isn't in `master`.

JosiahParry:
I spent some time revisiting this PR today. I would quite like to see it merged, however, I think in its current state it just isn't quite there yet. 

The two big things for me are the fact that the conversion to Robj are done in the context of Robj and not some more specific struct. In the case of `faer`, I think it makes most sense to add conversion to `RMatrix<f64>` then delegate to RMatrix's conversion to Robj. 

The second thing is that the current implementation crashes R (i think due to dereferencing?). I've tried taking a stab at this in a branch off of your branch, @0urobor0s. Lines 1-55 are the main changes I've made https://github.com/extendr/extendr/blob/fd3848f892faa211a9f4a8529139df6931754aae/extendr-api/src/optional/faer.rs#L7-L55. 

I made some rough integration tests as well for this purpose. All of the below work well! 

I got a bit caught up with lifetimes and the borrow checker when trying to get faer to work with `&Robj` and `&RMatrix<f64>`. Perhaps you or @CGMossa know a bit more about that.

```rust

#[extendr]
fn mat_to_mat(x: Mat<f64>) -> Mat<f64> {
    x
}

#[extendr]
fn mat_to_rmat(x: Mat<f64>) -> RMatrix<f64> {
    RMatrix::<f64>::from(x)
}

#[extendr]
fn mat_to_robj(x: Mat<f64>) -> Robj {
    x.into_robj()
}

#[extendr]
fn mat_to_rmatfloat(x: Mat<f64>) -> RMatrix<Rfloat> {
    RMatrix::<Rfloat>::from(x)
}

// convert to Mat<f64> from other things
#[extendr]
fn rmat_to_mat(x: RMatrix<f64>) -> Mat<f64> {
    Mat::<f64>::from(x)
}

#[extendr]
fn robj_to_mat(x: Robj) -> Mat<f64> {
    Mat::<f64>::try_from(x).unwrap()
}

// MatRef input
fn matref_to_mat(x: MatRef<'_, f64>) -> Robj {
    RMatrix::<f64>::from(x).into()
}
```
```r
library(testthat)

m <- matrix(rnorm(100), ncol = 10)
expect_identical(m, mat_to_rmat(m))
expect_identical(m, mat_to_robj(m))
expect_identical(m, mat_to_rmatfloat(m))
expect_identical(m, mat_to_mat(m))
expect_identical(m, rmat_to_mat(m))
expect_identical(m, robj_to_mat(m))
```

0urobor0s:
> The two big things for me are the fact that the conversion to Robj are done in the context of Robj and not some more specific struct. In the case of `faer`, I think it makes most sense to add conversion to `RMatrix<f64>` then delegate to RMatrix's conversion to Robj.

Agree with this.

> The second thing is that the current implementation crashes R (i think due to dereferencing?). I've tried taking a stab at this in a branch off of your branch, @0urobor0s. Lines 1-55 are the main changes I've made

Can you point to a test that crashes the R session? Applied your changes in the PR with some changes in https://github.com/extendr/extendr/pull/706/commits/2338bfdc77e3e3e6b4e449a2373b5cdadc0105a1.




> I got a bit caught up with lifetimes and the borrow checker when trying to get faer to work with `&Robj` and `&RMatrix<f64>`. Perhaps you or @CGMossa know a bit more about that.

I think this mostly because `as_real_slice()` doesn't have the necessary lifetime notation for it to work, with the change to `as_typed_slice()` and the necessary lifetime notation it works.

extendr-api/src/optional/faer.rs
```diff
@@ -0,0 +1,253 @@
...
+        if let Some(slice) = robj.as_typed_slice() {
+            let fmat = mat::from_column_major_slice::<f64>(slice, nrows, ncols);
+            Ok(fmat)
+        } else {
+            Err(Error::ExpectedReal(robj.clone()))
+        }
+    }
+}
+
+impl FromRobj<'_> for Mat<f64> {
```
CGMossa, line 102: These work now! Maybe you can add a better error message to them? I just used the generic one..

tests/extendrtests/tests/testthat/test-faer.R
```diff
@@ -0,0 +1,34 @@
...
+
+library(testthat)
+
+m <- matrix(rnorm(100), ncol = 10)
+expect_identical(m, mat_to_rmat(m))
+expect_identical(m, mat_to_robj(m))
+expect_identical(m, mat_to_rmatfloat(m))
+expect_identical(m, mat_to_mat(m))
+expect_identical(m, rmat_to_mat(m))
+expect_identical(m, robj_to_mat(m))
```
CGMossa, line 34: You're missing `matref_to_mat` from this list. It is not being tested here.

0urobor0s:
Trying to do something like:
```rust

extendr-api/Cargo.toml
JosiahParry: I suppose this should be fine for now. If we encounter any issues we can always revisit it. Nothing like a little bit of pain to encourage fixing a codebase.

JosiahParry (approved):
Thank you, @0urobor0s for your incredible patience with this! 

This looks very good now! The one failing CI check is because of an issue with {stringi} package and should not stop this from being merged in. 

To recap the current limitations:

- No NA handling
- No integer or complex support

If we need these, we can come back to it at a later point. But this is great!

0urobor0s:
Thank you @JosiahParry and @CGMossa, without this wouldn't have been merged.

I think only NA handling and complex support weren't implemented, since for integers we added this https://github.com/extendr/extendr/pull/706/files#diff-2d778acb1a1c79323f1402e44372732093b5b6cd4d892b462f1a40535ead7f42R131-R140

PR #711, opened 2024-03-24
fix ci: r-devel now uses rtools44 by default

CGMossa:
Following in the steps of extendr/libR-sys#220. This is affecting CI right now.


JosiahParry (approved):
Test with rtools44 passes and looks great!

PR #712, opened 2024-03-24
ci: Upload results of r-cmd-check

CGMossa:
We continue to see sporadic errors on ci.

The logs from r-cmd-check are not uploaded, when it fails.

We use our own custom invocation of `r-cmd-check` due to path issues or something or another I've never understood why we even have to be bogged down by rstudio path-nonsense.

Anyways, upload scheme adapted from https://github.com/r-lib/actions/blob/v2/check-r-package/action.yaml

Let us see if it works.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -208,11 +208,11 @@ jobs:
         id: id_cargo_r_cmd_check
         run: |
           cargo extendr r-cmd-check --error-on ${{ steps.error-on.outputs.level }} --check-dir extendrtests_check
-      - name: Upload results if it failed
+      - name: Upload `check-dir` if r-cmd-check failed on `{extendrtests}`
         if: ${{ failure() }}
         uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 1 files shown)


JosiahParry:
The check.log file does not make its way into the uploads: 
image

CGMossa:
This PR uploads the `check-dir` folder. If we were to make the logs generated, be part of the logs in the workflow, we'll have to echo / cat them, and then they'd be part of those. I'm unsure if that's just overcomplicating things for no effort.

JosiahParry:
This PR works well. I'm able to download the folder for the failing check. Interestingly, the test fails but no errors are reported in any logs. R CMD check fails but all tests pass. I suspect the issue may be due to rtools44 or r-devel and _not_ extendr seeing as all other tests pass with flying colors. If you concur, I will approve. 

image

CGMossa:
Fully agree. Let's merge.

JosiahParry (approved):

Issue #716, opened 2024-03-30
`throw_r_error` may crash R session


eitsupi:
The polars package seems to occasionally crash due to the following location.
https://github.com/extendr/extendr/blob/fb88c799126ba93795b070db2bdfc94c4aeb0d6b/extendr-api/src/thread_safety.rs#L59

For example: pola-rs/r-polars#993

```r
> library(polars)
> as_polars_df(mtcars)$select(pl$col("mpg", "cyl")$to_struct()$map_elements(\(x) x$mpg))
[src/series.rs:387:17] x = Struct(
    [
        Field {
            name: "mpg",
            dtype: Float64,
        },
        Field {
            name: "cyl",
            dtype: Float64,
        },
    ],
)
thread '<unnamed>' panicked at src/series.rs:388:17:
not yet implemented: this input type is not implemented
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread '<unnamed>' panicked at src/series.rs:55:1:
explicit panic
thread '<unnamed>' panicked at /usr/local/cargo/git/checkouts/extendr-569c02e1a164cdf5/b26adfe/extendr-api/src/thread_safety.rs:59:49:
called `Result::unwrap()` on an `Err` value: NulError(36, [117, 115, 101, 114, 32, 102, 117, 110, 99, 116, 105, 111, 110, 32, 112, 97, 110, 105, 99, 107, 101, 100, 58, 32, 109, 97, 112, 95, 101, 108, 101, 109, 101, 110, 116, 115, 0])
thread '<unnamed>' panicked at src/rdataframe/mod.rs:87:1:
explicit panic
thread '<unnamed>' panicked at /usr/local/cargo/git/checkouts/extendr-569c02e1a164cdf5/b26adfe/extendr-api/src/thread_safety.rs:59:49:
called `Result::unwrap()` on an `Err` value: NulError(30, [117, 115, 101, 114, 32, 102, 117, 110, 99, 116, 105, 111, 110, 32, 112, 97, 110, 105, 99, 107, 101, 100, 58, 32, 115, 101, 108, 101, 99, 116, 0])
fatal runtime error: failed to initiate panic, error 5
fish: Job 1, 'R' terminated by signal SIGABRT (Abort)
```

CGMossa:
Found it.
There are lines that look like this
```rs
                    let err_string = format!("user function panicked: {}\0",#r_name_str);
``` 
[...]

PR #718, opened 2024-03-30
fixex `throw_r_error` may crash R session #716

CGMossa:
Looking at the records, I had seen this bug before in 11ea513251f71889318d580a93d15200a7d83bf5, but I wasn't sure what was going on. Now, I'm very confident that this was because of a PR to use `CStr` more, but the feature-gated `result_list` wasn't tested, or something like that.


JosiahParry:
Note that this is related to issues https://github.com/extendr/extendr/issues/716 and https://github.com/pola-rs/r-polars/issues/993

JosiahParry (approved):
LGTM. Thank you for the fast fix! :O

JosiahParry:
Note that CI will be fixed by https://github.com/extendr/rextendr/pull/347

PR #721, opened 2024-04-05
No "as" for basic cases

latot:
Actually, the ```as``` keywords has a lot of effects, and one particular aspect of it is not to be exactly a way to convert between types, for example a i64 to i32 can be truncated, also will convert even if we lacks of the right traits to use ```TryFrom```/```Try```, which is not safe.

This PR will only handle basic cases of ```as``` and be replaced with the right trait.

Also, I have skip to use ```try_into```, that should be talked first, because use that in a lot of places also means to infer the type, which is great at some extent, but do it in each place could be complicated, and also means a less maintainable code.

You will notice there is a lot of ```unwrap``` around, as a simple fix is the right tool, because the code actually used ```as``` and does not considerate with it that the conversions could fail, so not all the functions has the ability to handle errors.
To not use ```unwrap``` and propagate the errors, there is a need of bigger changes.

### DIFF
--- a/extendr-api/src/deserializer.rs
+++ b/extendr-api/src/deserializer.rs
@@ -726,9 +726,9 @@ impl<'de> Visitor<'de> for RobjVisitor {
         E: serde::de::Error,
     {
         if v > i64::from(i32::MIN) && v <= i64::from(i32::MAX) {
-            i32::try_from(v)
+            i32::try_from(v)?.try_into()
         } else {
-            f64::try_from(v)
+            f64::try_from(v)?.try_into()
         }
     }
 
@@ -737,9 +737,9 @@ impl<'de> Visitor<'de> for RobjVisitor {
         E: serde::de::Error,
     {
         if v <= u64::from(i32::MAX) {
-            i32::try_from(v)
+            i32::try_from(v)?.try_into()
         } else {
-            f64::try_from(v)
+            f64::try_from(v)?.try_into()
         }
     }
 
--- a/extendr-api/src/deserializer.rs
+++ b/extendr-api/src/deserializer.rs
@@ -361,7 +361,7 @@ impl<'de> Deserializer<'de> for &'de Robj {
     where
         V: Visitor<'de>,
     {
-        visitor.visit_i128(i64::try_from(self.clone())? as i128)
+        visitor.visit_i128(i64::try_from(self.clone())?.try_into()?)
     }
 
     fn deserialize_u8<V>(self, visitor: V) -> Result<V::Value>
@@ -396,7 +396,7 @@ impl<'de> Deserializer<'de> for &'de Robj {
     where
         V: Visitor<'de>,
     {
-        visitor.visit_u128(u64::try_from(self.clone())? as u128)
+        visitor.visit_u128(u64::try_from(self.clone())?.try_into()?)
     }
 
     fn deserialize_f32<V>(self, visitor: V) -> Result<V::Value>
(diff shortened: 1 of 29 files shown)


extendr-api/src/deserializer.rs
```diff
@@ -725,21 +725,19 @@ impl<'de> Visitor<'de> for RobjVisitor {
     where
         E: serde::de::Error,
     {
-        if v > i32::MIN as i64 && v <= i32::MAX as i64 {
-            Ok((v as i32).into())
-        } else {
-            Ok((v as f64).into())
+        match i32::try_from(v) {
+            Ok(x) => x,
+            Err(_) => f64::try_from(v)?
```
latot, line 730: serialize section must be checked.

CGMossa: Great catch. Can you make this a separate issue?

latot: @CGMossa ready, I updated the code, now only will replace the "as" but will keep the logic on the code, there is a new issue for that things: https://github.com/extendr/extendr/issues/732

latot:
no idea what I'm missing testing locally.... because all tests works here, but in CI fails.

CGMossa:
I've been following along with this, and I'm very much pleased that you're attempting a substantial contribution to extendr! More of this please.
Extendr is a very particular project; It is far from a true rust-project, as it is more C/FFI heavy, and opinionated code, than anything else. There are compromises everywhere, and attempts at venturing forth where no-one else has been before. Everything is brittle, and sometimes it is easier to show, rather than tell. There are several serious PRs that have been left unmerged, or downright closed. PRs that targeted a missing functionality, that have interested parties. 

But that is the nature of this project. It isn't a straightforward where the objectives are, although the vision is more or less unified. The objectives vary with deeper understanding of what can be done between Rust and R.

This PR fails to show utility in the areas I'm focused on. I don't see it offering something better than the present. I also think it makes things more verbose than need be. If there is a rust release that permits full `TryFrom`-conversions between all standard types, then we can think about that! But that moves the MSRV further along, and that is a problem with the various platforms, extendr is expected to work on. 

But keep exploring! Keep opening PRs that illustrates certain ideas, and one day, the conversation might lead to progress. For now, I'd say that I'm not in favour of this particular PR.

lnicola:
> If there is a rust release that permits full TryFrom-conversions between all standard types, then we can think about that!

That seems unlikely, since it overloads the meaning of `TryFrom`: should `i32::try_from(3.14f)` fail the same way as `i32::try_from(i64::MAX)`? What about `f32::from(16_777_217)`?

> no idea what I'm missing testing locally.... because all tests works here, but in CI fails.

You need to run `cargo fmt`, as seen here:

image

Also, some of the changes here are bogus: for example https://github.com/extendr/extendr/pull/721/files#diff-f120a6cc51206b569a5056be368fed120b3e1d59bdc30dc4276de9692625bf94R929 can't work because there is no `impl TryFrom<i64> for f64`. It might seem to work for you because you're not testing with `--features serde`. If you do that, you'll get about 22 compile errors.

> will convert even if we lacks of the right traits to use TryFrom/Try, which is not safe

That's false. `as` is safe (or rather, not _unsound_). Can it lead to loss of precision or truncation? Yes, but that's either unavoidable or a bug in the code. Bugs happen anyway.

> because the code actually used as and does not considerate with it that the conversions could fail

As I mentioned above, you need to define what failure means. Do you want to replace truncation with a panic? That might not be very popular with the users of the crate.

---

It might be a good idea to split this and focus on the more important changes first. Are you fixing a bug? File that as a separate PR. Can you replace an `unwrap` with `?`? Great!

The other changes might be more controversial. And most of all, using `as` for widening conversions (`f32` to `f64`) is completely correct. There is a reason why those Clippy lints are allowed by default, including the one that's clearly spelled out in the docs:

> Note that this lint is specialized in linting every single use of as regardless of whether good alternatives exist or not. If you want more precise lints for as, please consider using these separate lints: unnecessary_cast, cast_lossless/cast_possible_truncation/cast_possible_wrap/cast_precision_loss/cast_sign_loss, fn_to_numeric_cast(_with_truncation), char_lit_as_u8, ref_to_mut and ptr_as_ptr.

latot:
@lnicola you are right, tomorrow I'll check all this again, and split it, to start with the most crucial ones first.

PR #722, opened 2024-04-06
extendr-impl: Support returning `&Self` and `&mut Self`

### LINKED ISSUE #614: Returning `self` from `extendr`-class creates a clone?
```R
Failure ('test-classes.R:15:3'): Exported class self ptr works
x$me() (`actual`) not equal to `x` (`expected`).

`actual` is <pointer: 0x0000024de24e44e0>
`expected` is <pointer: 0x0000024de24e44c0>

`actual` is attr(,"class")
`expected` is attr(,"class")

`actual` is [1] "MyClass"
`expected` is [1] "MyClass"

Failure ('test-classes.R:16:3'): Exported class self ptr works
x[["me"]]() (`actual`) not equal to `x` (`expected`).

`actual` is <pointer: 0x0000024de24e4590>
`expected` is <pointer: 0x0000024de24e44c0>

`actual` is attr(,"class")
`expected` is attr(,"class")

`actual` is [1] "MyClass"
`expected` is [1] "MyClass"
```

CGMossa:
Fixes #614 

This allows `extendr-impl` types to return `&Self` / `&mut Self`.
The logic is, that the `Robj` that was provided to the wrapper, is itself a reference-type,
that should be returned to R, instead of recreating the ExternalPtr.

For R users, this unlocks defining pipe-friendly interfaces using `extendr-impl`.

@sorhawell I believe this was a blocker for you, and making a pipe-friendly polars interface, right?

### DIFF
--- a/extendr-macros/src/extendr_impl.rs
+++ b/extendr-macros/src/extendr_impl.rs
@@ -5,35 +5,104 @@ use syn::{ItemFn, ItemImpl};
 
 use crate::wrappers;
 
-/// Handle trait implementations.
+/// Make inherent implementations available to R
+///
+/// The extendr_impl function is used to make inherent implementations
+/// avaialble to R as an environment. By adding the [`extendr`] attribute
+/// macro to an `impl` block (supported with `enum`s and `struct`s), the
+/// methods in the imple block are made available as functions in an
+/// environment.
+///
+/// On the R side, an environment with the same name of the inherent
+/// implementation is created. The environment has functions within it
+/// that correspond to each method in the impl block. Note that in order
+/// for an impl block to be compatible with extendr (and thus R), its return
+/// type must be able to be returned to R. For example, any struct that might
+/// be returned must _also_ have an `#[extendr]` annotated impl block.
 ///
 /// Example:
-/// ```ignore
+/// ```
 /// use extendr_api::prelude::*;
-/// #[derive(Debug)]
+///
+/// // a struct that will be used internal the People struct
+/// #[derive(Clone, Debug, IntoDataFrameRow)]
 /// struct Person {
-///     pub name: String,
+///     name: String,
+///     age: i32,
 /// }
+///
+/// // This will collect people in the struct
 /// #[extendr]
-/// impl Person {
+/// #[derive(Clone, Debug)]
+/// struct People(Vec<Person>);
+///
+/// #[extendr]
+/// /// @export
+/// impl People {
+///
+///     // instantiate a new struct with an empty vector
 ///     fn new() -> Self {
-///         Self { name: "".to_string() }
+///         let vec: Vec<Person> = Vec::new();
+///         Self(vec)
 ///     }
-///     fn set_name(&mut self, name: &str) {
-///         self.name = name.to_string();
+///
+///     // add a person to the internal vector
+///     fn add_person(&mut self, name: &str, age: i32) -> &mut Self {
+///         let person = Person {
+///             name: String::from(name),
+///             age: age,
+///         };
+///
+///         self.0.push(person);
+///
+///         // return self
+///         self
 ///     }
-///     fn name(&self) -> &str {
-///         self.name.as_str()
+///     
+///     // Convert the struct into a data.frame
+///     fn into_df(&self) -> Robj {
+///         let df = self.0.clone().into_dataframe();
+///
+///         match df {
+///             Ok(df) => df.as_robj().clone(),
+///             Err(_) => data_frame!(),
+///         }
 ///     }
+///
+///     // add another `People` struct to self
+///     fn add_people(&mut self, others: &People) -> &mut Self {
+///         self.0.extend(others.0.clone().into_iter());
+///         self
+///     }
+///
+///     // create a function to print the self which can be called
+///     // from an R print method
+///     fn print_self(&self) -> String {
+///         format!("{:?}", self.0)
+///     }
+///
 /// }
-/// #[extendr]
-/// fn aux_func() {
-/// }
-/// // Macro to generate exports
+///
+/// // Macro to generate exports.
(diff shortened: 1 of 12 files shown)


extendr-macros/src/wrappers.rs
```diff
@@ -50,7 +75,7 @@ pub fn make_function_wrappers(
 
     let call_name = if has_self {
         let is_mut = match inputs.iter().next() {
-            Some(FnArg::Receiver(ref reciever)) => reciever.mutability.is_some(),
```
JosiahParry, line 53: nice typo catch

extendr-macros/src/wrappers.rs
```diff
@@ -120,6 +145,43 @@ pub fn make_function_wrappers(
...
+    let return_is_ref_self = {
+        match sig.output {
+            syn::ReturnType::Default => false,
+            syn::ReturnType::Type(_, ref return_type) => match return_type.as_ref() {
+                // matches `-> Self`
+                // Type::Path(type_path) => type_path.path.is_ident("Self"),
+                // matches `-> &Self` / `-> &mut Self`
+                Type::Reference(ref reference_type) => {
+                    if let Type::Path(path) = reference_type.elem.as_ref() {
+                        path.path.is_ident("Self")
```
JosiahParry, line 159: Is it possible to reorder the comments? I'm not sure which is applying to what. I _think_ I've got it right? Except for the `Type::Path(type_path) => type_path.path.is_ident("Self"),` seems like it might be commented out as old code? 

```suggestion
             // matches `-> Self`
            syn::ReturnType::Type(_, ref return_type) => match return_type.as_ref() {
                Type::Reference(ref reference_type) => {
                    // Type::Path(type_path) => type_path.path.is_ident("Self"),
                    // matches `-> &Self` / `-> &mut Self`
                    if let Type::Path(path) = reference_type.elem.as_ref() {
                        path.path.is_ident("Self")
```

extendr-macros/src/wrappers.rs
```diff
@@ -120,6 +145,43 @@ pub fn make_function_wrappers(
...
+        quote!(Ok(extendr_api::Robj::from(#call_name(#actual_args))))
+    } else {
+        // instead of converting &Self / &mut Self, pass on the passed
+        // ExternalPtr<Self>
+        quote!(
+            let _return_ref_to_self = #call_name(#actual_args);
+            //FIXME: find a less hardcoded way to write `_self_robj`
+            Ok(_self_robj)
+        )
+    };
```
JosiahParry, line 182: I suggest flipping the order of the if else so that it is more readable instead of negating the bool value

```suggestion
    let return_type_conversion = if return_is_ref_self {
         // instead of converting &Self / &mut Self, pass on the passed
        // ExternalPtr<Self>
        quote!(
            let _return_ref_to_self = #call_name(#actual_args);
            //FIXME: find a less hardcoded way to write `_self_robj`
            Ok(_self_robj)
        )
    } else {
      quote!(Ok(extendr_api::Robj::from(#call_name(#actual_args))))
    };
```

extendr-macros/src/wrappers.rs
```diff
@@ -120,6 +145,43 @@ pub fn make_function_wrappers(
...
+
+    // TODO: the unsafe in here is unnecessary
```
JosiahParry, line 184: If the unsafe is removed, will this still compile? If so, should it be removed?

extendr-macros/src/wrappers.rs
```diff
@@ -1,3 +1,28 @@
...
+//!
+//! You may also manually implement these wrappers, in order to do special
+//! type-checking, or other annotation, that could be more convenient to do
+//! on the R-side. The C-functions are named according to `"{WRAP_PREFIX}{prefix}{mod_name}"`.
+//! See [`WRAP_PREFIX`], and note that `prefix` is set specifically for methods in
+//! `extendr`-impl blocks, while for functions have no prefix.
+//!
+//! [`R_NilValue`]: ::libR_sys::R_NilValue
+//! [`SEXP`]: ::libR_sys::SEXP
+
```
JosiahParry, line 25: this is wildly helpful!

JosiahParry (changes requested):

JosiahParry:
@CGMossa this looks pretty neat! Also great catches on all of the typos and adding doc!! I made some small requested changes (all related to readability). Then I'll test this locally.

extendr-macros/src/wrappers.rs
CGMossa: Not really old code, but it is here I signal that taken `self: Self` is not handled, and will be `false`, and it is not allowed.
Also, now that I think about it, people might write

```rs

struct Model;

#[extendr]
impl Model {
    fn take(self: Model) {
        
    }
}
```
... which is frightening to think about. I should find a solution for this...

extendr-macros/src/wrappers.rs
CGMossa: I was warned that there is an unnecessary unsafe. I can't myself see which one it is.

CGMossa:
Okay! I've reworded and refactored a little bit, hopefully it is more clear to you (and me.. because I got confused faaast).

CGMossa:
Since we can have snapshot tests in the extendr code, I've added such tests here. Note that there is an overlap with the snapshot test we already have, but I guess we'll have more of these, so separating them might be a good idea.

extendr-macros/src/extendr_impl.rs
```diff
@@ -5,35 +5,104 @@ use syn::{ItemFn, ItemImpl};
 
 use crate::wrappers;
 
-/// Handle trait implementations.
+/// Make inherent implementations available to R
```
JosiahParry, line 8: @CGMossa I've updated the doc here.

JosiahParry (approved):

PR #726, opened 2024-04-06
`#[extendr]`-impl: Return the right reference

### LINKED ISSUE #725: extendr-impl: returning &Self from impl always returns the calling item
https://github.com/extendr/extendr/pull/722 enabled an impl to return itself and not a clone. However, there is a limitation to the implementation in that the return is a _different_ object the original one is always returned. 
Also related to #614 

For example this will not return the `other`

```rust
struct People;
#[extendr]
impl People {

    fn new() -> Self {
        People {}
    }
  
    fn return_other(&self, other: &'static People) -> &Self {
        other
    }
}
```

From R 

```r
x <- People$new()
y <- People$new()

identical(x, y)
#> TRUE
```

CGMossa:
Following the polishing of `ExternalPtr` is this PR on returning the _right_ reference.

First, we remove the ability to take `&T` and make `ExternalPtr<T>`. This is a flawed construct: Making an `ExternalPtr` is not just wrapping a pointer, it is also signing a pointer up for destruction. Converting a `&T` to `ExternalPtr<T>` may cause spurious double-free events.

The Rules of ExternalPtr are simple:

* From an owned type, one may create an Owned Pointer, and from an owned pointer, one can instantiate a true `ExternalPtr`.
* From `&ExternalPtr<T>` you can create `&T`, and similarly for `&mut ExternalPtr<T>` and `&mut T`.

So, should there be no references used or returned? A reference that comes from a true `ExternalPtr<T>` may be used. We embed this logic on the wrapping function, rather than do any C-crimes.

If a reference corresponds to any of the `ExternalPtr<T>` in the argument list, then that is returned. 

Fixes #725

Depends on
- [x] #765 
- [x] https://github.com/extendr/extendr/pull/766


CGMossa:
At this stage, I now check the address of the returned-object, and see if it matches the ExternalPtr from Self.
```rs
Error (test-test-submodules.R:32:3): Return the right externalptr
Error in `x$max_ref(y)`: only able to return reference to self
Backtrace:
    ▆
 1. └─x$max_ref(y) at test-test-submodules.R:32:3
 ```
 
Next step is to go through all arguments, that are externalptr, and return the one that maches the output, or error otherwise.

Thankfully, it is impossible to create a local variable and return a reference from it in Rust.

**Comment by CGMossa** [comment]
This is ready for _feedback_, not necessarily review.

**Review by CGMossa — COMMENTED** [review]
- [ ] Remove direct usage of libR-sys in user-facing proc-macro

**Comment by CGMossa** [comment]
This is ready for review. I had to export (privately) one thing from `libR-sys`. Lesson learned: Derive-macros cannot do or access things, that its caller can't do themselves. Meaning, we can't use unimported crates, we can't use unexported types, etc.

File: extendr-api/src/wrapper/externalptr.rs [hunk 1]
```diff
@@ -248,21 +248,8 @@ impl<T> TryFrom<&Robj> for ExternalPtr<T> {
...
-                _marker: std::marker::PhantomData,
-            };
-            Ok(res)
-        } else {
-            Err(Error::ExpectedExternalPtrType(
-                clone,
-                std::any::type_name::<T>().into(),
-            ))
-        }
+        let result: &Self = robj.try_into()?;
```
Ilia-Kosenkov, line 251: I think this one is also part of another PR

extendr-macros/src/wrappers.rs
```diff
@@ -91,11 +91,27 @@ pub(crate) fn make_function_wrappers(
...
         .map(|input| translate_formal(input, self_ty))
         .collect::<syn::Result<Punctuated<FnArg, Token![,]>>>()?;
 
+    // extract the names of the arguments only (`mut` are ignored in `formal_args` already)
+    let sexp_args = formal_args
+        .clone()
+        .into_iter()
+        .map(|x| match x {
+            // the wrapper doesn't use `self` arguments
+            FnArg::Receiver(_) => unreachable!(),
```
Ilia-Kosenkov, line 106: Is this 'never happens' or 'should not happen'? Maybe we can `panic!()` with a more informative message?

extendr-macros/src/wrappers.rs
```diff
@@ -219,7 +242,7 @@ pub(crate) fn make_function_wrappers(
                     // included in panic. The advantage would be the panic cause could be included
                     // in the R terminal error message and not only via std-err.
                     // but it should be handled in a separate function and not in-lined here.
-                    let err_string = format!("user function panicked: {}",#r_name_str);
+                    let err_string = format!("user function panicked: {}", #r_name_str);
```
Ilia-Kosenkov, line 245:
```suggestion
                    let err_string = format!("User function panicked: {}", #r_name_str);
```

tests/extendrtests/tests/testthat/test-test-submodules.R
```diff
@@ -21,3 +21,25 @@ test_that("Classes defined in submodules also works", {
...
+  expect_false(isTRUE(identical(x, y)))
+
+  max_class <- x$max_ref(y)
+  expect_identical(max_class, y)
+  expect_false(isTRUE(identical(x, max_class)))
+
+  # check if you can have regular arguments as well
+  x <- MySubmoduleClass$new()
+  x$set_a(22)
+  x$max_ref_offset(y, 30)
```
Ilia-Kosenkov, line 39: Why don't you use the return value here?

tests/extendrtests/tests/testthat/test-test-submodules.R
```diff
@@ -21,3 +21,25 @@ test_that("Classes defined in submodules also works", {
     isTRUE(identical(x$me_owned(), x))
   )
 })
+
+test_that("Return the right externalptr", {
```
Ilia-Kosenkov, line 25: Again, I would say I see here three distinct tests, perhaps it is worth splitting it into smaller pieces? (And also having all externalptr-related tests in a single file, not just some `test-submodules`)?

extendr-macros/src/wrappers.rs
CGMossa: I believe this is never happens, and will never happen. In the definition of `formal_args`, the `self` is replaced with a variable name, as `extern "C"` functions cannot take `Self`, and never will. Then, we merely extract the produced argument names, to have all the `SEXP`-typed variables in the wrapper available to us. 

But `unreachable` is just a panic, so if you want to suggest a message, then we can do that.

extendr-api/src/wrapper/externalptr.rs
CGMossa: You're absolutely right. I put it in the top post, and now, it has been merged, and I've rebased.

tests/extendrtests/tests/testthat/test-test-submodules.R
CGMossa: I'm using it now. Thanks!

tests/extendrtests/tests/testthat/test-test-submodules.R
CGMossa: I've split off the `externalptr` tests into separate rust modules, testthat file, and everything. Hope this satisifies this point.

CGMossa:
@Ilia-Kosenkov I've dealt with the comments. I'd love it if you could give this another pass.
The PR has become a bit bigger, but that's due to the requested changes, and their impact.

PR #736, opened 2024-04-12
Replace buggy number conversions with tryfrom

latot:
This is a first part of replace buggy number conversions, it covers most of cases where we transform between numerical types and the conversion can fails.

There is still some cases like:

extendr-api/src/scalar/rint.rs:189:26
```
let result = v as i32;
```

But will not be included here.

### DIFF
--- a/extendr-api/src/wrapper/altrep.rs
+++ b/extendr-api/src/wrapper/altrep.rs
@@ -155,9 +155,12 @@ fn manifest(x: SEXP) -> SEXP {
     single_threaded(|| unsafe {
         Rf_protect(x);
         let len = XLENGTH_EX(x);
-        let data2 = Rf_allocVector(TYPEOF(x) as u32, len as R_xlen_t);
+        let data2 = Rf_allocVector(
+            u32::try_from(TYPEOF(x)).unwrap(),
+            R_xlen_t::try_from(len).unwrap(),
+        );
         Rf_protect(data2);
-        match TYPEOF(x) as u32 {
+        match u32::try_from(TYPEOF(x)).unwrap() {
             INTSXP => {
                 INTEGER_GET_REGION(x, 0, len as R_xlen_t, INTEGER(data2));
             }
@@ -565,8 +568,8 @@ impl Altrep {
                 Robj::from_sexp(class),
                 Robj::from_sexp(state),
                 Robj::from_sexp(attr),
-                objf as i32,
-                levs as i32,
+                i32::try_from(objf).unwrap(),
+                i32::try_from(levs).unwrap(),
             )
             .get()
         }
@@ -618,7 +621,10 @@ impl Altrep {
         }
 
         unsafe extern "C" fn altrep_Length<StateType: AltrepImpl + 'static>(x: SEXP) -> R_xlen_t {
-            Altrep::get_state::<StateType>(x).length() as R_xlen_t
+            Altrep::get_state::<StateType>(x)
+                .length()
+                .try_into()
+                .unwrap()
         }
 
         unsafe extern "C" fn altvec_Dataptr<StateType: AltrepImpl + 'static>(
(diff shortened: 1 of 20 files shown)


extendr-api/src/wrapper/altrep.rs
```diff
@@ -155,9 +155,12 @@ fn manifest(x: SEXP) -> SEXP {
     single_threaded(|| unsafe {
         Rf_protect(x);
         let len = XLENGTH_EX(x);
-        let data2 = Rf_allocVector(TYPEOF(x) as u32, len as R_xlen_t);
+        let data2 = Rf_allocVector(
+            u32::try_from(TYPEOF(x)).unwrap(),
```
Ilia-Kosenkov, line 159: This is safe

extendr-api/src/wrapper/altrep.rs
```diff
@@ -155,9 +155,12 @@ fn manifest(x: SEXP) -> SEXP {
...
+            R_xlen_t::try_from(len).unwrap(),
+        );
         Rf_protect(data2);
-        match TYPEOF(x) as u32 {
+        match u32::try_from(TYPEOF(x)).unwrap() {
```
Ilia-Kosenkov, line 163: And this, see comments above

Ilia-Kosenkov (changes requested):
Overall good work, but there are two main issues here:
- Some conversions are inherently safe and should not be wrapped into `TryFrom` 
- Other conversions repeat throughout the codebase. This increases cognitive load, you should either create simple helpers or trivial traits that can then be used. Basically, abstract away `TInto::try_into::<TFrom>(&TFrom).unwrap()` or something like this/

CGMossa:
Hello again!

Unfortunately, I'm still not in favour of the changes proposed here.
The reason for this, is that the conversations currently present should be eliminated elsewhere, or through a refactor.
Take for instance the `TYPEOF`-conversions. I've submitted PRs in `libR-sys`-repo that deals with these in
https://github.com/extendr/libR-sys/pull/210 and https://github.com/extendr/libR-sys/pull/211. These may be reviewed and merged now, which will remove these `as _` conversations.

Similarly, the spot where you've got things like `INTEGER()` would be removed in a refactor soon. There is no need
to make these conversions more "safe" for multiple reasons: This code will be removed in a larger refactor that we've been battling with (nicknamed `use-try-from` as default). They add noise, without propagating the error elsewhere in the API. When we are converting things that come from FFI, I'm not sure the conversions can be made "safer" than `as` anyways. `as`-conversions are C-like conversions to begin with.

latot:
@CGMossa I mostly agree with you, this PR is far from the ideal fix, a lot things should be fixed by a refactor, but this is the point, to solve by refactor you need know a lot better the program, also takes time, so if the refactor can't be done in the short term, this PR would be good enough. As you says, how there is already PRs that could solve this, and in a better way, that is great! I would like to know it before before spent my time in this particular case and could be do other thing.

The ```as``` case is pretty simple, should only/mainly be used if the conversion is done right, like from u8 to u16, but in this case, the PR only focus in conversions that can not keep the numbers without truncation, if you want to truncate that numbers then is fine, if not then ```as``` is not the way, even en C/C++ could be a problem. Use the TryFrom solve this.

Maybe just wait until the PRs you says, and then check again this, to see if there is some unexpected truncation left.

CGMossa:
Dear @latot

Again, I thank you very much for your contribution and attention with the development of extendr.
One of the larger refactoring has just landed in #742. It has taken some effort, but it is finally there.
I would appreciate it if you went there, and gave it a review. You have already went through extendr in many of the same places, and it would be good to have as many attentive eyes on such a large PR as possible.

I'd also like to encourage participating in reviewing PRs, and replying to issues, as we have very few conversations going in-depth into what are all the aspects that need attention.

While a community review doesn't yield approval on a PR, it shows approval of the community, which in many cases are much, much more valuable. 

I'll make an effort to land a few more refactoring very soon.

CGMossa:
Dear @latot 
Today, #742 just landed, which addresses quite many of these conversions directly. It took a lot of effort but it is in. I would like to close this PR as it has been superseded.

If I have overlooked a conversion, or similar, there are more things you'd like to contribute to extendr,
I'd very much encourage you to open a new PR.

Godspeed!

PR #744, opened 2024-04-28
Add DLLInfo to AltRep class registration

### LINKED ISSUE #743: Extendr doesn't create the ALTREP class in the supposed way?
(There's no official documentation, so I might be wrong.)

An ALTREP class is created by `R_make_alt<TYPE>_class()`. Here's the code in extendr:

https://github.com/extendr/extendr/blob/b3118a7aeaf13875605edb339e70829321601324/extendr-api/src/wrapper/altrep.rs#L657-L681

In this function, extendr chooses to use `std::ptr::null_mut()` for the third argument, but this seems unusual.

You can search on GitHub for `R_make_altreal_class()`, for example. You should notice all the other packages pass `dll`, which is a `DllInfo*`.

https://github.com/search?q=org%3Acran+R_make_altreal_class&type=code

These functions are supposed to be executed in the initialization routine like this.

```c
void R_init_FOO(DllInfo *dll) {  // FOO is your package name
  R_registerRoutines(dll, NULL, CallEntries, NULL, NULL);
  R_useDynamicSymbols(dll, FALSE);

  init_altrep_class(dll); // this!
}
```

For a real example, this is how the vroom package does (which seems generated cpp11's undocumented? decorator `[[cpp11::init]]`).

https://github.com/cran/vroom/blob/db3e8321804434afb3b48f4947a11d30f1db4b26/src/cpp11.cpp#L165-L179

So, can this be a real problem? I'm not sure. In the current implementation of R, it seems users might be in trouble when unloading the DLL, but it's probably relatively rare.

https://github.com/wch/r-source/blob/ab9c4114cd774d359f6d4c3b4a69e7ba9d2864c8/src/main/Rdynload.c#L712C1-L712C45
https://github.com/wch/r-source/blob/ab9c4114cd774d359f6d4c3b4a69e7ba9d2864c8/src/main/altrep.c#L96-L105  (in extendr's case, `R_ExternalPtrAddr(iptr) == dll` is false because extendr doesn't capture the `DllInfo*`).

CGMossa:
Fixes #743

### DIFF
--- a/extendr-api/src/wrapper/altrep.rs
+++ b/extendr-api/src/wrapper/altrep.rs
@@ -654,29 +654,24 @@ impl Altrep {
             let csname = std::ffi::CString::new(name).unwrap();
             let csbase = std::ffi::CString::new(base).unwrap();
 
+            let dll_info = libR_sys::R_getEmbeddingDllInfo();
             let class_ptr = match ty {
                 Rtype::Integers => {
-                    R_make_altinteger_class(csname.as_ptr(), csbase.as_ptr(), std::ptr::null_mut())
+                    R_make_altinteger_class(csname.as_ptr(), csbase.as_ptr(), dll_info)
                 }
-                Rtype::Doubles => {
-                    R_make_altreal_class(csname.as_ptr(), csbase.as_ptr(), std::ptr::null_mut())
-                }
+                Rtype::Doubles => R_make_altreal_class(csname.as_ptr(), csbase.as_ptr(), dll_info),
                 Rtype::Logicals => {
-                    R_make_altlogical_class(csname.as_ptr(), csbase.as_ptr(), std::ptr::null_mut())
+                    R_make_altlogical_class(csname.as_ptr(), csbase.as_ptr(), dll_info)
                 }
-                Rtype::Raw => {
-                    R_make_altraw_class(csname.as_ptr(), csbase.as_ptr(), std::ptr::null_mut())
-                }
+                Rtype::Raw => R_make_altraw_class(csname.as_ptr(), csbase.as_ptr(), dll_info),
                 Rtype::Complexes => {
-                    R_make_altcomplex_class(csname.as_ptr(), csbase.as_ptr(), std::ptr::null_mut())
+                    R_make_altcomplex_class(csname.as_ptr(), csbase.as_ptr(), dll_info)
                 }
                 Rtype::Strings => {
-                    R_make_altstring_class(csname.as_ptr(), csbase.as_ptr(), std::ptr::null_mut())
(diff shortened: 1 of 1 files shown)


CGMossa:
@yutannihilation Do you think this suffices as a fix? Because otherwise I have to do a very long dance to... take all the registery code and embed it in the same way I see there.

* `entrypoint.c` generation lives in `rextendr`, which now, has to be moved to `extendr`. 
* Include the library `#include <R_ext/Rdynload.h>` and change the `void` to `DllInfo`.
* Somehow, start registering altrep so they can be exported together, probably by extending `#[extendr]` or something smart/convoluted.

yutannihilation:
> Because otherwise I have to do a very long dance to...

Yeah, I think you are right. It's probably going to be a very long way. Honestly, I have no idea how extendr can achieve it. I don't know the function, but `R_getEmbeddingDllInfo()` seems for embedded usage.

https://cran.r-project.org/doc/manuals/R-exts.html#Registering-symbols

> `entrypoint.c` generation lives in `rextendr`

I think this is not correct. All the actual logic is generated by the macros.

https://github.com/extendr/rextendr/blob/e5072fca6c7db8b67ea010cac6aa92dec29fdb7a/inst/templates/entrypoint.c#L7C1-L7C40

yutannihilation:
Just in case you are interested in, here's what I'm currently trying to implement in my framework.

https://github.com/yutannihilation/savvy/pull/206

CGMossa:
Alright. Thanks for the assist! I've marked it DRAFT, as now I see this is definitely wrong:
```c
/* returns DllInfo used by the embedding application.
   the underlying "(embedding)" entry is created if not present */
DllInfo *R_getEmbeddingDllInfo(void)
{
    DllInfo *dll = R_getDllInfo("(embedding)");
    if (dll == NULL) {
	int which = addDLL(Rstrdup("(embedding)"), "(embedding)", NULL);
	dll = LoadedDLL[which];
	/* make sure we don't attempt dynamic lookup */
	R_useDynamicSymbols(dll, FALSE);
    }
    return dll;
}
```
But there is a `R_getDllinfo`. Maybe I can use that.

yutannihilation:
> But there is a `R_getDllinfo`. Maybe I can use that.

FYI:
https://github.com/r-devel/r-svn/commit/b5582bb6dc31e988de5c33ebb28932901c2206ae#diff-048b7b7178313f203daf48b349a0f392b2bd11d397a5f207bf43629c98e1e1d9R88

CGMossa:
Good call. I also knew it would be problematic otherwise, because it is not
the right dll anyways. It has to be the actual package dll.. I better just
roll up my sleeves and make it happen.. By writing a macro... But I have
some maintenance PRs first, plus a few inventions... I hope the extendr
contributor crew approve them soon.
Cheers

PR #745, opened 2024-04-29
Add `Attribute` to all vector types

CGMossa:
It is not posible to set the attributes to vector-types. This is currently achievable by converting to `Robj`, setting attribute, and converting back. This indirection should not be necessary.


JosiahParry:
This is great! We'll need to add some integration tests, though—I know, I know. 

I can try to do it this evening.

JosiahParry:
At present this does not work because it requires the value to be mutable. It means that `.set_attrib()` cannot be used on a variable that is provided as an argument without first reassigning it a la `let mut x = x`. Note that `.as_mut()` will not work either because it returns a slice. 

```rust
#[extendr]
fn strings_named(x: Strings) -> Strings {
    let mut x = x;
    x.set_attrib(
        "names",
        x.iter()
            .map(|xi| xi.as_str().to_string())
            .collect::<Vec<_>>(),
    )
    .unwrap();
    x
}
```

This implementation implies that it should only be used on items created inside the body of the function. Now, this makes sense to some degree because the attributes are added to the SEXP by reference and not by cloning. 

Note that this behavior is true today, though it is more opaque:

``` r
library(rextendr)

rust_function(r"-(
fn tst(x: Robj) {
  x.set_attrib("hello", &["world"]);
}
)-")
#> ℹ build directory: '/private/var/folders/wd/xq999jjj3bx2w8cpg7lkfxlm0000gn/T/RtmpuaE5gG/file180d32dbc0427'
#> ✔ Writing '/private/var/folders/wd/xq999jjj3bx2w8cpg7lkfxlm0000gn/T/RtmpuaE5gG/file180d32dbc0427/target/extendr_wrappers.R'

x <- list()
tst(x)
x
#> list()
#> attr(,"hello")
#> [1] "world"
```

JosiahParry:
For this reason, I actually think this PR does a _better_ job than the status quo for handling attributes. I think, however, this is another reason why  we should allow mutable arguments.

CGMossa:
Snapshot tests generated, and various warnings have been eliminated.

I think the snapshot file with `new` in it should be deleted. Just did it.

@JosiahParry Maybe we should attempt to document the `Attribute` better?

tests/extendrtests/src/rust/src/attributes.rs
```diff
@@ -0,0 +1,47 @@
+use extendr_api::prelude::*;
+
+/// Adds an attribute to a vector of doubles
+/// 
+/// @param x Vector of doubles 
+/// 
+/// @export
```
CGMossa, line 7: I highly recommend not exporting functions. #[extendr] macro makes it available for testing on it is own.
Exported things has to be documented, as per `r-cmd-check`.
This means I've had to endure a cycle of:
```sh
cargo extendr document
cargo extendr devtools-test
cargo extendr devtools-test -a
cargo extendr r-cmd-check
```

To ensure that this was documented appropriately, to pass CI.

CHANGELOG.md
```diff
@@ -8,11 +8,14 @@
 - Adds `From<T> for Robj` for impl blocks marked with `#[extendr]` macro
 - The new `ExpectedExternalNonNullPtr` error variant provides a more informative error when a null pointer is accessed
 - `RArray::data_mut` provides a mutable slice to the underlying array data [[#657]](https://github.com/extendr/extendr/pull/657)
+- Implements the `Attributes` trait for all wrapper structs, allowing for easy access and setting of the attributes of an R object [[#745]](https://github.com/extendr/extendr/pull/745). This comes with breaking changes. See below.
 
 ### Changed
 
+- Breaking: the `Attributes` trait now requires a mutable reference to `self`. [[#745]](https://github.com/extendr/extendr/pull/745). Previously `.set_attrib()` would modify an object in place without requiring a mutable reference. `.set_attrib()` still modifies the object in place, but requires a mutabl reference to `self`.
```
JosiahParry, line 15: @CGMossa please confirm that this is sufficiently documented

JosiahParry (approved):
This looks great now! Thank you for your help @CGMossa. Please review my additions to the changelog. Once you approve of them, I think we should merge this.

tests/extendrtests/src/rust/src/attributes.rs
```diff
@@ -0,0 +1,42 @@
+use extendr_api::prelude::*;
+
+#[extendr]
+fn dbls_named(x: Doubles) -> Doubles {
```
CGMossa, line 4: Now that the `@export` is removed, the `cargo extendr document` needs to be called to update wrappers.
I'm pushing a commit that does this now.

CHANGELOG.md
CGMossa: The second statement should sound:
> - Breaking: the `Attributes` trait now returns a mutable reference to `Self`. [[#745]](https://github.com/extendr/extendr/pull/745). Previously `.set_attrib()` would modify an object in place, and then return an untyped owned pointer (Robj). Instead, now we return `&mut Self`.

CHANGELOG.md
```diff
@@ -8,11 +8,14 @@
 - Adds `From<T> for Robj` for impl blocks marked with `#[extendr]` macro
 - The new `ExpectedExternalNonNullPtr` error variant provides a more informative error when a null pointer is accessed
 - `RArray::data_mut` provides a mutable slice to the underlying array data [[#657]](https://github.com/extendr/extendr/pull/657)
+- Implements the `Attributes` trait for all wrapper structs, allowing for easy access and setting of the attributes of an R object [[#745]](https://github.com/extendr/extendr/pull/745). This comes with breaking changes. See below.
```
CGMossa, line 11:
```suggestion
- Implements the `Attributes` trait for all R vector wrapper structs (e.g. `Integers` , `Doubles`, `Strings`, etc.), allowing for easy access and setting of the attributes of an R object [[#745]](https://github.com/extendr/extendr/pull/745). This comes with breaking changes. See below.
```

CGMossa (commented):
I've left my comments. Hopefully, this suffices.

JosiahParry:
@CGMossa thank you for your work! All checks except the latest on windows due to CI are passing. Integration tests are looking good as well.

Thank you again for your quick collaboration on getting this through!!

PR #748, opened 2024-04-29
Various documentation updates

CGMossa:
These are plucked from the fork I've been toying with. I'm open to changes, but mostly I want move this along

PR #749, opened 2024-04-30
ci: slide rtools versions since r4.4 was released

CGMossa:
Failure observed in various PRs are due to incorrect version of Rtools being used in CI script. These weren't updated with the release of R 4.4. 

On `libR-sys`, we've made a change on this, but it still requires manual updating after each new release. We should find a solution for this going forward...

This was made by advice of @yutannihilation


JosiahParry (commented):
I think we can close this as the CI was updated in #748

PR #750, opened 2024-05-03
Attribute takes `Self` and return `Self` 

CGMossa:
Since we are already using the word breaking, we might as well explore this strategy:
Instead of taking `&mut Self` and turning `Self `, what if we took `Self` anyways?

It makes a bit of since, because R-objects (and its wrappers) are pointers to begin with, meaning they are references to begin with.

### DIFF
--- a/extendr-api/src/deserializer.rs
+++ b/extendr-api/src/deserializer.rs
@@ -802,9 +802,8 @@ impl<'de> Visitor<'de> for RobjVisitor {
             keys.push(key);
             values.push(value);
         }
-        let mut result = List::from_values(values);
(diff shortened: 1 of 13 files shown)


JosiahParry:
`Robj` might be a type of reference, but it isn't annotated as such—e.g. we are not always typing `&Robj`. 

This PR would revert the changes in #745 which requires `&mut self` to modify attributes. The current state of affairs takes `self` and returns `Self` **and** modifies the original struct in place. This is undesirable. By having the `&mut ` annotation we make it clear what is happening. As such, I think we should prefer to the behavior in #745 and acknowledge that the change is breaking _for good reason_.

PR #752, opened 2024-05-04
`extendr-macros`: Accept argument alias with `mut` in front 

CGMossa:
Follow-up to #745, this PR fixes the processing of `#[extendr]`-`fn`to allow for

```rs
#[extendr]
fn do_magic_to(mut x: Doubles) {
  // ...
}
```
while prior to this, one had to
```rs
#[extendr]
fn do_magic_to2(x: Doubles) {
  let mut x = x;
  //...
}
```

Thus, this PR is a bug fix, than an expansion of the API. 

Allow me to state what this PR _does not do_:
* You are no more allowed to change passed data from R in-place, before or after this PR.
If extendr-api permitted the manipulation of the internal data, than that is the same, and
this PR does not allow you more capabilities to mutate data.

There is a difference between having a mutable _alias_ and a mutable reference to data.

### DIFF
--- a/extendr-macros/src/wrappers.rs
+++ b/extendr-macros/src/wrappers.rs
@@ -354,14 +354,13 @@ pub fn translate_formal(input: &FnArg, self_ty: Option<&syn::Type>) -> syn::Resu
         // function argument.
         FnArg::Typed(ref pattype) => {
             let pat = &pattype.pat.as_ref();
-            let pat_ident = match pat {
-                syn::Pat::Ident(ref pat_ident) => &pat_ident.ident,
-                _ => {
-                    return Err(syn::Error::new_spanned(
-                        input,
-                        "failed to translate name of argument",
-                    ))
-                }
+            let pat_ident = if let syn::Pat::Ident(ref pat_ident) = pat {
+                &pat_ident.ident
+            } else {
+                return Err(syn::Error::new_spanned(
+                    input,
+                    "failed to translate name of argument",
+                ));
             };
             Ok(parse_quote! { #pat_ident : extendr_api::SEXP })
         }
--- a/extendr-macros/src/wrappers.rs
+++ b/extendr-macros/src/wrappers.rs
@@ -353,17 +353,9 @@ pub fn translate_formal(input: &FnArg, self_ty: Option<&syn::Type>) -> syn::Resu
     match input {
         // function argument.
         FnArg::Typed(ref pattype) => {
-            let pat = &pattype.pat.as_ref();
+            let pat = pattype.pat.as_ref();
             // ensure that `mut` in args are ignored in the wrapper
-            let pat_ident = match pat {
-                syn::Pat::Ident(ref pat_ident) => &pat_ident.ident,
-                _ => {
-                    return Err(syn::Error::new_spanned(
-                        input,
-                        "failed to translate name of argument",
-                    ));
-                }
-            };
+            let pat_ident = translate_only_alias(pat)?;
             Ok(parse_quote! { #pat_ident: extendr_api::SEXP })
         }
         // &self / &mut self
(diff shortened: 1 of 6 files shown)


tests/extendrtests/src/rust/src/altrep.rs
```diff
@@ -47,7 +47,7 @@ fn new_usize(robj: Integers) -> Altrep {
 
 #[cfg(not(use_r_altlist))]
 #[extendr]
-fn new_usize(robj: Integers) -> Robj {
+fn new_usize(_robj: Integers) -> Robj {
```
JosiahParry, line 50: i love that these tests double count now!

extendr-macros/src/wrappers.rs
```diff
@@ -354,7 +354,16 @@ pub fn translate_formal(input: &FnArg, self_ty: Option<&syn::Type>) -> syn::Resu
...
+            let pat_ident = match pat {
+                syn::Pat::Ident(ref pat_ident) => &pat_ident.ident,
+                _ => {
+                    return Err(syn::Error::new_spanned(
+                        input,
+                        "failed to translate name of argument",
+                    ))
+                }
+            };
+            Ok(parse_quote! { #pat_ident : extendr_api::SEXP })
```
JosiahParry, line 366: Does this catch errors where previously, they were swallowed by the `parse_quote!` macro?

extendr-macros/src/wrappers.rs
```diff
@@ -381,7 +390,15 @@ fn translate_meta_arg(input: &mut FnArg, self_ty: Option<&syn::Type>) -> syn::Re
...
             let pat = pattype.pat.as_ref();
             let ty = pattype.ty.as_ref();
-            let name_string = quote! { #pat }.to_string();
+            let pat_ident = if let syn::Pat::Ident(ref pat_ident) = pat {
+                &pat_ident.ident
+            } else {
+                return Err(syn::Error::new_spanned(
+                    input,
+                    "failed to translate name of argument",
+                ));
```
JosiahParry, line 399: Do you mind swapping this to a `match` statement like you have about? Its easier to understand 😅 

Are these `syn` patterns that you are now catching that previously weren't?

extendr-macros/src/wrappers.rs
CGMossa: I don't think so. It is hard to reason about this. But basically, I think it is because parsing this requires handling all the cases.. but here I don't think there are any other cases. I think `mut arg1` was simply being plastered everywhere, instead of `arg1` being extracted and used everywhere, and then `mut` just being amended when needed..

extendr-macros/src/wrappers.rs
CGMossa: Done

extendr-macros/src/wrappers.rs
```diff
@@ -354,7 +354,17 @@ pub fn translate_formal(input: &FnArg, self_ty: Option<&syn::Type>) -> syn::Resu
...
+            // ensure that `mut` in args are ignored in the wrapper
+            let pat_ident = match pat {
+                syn::Pat::Ident(ref pat_ident) => &pat_ident.ident,
+                _ => {
+                    return Err(syn::Error::new_spanned(
+                        input,
+                        "failed to translate name of argument",
+                    ));
+                }
+            };
```
Ilia-Kosenkov, line 366: Duplicate code fragment

CGMossa: What do you suggest I do?

Ilia-Kosenkov: Ideally of course extract it as a function, but if it cannot be done (I do not see exact types here so maybe there is a mismatch), then let's leave it as is

CGMossa: Good suggestion; I personally don't like how many functions this part of extendr has.. but I think I've made it work! Please check the latest commit.

Ilia-Kosenkov (approved):

PR #754, opened 2024-05-11
Various maintenance

CGMossa:
- [x] Added `non-api` option that exposes functionality that was considered API prior to R 4.4. We are marking these `unsafe`. This is considered a breaking change.
- [x] As we are now testing multiple versions of R, actively, we may switch to R environments that does not have all the necessary packages for development.
This PR adds a check and prints a message with the useful command-line instruction to install missing packages.
- [x] I've also added instructions for using development versions of extendr, with rextendr in `tests/extendrtests/README.md`. 
- [x] This PR became contingent on this https://github.com/extendr/rextendr/pull/352

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,6 +45,7 @@ jobs:
 
     env:
       R_REMOTES_NO_ERRORS_FROM_WARNINGS: true
+      REXTENDR_SKIP_DEV_TESTS: TRUE
 
       # This environment variable enables support for pseudo multi-target cargo builds.
       # Current stable Rust does not support multi-targeting,
@@ -286,6 +287,7 @@ jobs:
 
     env:
       R_REMOTES_NO_ERRORS_FROM_WARNINGS: true
+      REXTENDR_SKIP_DEV_TESTS: TRUE
 
     # PowerShell core is available on all platforms and can be used to unify scripts
     defaults:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -239,7 +239,7 @@ jobs:
                 "\" }"),
 
               # uncomment this line when we need to depend on the dev version of libR-sys
-              'libR-sys = { git = "https://github.com/extendr/libR-sys", features = array("non-api", 1) }',
+              # 'libR-sys = { git = "https://github.com/extendr/libR-sys", features = array("non-api", 1) }',
 
               sep = ";")
           
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -238,7 +238,7 @@ jobs:
                 "\" }"),
 
               # uncomment this line when we need to depend on the dev version of libR-sys
-              'libR-sys = { git = "https://github.com/extendr/libR-sys", features = ["non-api"] }',
+              'libR-sys = { git = "https://github.com/extendr/libR-sys", features = list("non-api") }',
 
               sep = ";")
           
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -247,7 +247,6 @@ jobs:
 
           github_env_file <- Sys.getenv("GITHUB_ENV")
           cat(patch.crates_io, file = github_env_file, append = TRUE)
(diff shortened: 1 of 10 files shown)


Ilia-Kosenkov:
Bindings are broken?

CGMossa:
Yes!
We made this decision the other day here https://github.com/extendr/libR-sys/pull/236

The `non-api` feature can make things resolve for now, but I need to find a way to replace those missing API bits in extendr.

Ilia-Kosenkov:
After a conversation in DMs, I believe these errors arise from functions that should not be part of public API by default, nor should they be not `unsafe`. They are not used in the project at all, perhaps they are used downstream. To address this, I suggest to hide them behind `non-api` feature gate. When `non-api` is enabled, both `libR-sys` and `extendr-api` will behave as before. When it is disabled, `extendr-api` will be safer and more lightweight. Also let's mark them `unsafe` since they deal with raw things.
And add an entry in docs explaining why this decision is made and what to expect from feature flag. Thanks!

extendr-api/src/wrapper/environment.rs
```diff
@@ -97,6 +98,7 @@ impl Environment {
         }
     }
 
+    #[cfg(feature = "non-api")]
     /// Set the environment flags.
     pub fn set_envflags(&mut self, flags: i32) -> &mut Self {
```
Ilia-Kosenkov, line 103: I'd make this unsafe for sure, it accepts flags as `i32`.

extendr-api/tests/debug_tests.rs
```diff
@@ -19,6 +19,7 @@ fn test_debug() {
         assert_eq!(format!("{:?}", r), "base_env()");
         let r = Environment::new_with_parent(global_env());
         assert_eq!(format!("{:?}", r), "<environment>");
+        #[cfg(feature = "non-api")]
         let r = Promise::from_parts(r!(1), global_env())?;
         assert_eq!(format!("{:?}", r), "Promise { code: 1, environment: global_env() }");
```
Ilia-Kosenkov, line 24: You need feature gate on assert too I think

.github/workflows/test.yml
```diff
@@ -238,7 +238,7 @@ jobs:
                 "\" }"),
 
               # uncomment this line when we need to depend on the dev version of libR-sys
-              'libR-sys = { git = "https://github.com/extendr/libR-sys" }',
+              'libR-sys = { git = "https://github.com/extendr/libR-sys", features = array("non-api", 1) }',
 
               sep = ";")
```
Ilia-Kosenkov, line 244: Set here also `REXTENDR_SKIP_DEV_TESTS=TRUE`. Once https://github.com/extendr/rextendr/pull/352 is merged, it should work.

Ilia-Kosenkov:
I think we screwed up appending env variable, can you please set it using `env:` section of yaml file?

.github/workflows/test.yml
```diff
@@ -286,6 +287,7 @@ jobs:
 
     env:
       R_REMOTES_NO_ERRORS_FROM_WARNINGS: true
+      REXTENDR_SKIP_DEV_TESTS: TRUE
```
Ilia-Kosenkov, line 290: Not needed here, we do not run rextendr tests with bindgen

.github/workflows/test.yml
```diff
@@ -238,14 +239,15 @@ jobs:
...
-              'libR-sys = { git = "https://github.com/extendr/libR-sys" }',
+              # 'libR-sys = { git = "https://github.com/extendr/libR-sys", features = array("non-api", 1) }',
 
               sep = ";")
           
           patch.crates_io <- paste0("REXTENDR_PATCH_CRATES_IO=", patch.crates_io)
 
           github_env_file <- Sys.getenv("GITHUB_ENV")
           cat(patch.crates_io, file = github_env_file, append = TRUE)
+          cat("REXTENDR_SKIP_DEV_TESTS=TRUE", file = github_env_file, append = TRUE)
```
Ilia-Kosenkov, line 250: You need to remove this

Ilia-Kosenkov (approved):

PR #757, opened 2024-05-12
Fix integerish conversions

### LINKED ISSUE #720: Parsing f64 to i32 where the number overflows i32 causes to get the max value instead NA or overflow
Hi all!, I found this by accident, but is not good...:

```R
library(rextendr)
code <- r"(
#[extendr]
fn hello(x: i32) {
    println!("{x}");
}
)"

rust_source(code = code)

hello(8314042901901001)
2147483647
```

No idea if other type has this behavior, but I think this should never happens.

Thx!

Ilia-Kosenkov:
Fixes #720.
- Rework scalar `f64` (`Rfloat`) to integer type conversions (both `FromRobj` and `TryFrom`) by extracting all logic into a single `FloatToInt` trait. 
- Properly handle all edge cases like `NA`, over-/under-flow, subnormality
- Disallow rounding casts that result in loss of accuracy (e.g., `4.2f64` no longer maps to `4i32`)
- Allow only `f64` values that are exactly representable by target integer type

Also update `xtask`:
- `cargo extendr devtools-test` now gained `-f`/`--filter` option that controls which tests to execute. Value is passed verbatim to `filter` parameter of `devtools::test()`


extendr-api/src/conversions/try_into_int.rs
```diff
@@ -0,0 +1,148 @@
...
+    fn try_into_int(&self) -> Result<T, ConversionError>
+    where
+        Self: Sized;
+}
+
+macro_rules! impl_into_integerish {
+    ($float_type:ty, $int_type:ty) => {
+        impl FloatToInt<$int_type> for $float_type {
+            fn try_into_int(&self) -> Result<$int_type, ConversionError> {
+                if *self == <$float_type>::default() {
```
CGMossa, line 30: Might be an idea to use is zero function or something like that.

Ilia-Kosenkov: I believe this is the easiest way for us to check for exact `0.0` and return `0` afterward. Without involving external traits.

CGMossa: Is_zero doesn't exists. Lol. Probably num-traits thing

CHANGELOG.md
```diff
@@ -19,11 +19,15 @@
 - Potentially breaking: `RArray::from_parts` no longer requires a pointer to the underlying data vector [[#657]](https://github.com/extendr/extendr/pull/657)
 - Breaking change: In `AltRep` the `unserialize_ex`, `set_parent`, `set_envflags` are
 now hidden behind the feature flag `non-api`. Also, `Promise::from_parts` is marked as non-API.
+- Floating point numbers with decimal part can no longer be converted to integer types via
```
CGMossa, line 22: I'd call this a breaking change.

CGMossa (approved):
Lgtm.

Personally, I didn't like the way it was done before.

We should strive for a performant conversion, if you ask me, but that's another discussion altogether.
I believe if the num crate does this in a more clever way, we ought to consider it.

Bytemuck does not do conversions. I checked. It merely reinterprets bytes.

CGMossa:
Please consider this https://doc.rust-lang.org/std/primitive.f64.html#method.to_int_unchecked
Before merging.

CHANGELOG.md
Ilia-Kosenkov: It certainly is. What do you want me to do?

CGMossa: Nothing! We starting prefacing these things with breaking change 🤷‍♂️

Ilia-Kosenkov:
It practically does what `f64 as i32` does, with perhaps tiny variations. It still will overflow/underflow to max/min value (what we observed in the original issue), it will break on various `NaN`, it will not propagate `NA`, which we definitely want. So it is practically useless to us. We are looking for a `_checked` version, which is not yet present.

Since I extracted the logic as much as I could, if we find a better solution to map an `f64` into `i32 | ConversionError`, it is now easy to implement, replacing code in one place.

CHANGELOG.md
Ilia-Kosenkov: What would you want me to do?

CGMossa: Write: Breaking Change:

Ilia-Kosenkov: Sorry, sent one message from the phone and another from my PC. Let me fix it

PR #758, opened 2024-05-13
Convert integer-ish values based on `f64::classify()`

Ilia-Kosenkov:
Switch to `f64::classify()` to determine what type of float we are dealing with. Simplifies handling of `NaN`, subnormals, both `-0.0` and `0.0`, as well as infinities.

### DIFF
--- a/extendr-api/src/conversions/try_into_int.rs
+++ b/extendr-api/src/conversions/try_into_int.rs
@@ -1,4 +1,5 @@
 use std::fmt::{Debug, Display, Formatter};
+use std::num::FpCategory;
 
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub enum ConversionError {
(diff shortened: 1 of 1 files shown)


extendr-api/src/conversions/try_into_int.rs
```diff
@@ -27,25 +28,27 @@ macro_rules! impl_into_integerish {
     ($float_type:ty, $int_type:ty) => {
         impl FloatToInt<$int_type> for $float_type {
             fn try_into_int(&self) -> Result<$int_type, ConversionError> {
-                if *self == <$float_type>::default() {
-                    return Ok(<$int_type>::default());
+                match self.classify() {
+                    FpCategory::Nan | FpCategory::Subnormal => Err(ConversionError::NotIntegerish),
```
CGMossa, line 32: I think this is correct. Can you explain it to me though?

extendr-api/src/conversions/try_into_int.rs
```diff
@@ -27,25 +28,27 @@ macro_rules! impl_into_integerish {
...
+                    FpCategory::Normal => {
+                        let truncated_value = self.trunc();
+                        if truncated_value < <$int_type>::MIN as $float_type {
+                            return Err(ConversionError::Underflow);
+                        }
+                        if truncated_value > <$int_type>::MAX as $float_type {
+                            return Err(ConversionError::Overflow);
+                        }
+                        if !truncated_value.eq(self) {
+                            return Err(ConversionError::NotIntegerish);
```
CGMossa, line 47: I also want to ask this.. Why don't we `.clamp` it, and say that the conversion is following the Saturating mode, of https://doc.rust-lang.org/std/num/struct.Saturating.html ?

I'm not saying anything against what you've done. I just somehow, sometimes, feel like we have too many errors. Like just give the user _something_ even if the request was absurd.

Ilia-Kosenkov: Because it will revert to precious behavior -- silently changing values https://github.com/extendr/extendr/issues/720#issue-2221351632

extendr-api/src/conversions/try_into_int.rs
Ilia-Kosenkov: Neither `NaN` nor a subnormal float are representable by integers (unless you want to default to `0`). It is quote challenging to get a subnormal float manually, so if we get one, I prefer to fail. 
Don't forget, these conversions are designed to handle things like indices, where people call it as `my_func(x, 4)`. And `4` is actual `4.0f64` and we want to cast it to `4i32` or `4usize`. Every other case should be a hard error

CGMossa: Alright. Good enough for me, and for now

extendr-api/src/conversions/try_into_int.rs
CGMossa: > Because it will revert to precious behavior -- silently changing values [#720 (comment)](https://github.com/extendr/extendr/issues/720#issue-2221351632)

But isn't that _fine_?

Ilia-Kosenkov: I would say absolutely not, one of the reasons people criticize `as` operators in Rust because of this sh1t. If you pass stupidly large value that does not fit in your target type, this is an error, not `i32::MAX`

Ilia-Kosenkov:
Looks interesting, but I cannot apply this right now. Will check in the evening

CGMossa:
> Switch to `f64::classify()` to determine what type of float we are dealing with. Simplifies handling of `NaN`, subnormals, both `-0.0` and `0.0`, as well as infinities.

It also accurately, and efficiently compares the `0` given by R. I think that matters a lot.

CGMossa:
Also, while we are already on this topic, can't we have conversion to `usize`?

Ilia-Kosenkov:
`usize` conversion is supported

CGMossa (approved):

PR #761, opened 2024-05-14
Faster string conversion

CGMossa:
This PR is meant to improve how we convert character-vectors to rust strings.

First, observe that
- Length of `STRSXP` is the number of embedded `CHARSXP` in the character-vector
- Length of `CHARSXP` are the number of characters in a string. 

Then, let's first look at `to_str`. This function scans a C-string until it meets `\nul`, which
is then what determines the length of the string. Thus, for every string, we iterate through all the
bytes, until we find `0`/`\nul`. This is unnecessary. Then, there is the wonky conversion `sexp_to_str`.
Removing those two is what leads to the rest of the PR.

Every symbol is a permanent object, that contains a protected string, and `PRINT_NAME` refers to the underlying `CHARSXP`. 

- [ ] Remove all the `assert_eq!` for checking if the string is a `CHARSXP`. I don't think it is worth it.

This PR was largely motivated by this post https://github.com/extendr/extendr/pull/646#issuecomment-1752582837.


extendr-api/src/functions.rs
```diff
@@ -131,7 +132,7 @@ pub fn new_env(parent: Environment, hash: bool, capacity: i32) -> Environment {
         .unwrap()
 }
 
-/// The base environment; formerly R_NilValue
+/// The base environment; formerly `R_NilValue`
```
JosiahParry, line 135: Is this formerly or formally?

CGMossa: You know what? I think it is formerly. And _actually_ I think something weird happens when you use `NilValue` as environmnet.

CGMossa:
I've reduced, refactored, trimmed, and commented. I think this is one of the few PRs with a lot of improvements, for very few changes. But these changes are subtle.
I'm going to mark it ready for review now.

extendr-api/src/iter.rs
```diff
@@ -55,22 +55,16 @@ impl StrIter {
     }
 }
 
-// Get a string reference from a CHARSXP
-fn str_from_strsxp<'a>(sexp: SEXP, index: isize) -> &'a str {
+// Get a string reference from a `CHARSXP`
+pub(crate) fn str_from_strsxp<'a>(sexp: SEXP, index: isize) -> &'a str {
     single_threaded(|| unsafe {
-        if index < 0 || index >= Rf_xlength(sexp) {
+        let charsxp = STRING_ELT(sexp, index);
```
Ilia-Kosenkov, line 61: Why did you remove bounds check? This is C-level vulnerability

CGMossa: Good question. First, `STRING_ELT` does detection on its own. Secondly, we control the state (and especially initial state) of the iterators, and we should really just define how they are instantiated correctly, instead of having checks everywhere.

I just checked, because I forgot this.. `STRING_ELT` return a `R_NilValue` if you access something outside the bounds. I think we should just handle that case and not have range checks everywhere, especially since R does it already.

CGMossa: I've added and changed things. I think this works now. But I have a problem with the code organisation... maybe you can suggest a good order (wrapping into functions)?

Ilia-Kosenkov: Do you have a source code/document reference for that? Because if you actually _checked_ it, you could have just read zeroed memory, which was interpreted as `NULL` or something like that

extendr-api/src/iter.rs
```diff
@@ -55,22 +55,18 @@ impl StrIter {
...
 
-// Get a string reference from a CHARSXP
-fn str_from_strsxp<'a>(sexp: SEXP, index: isize) -> &'a str {
+// Get a string reference from a `CHARSXP`
+pub(crate) fn str_from_strsxp<'a>(sexp: SEXP, index: usize) -> &'a str {
     single_threaded(|| unsafe {
-        if index < 0 || index >= Rf_xlength(sexp) {
+        let charsxp = STRING_ELT(sexp, index as _);
+        if charsxp == R_NilValue {
+            panic!("out of bound error")
```
Ilia-Kosenkov, line 63:
```suggestion
            panic!("Index out of bounds error")
```

extendr-api/src/iter.rs
CGMossa: I looked at the source code. I have to do that, because nothing is documented. And it checked by these intricate/altrep secure ways.. But the think is, you get `NilValue` back. Which is distinctly different than C's `NULL` pointer. I don't get C `NULL`. I get `R_NilValue` aka R `NULL`.

extendr-api/src/iter.rs
```diff
@@ -89,14 +85,26 @@ impl Iterator for StrIter {
...
+            } else if TYPEOF(vector) == SEXPTYPE::STRSXP {
+                Some(str_from_strsxp(vector, i))
+            } else if TYPEOF(vector) == SEXPTYPE::CHARSXP {
+                if vector == R_NaString {
+                    Some(<&str>::na())
+                } else if vector == R_BlankString {
+                    Some("")
+                } else {
+                    Some(rstr::charsxp_to_str(vector))
+                }
```
Ilia-Kosenkov, line 99: This fragment is duplicated

extendr-api/src/iter.rs
CGMossa: Thanks! I'm going to remove this part, because it is stupid to have a panic here. I just need to think a bit clearer.. I'll ping when I've restructured it.

extendr-api/src/iter.rs
CGMossa: You're right. I'll move things around and ping later.

extendr-api/src/iter.rs
```diff
@@ -89,14 +85,26 @@ impl Iterator for StrIter {
...
+            } else if Rf_isFactor(vector).into() {
+                // factor support: factor is an integer, and we need
+                // the value of it, to retrieve the assigned label
+                let j = *(INTEGER(vector).add(i));
```
Ilia-Kosenkov, line 103: Can't we do some sane transmute into `&[i32]` and then use indexer? This aint no C language! (I know it is written like this originally, but we can do better). This will probably cost us a bounds check at runtime, but I believe it is worth it for the readability sake. What do you think?

CGMossa: You're absolutely right. I'm looking at it right now. There is an INTEGER_ELT... like bro, we should maybe use that.

extendr-api/src/iter.rs
```diff
@@ -178,34 +186,61 @@ pub trait AsStrIter: GetSexp + Types + Length + Attributes + Rinternals {
     fn as_str_iter(&self) -> Option<StrIter> {
         let i = 0;
         let len = self.len();
-        match self.sexptype() {
-            SEXPTYPE::STRSXP => unsafe {
+        if self.sexptype() == SEXPTYPE::STRSXP {
```
Ilia-Kosenkov, line 189: I would say this implementation (though it already was written like this is flawed). We use the same iterator to process three different types only to 'differentiate' between them at the element access level. It should be something like `Option<Box<dyn StrIter>>`. But maybe not for this PR

CGMossa: Yes.. You're right. So what I thought was.. the factor part should be its own thing anyways, and started a draft for that here #753. But you're right. This should happen at the instance level, and maybe be split by enum or something like that. I'll give it a go in another PR, for now, let us clean this up and take the win from the fast conversion of `CHARSXP` afforded.

I'd appreciate an issue on it.

CGMossa:
I've moved, refactored, deleted things. This should be it at this point. 
Only thing maybe, is that `str_from_strsexp` is now only doing one thing, and maybe that should be done the two places where `str_from_strsexp` is being called.

extendr-api/src/iter.rs
```diff
@@ -89,14 +77,23 @@ impl Iterator for StrIter {
...
+            } else if TYPEOF(vector) == SEXPTYPE::STRSXP {
+                str_from_strsxp(vector, i)
+            } else if TYPEOF(vector) == SEXPTYPE::CHARSXP {
+                rstr::charsxp_to_str(vector)
+            } else if Rf_isFactor(vector).into() {
+                // factor support: factor is an integer, and we need
+                // the value of it, to retrieve the assigned label
+                let j = std::slice::from_raw_parts(INTEGER(vector), self.len as _);
+                let j = j.get(i)?;
+                // assert_eq!(TYPEOF(self.levels), STRSXP, "levels of a factor must always be a character-vector");
```
Ilia-Kosenkov, line 91: Drop commented-out code

extendr-api/src/iter.rs
```diff
@@ -89,14 +77,23 @@ impl Iterator for StrIter {
...
+                // assert_ne!(j, 0, "invalid factor, where level/label i 0-indexed");
+                let j = j
+                    .checked_sub(1)
+                    .expect("the factor integer had an invalid value in it");
```
Ilia-Kosenkov, line 95:
```suggestion
                    .expect("The factor integer has an invalid value in it.");
```

extendr-api/src/robj/mod.rs
```diff
@@ -352,6 +352,9 @@ impl Robj {
                 let sexp = self.get();
                 use SEXPTYPE::*;
                 match self.sexptype() {
+                    // a character vector contains `CHARSXP`, and thus you
+                    // seldomly have `Robj`'s that are `CHARSXP` themselves
+                    CHARSXP => sexp == libR_sys::R_NaString,
```
Ilia-Kosenkov, line 357: If `match` branches precedence is important and `CHARSXP` is rare, move it to the end of the clause?

CGMossa:
@Ilia-Kosenkov thank you for the second round! I've pushed the suggested changes.

extendr-api/src/iter.rs
```diff
@@ -89,14 +77,21 @@ impl Iterator for StrIter {
...
-                Some(<&str>::na())
+                None
+            } else if TYPEOF(vector) == SEXPTYPE::STRSXP {
+                str_from_strsxp(vector, i)
+            } else if TYPEOF(vector) == SEXPTYPE::CHARSXP {
+                rstr::charsxp_to_str(vector)
+            } else if Rf_isFactor(vector).into() {
+                // factor support: factor is an integer, and we need
+                // the value of it, to retrieve the assigned label
+                let j = std::slice::from_raw_parts(INTEGER(vector), self.len as _);
```
Ilia-Kosenkov, line 89: Can we have a meaningful name for this? It is not a trivial loop index.

extendr-api/src/robj/mod.rs
```diff
@@ -357,6 +357,9 @@ impl Robj {
                     LGLSXP => *(LOGICAL(sexp)) == libR_sys::R_NaInt,
                     REALSXP => R_IsNA(*(REAL(sexp))) != 0,
                     CPLXSXP => R_IsNA((*COMPLEX(sexp)).r) != 0,
+                    // a character vector contains `CHARSXP`, and thus you
+                    // seldomly have `Robj`'s that are `CHARSXP` themselves
```
Ilia-Kosenkov, line 361:
```suggestion
                    // seldom have `Robj`'s that are `CHARSXP` themselves
```

extendr-api/src/robj/mod.rs
```diff
@@ -589,21 +592,22 @@ impl Robj {
     ///    assert_eq!(robj1.as_str(), Some("xyz"));
     ///    assert_eq!(robj2.as_str(), None);
     /// }
-    /// ```
+    /// ```
```
Ilia-Kosenkov, line 595:
```suggestion
    /// ```
```

extendr-api/src/wrapper/rstr.rs
```diff
@@ -17,11 +17,22 @@ pub struct Rstr {
     pub(crate) robj: Robj,
 }
 
-pub(crate) unsafe fn sexp_to_str(sexp: SEXP) -> &'static str {
-    if sexp == R_NaString {
-        <&str>::na()
+/// Returns a rust string-slice based on the provided `SEXP`, who is of type
```
Ilia-Kosenkov, line 20:
```suggestion
/// Returns a rust string-slice based on the provided `SEXP`, which is of type
```

Ilia-Kosenkov (changes requested):
👍

extendr-api/src/iter.rs
CGMossa: I've renamed it to `level_index`.

Ilia-Kosenkov (approved):

PR #762, opened 2024-05-15
[urgent!] `#[extendr]`-impl requires `libR-sys`

CGMossa:
Currently on `master`,  `#[extendr]`-impl requires that the user have `libR-sys` as a dependency to compile. **This is a big problem**

The culprit is the extraction of a reference / mutable reference from `Robj`. We use something
from `libR-sys` directly, thus the user needs to have `libR-sys` available.
We do not export `libR-sys`, so we can't depend on it by way of `extendr-api`. Meaning, there is no `pub use libR_sys;` in `extendr-api`. I don't think we should do that either.

Instead, I think we should unify the interface of `#[extendr]`-impl to fully embrace `ExternalPtr<T>`.

This is what this PR is about!

## Details

I take a `&Robj`, turn that (unsafely) to `&ExternalPtr<T>`, and from there I extract (unsafely) the `&T`.
Similarly, this is done for `&mut T`.
To ensure that this is okay, we have to make `ExternalPtr<T>` transparent.

**Going from `&Robj` to `ExternalPtr<T>`**: The `Robj` could be something else than an `externalptr`. One could have put in a `list()` instead. Thus, we make a type-check on `externalptr`-type.

**Going from `&ExternalPtr<T>` to `&T`**: The pointer may be `NULL`.

All of these conditions has been unified in an error-aware chain of calls. All the low-level pointer
manipulations has been replaced by Rust approved abstraction.
_There is one more abstraction that we can employ, the `NonNull<T>` one.. But let us wait on that for another PR.

Finally, the finalizer has been removed, because we can rely on `ExternalPtr<T>`, and its clean way of doing this. To that end, `ExternalPtr<T>` was missing an `Attributes`-impl.

## Conclusion

Actually there is way more: We urgently need to approve this, but what is more important,
is to remove the dependency of `libR-sys` from our integration test crate `tests/extendrtests/src/rust`. This can be done, but it is going to be a bit contentious. I'll do that in a follow-up PR.

### DIFF
--- a/extendr-api/src/wrapper/externalptr.rs
+++ b/extendr-api/src/wrapper/externalptr.rs
@@ -24,7 +24,7 @@ use std::fmt::Debug;
 ///     assert_eq!(*extptr2, 1);
 /// }
 /// ```
-///
+#[repr(transparent)]
 #[derive(PartialEq, Clone)]
 pub struct ExternalPtr<T> {
     /// This is the contained Robj.
@@ -59,6 +59,9 @@ impl<T> Length for ExternalPtr<T> {}
 
 /// rtype() and rany()
 impl<T> Types for ExternalPtr<T> {}
+
+/// `set_attrib`
+impl<T> Attributes for ExternalPtr<T> {}
 
 /// as_*()
 impl<T> Conversions for ExternalPtr<T> {}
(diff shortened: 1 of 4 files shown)


Ilia-Kosenkov (approved):
Approved with permission to merge bypassing CI checks since they fail on third party dependency regression & main test suite is all green.

PR #763, opened 2024-05-17
Fix incorrect casting in `AsTypedSlice`...

CGMossa:
I recently learned that casting a pointer, is akin to reinterpreting bytes, and not actual conversion.

This means that our `AsTypedSlice` for `u32` is technically invalid.

```rs
make_typed_slice!(u32, INTEGER, INTSXP);
```
To showcase this

```rs
// Given 
[-32, 3, i32::MAX]
// becomes
    [
        4294967264,
        3,
        2147483647,
    ],
```

For now, I've added a small extension to the macro. But this whole bit could use a refactor. I'm on it..

(ignore branch name)

### DIFF
--- a/extendr-api/src/prelude.rs
+++ b/extendr-api/src/prelude.rs
@@ -77,4 +77,3 @@ pub use ::either::*;
 
 #[cfg(feature = "faer")]
 pub use ::faer::*;
-


CGMossa:
While this work for reading, it doesn't work for writing. What if we write a `u32`, that don't fit within `i32`? That would be wonky. It is probably best to remove this, or come up with a solution, where references are made, but not _mutable references_.

PR #765, opened 2024-05-20
STANDALONE: `ExternalPtr` is a pointer

CGMossa:
One cannot `PartialEq` or `Clone` the `ExternalPtr<T>` in the traditional sense, by relying on the 
trait on `T`. Instead, we treat `ExternalPtr` like a pointer. This behaviour is now documented, and the
alternative way to perform comparison is also mentioned.

### DIFF
--- a/extendr-api/src/wrapper/externalptr.rs
+++ b/extendr-api/src/wrapper/externalptr.rs
@@ -9,6 +9,14 @@
 //! Neither `tag` nor `prot` are attributes, therefore to use `ExternalPtr` as
 //! a class in R, you must decorate it with a class-attribute manually.
 //!
+//! **Beware**: Equality (by way of `PartialEq`) does not imply equality of value,
+//! but equality of pointer. Two objects stored as `ExternalPtr` may be equal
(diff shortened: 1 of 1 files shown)


extendr-api/src/wrapper/externalptr.rs
```diff
@@ -260,3 +285,73 @@ impl<T: Debug> std::fmt::Debug for ExternalPtr<T> {
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use extendr_engine::with_r;
+
+    #[test]
+    fn partial_eq_of_externalptr() {
```
Ilia-Kosenkov, line 307: I'd prefer to have this split into a couple more specific tests. Or at least into 'these should be equal', 'these should be unequal', etc. Then if one test fails, it will be easier to understand what is going on.

CGMossa: I've tried to split it. Can you look at it again plase?

Ilia-Kosenkov (approved):

PR #767, opened 2024-05-20
bug: removed `&[u32]` interpretation of `Robj` 

CGMossa:
In `AsTypedSlice`, we assume that one may convert a `&[i32]` to a `&[u32]`. That works, unless any of the
`i32` being _reinterpreted_ are negative values. This was done in #763, but then a _mutable slice_ would not work either.
For now, we have to remove these, and only adopt them, when we can only implement a reference access, and said
reference access may only work on non-negative slices.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,13 +21,17 @@
   place, and then return an untyped owned pointer (Robj). Instead, now we return `&mut Self`.
 - [_Potentially breaking_]: `RArray::from_parts` no longer requires a pointer to the underlying data
   vector [[#657]](https://github.com/extendr/extendr/pull/657)
-- [**Breaking change**]: In `AltRep` the `unserialize_ex`, `set_parent`, `set_envflags` are
-now hidden behind the feature flag `non-api`. Also, `Promise::from_parts` is marked as non-API.
-- [**Breaking change**]: Floating point numbers with decimal part can no longer be converted to integer types via
-  rounding [[#757]](https://github.com/extendr/extendr/pull/757)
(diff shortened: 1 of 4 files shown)


Ilia-Kosenkov:
The intention was to exactly support that -- reinterpreting memory without copying. I would not say it is a problem, because accepting `&[i32]` already means you sacrifice `NA` checks, so if you go for `&[u32]`, you should know what you do.

CGMossa:
> Member

It is not just NA checks. If you insert a value larger than `2_147_483_647` (`i32::max`) you'll get problems.

Ilia-Kosenkov:
Yes but one can argue that this can be an acceptable tradeoff. But we can ditch it if you insist

CGMossa:
Sure, but I believe the trade-off should be apparent from the API. I'd like to close these "unsafeness", and then propose an API that permits these sort of actions, but it is more clear, that the action is 100% safe.
Like, I was fooled by it, :P

Ilia-Kosenkov (changes requested):
Write this down in the changelog.

CGMossa:
Dear @Ilia-Kosenkov, I must apologise, as sometimes, I forget my own intentions. 

I believe we **either** accept this PR, or we accept https://github.com/extendr/extendr/pull/768.
In #768, I have a check on the read slice, as to only return non-negative values as `u32`, and I block the ability
to produce `&mut [u32]`, which is the issue I've complained about here.

Then, one should notice, that in our `ndarray`, we only provide non-read Views, and not mutable views. Therefore, there is no need to being pedantic about this (if #768 is merged that is) in feature `ndarray`.

Ilia-Kosenkov (approved):

PR #768, opened 2024-05-20
fix: Added range-checking for integer slices

CGMossa:
Second attempt at dealing with the issue brought up in #763, however, this time, we actually remove the ability
to write into unsigned integer slices, as they won't fit in R.

### DIFF
--- a/extendr-api/src/robj/mod.rs
+++ b/extendr-api/src/robj/mod.rs
@@ -803,41 +803,31 @@ make_typed_slice!(c64, COMPLEX, CPLXSXP);
 make_typed_slice!(Rcplx, COMPLEX, CPLXSXP);
 make_typed_slice!(Rcomplex, COMPLEX, CPLXSXP);
 
-macro_rules! make_typed_slice_unsigned_int {
-    ($type: ty, $fn: tt, $($sexp: tt),* ) => {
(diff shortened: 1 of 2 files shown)


extendr-api/src/robj/mod.rs
```diff
@@ -804,6 +803,42 @@ make_typed_slice!(c64, COMPLEX, CPLXSXP);
...
+            fn as_typed_slice_mut(&mut self) -> Option<&'a mut [$type]> {
+                None
+            }
+        }
+    }
+}
+
+make_typed_slice_unsigned_int!(u32, INTEGER, INTSXP);
+make_typed_slice_unsigned_int!(u64, INTEGER, INTSXP);
+make_typed_slice_unsigned_int!(u128, INTEGER, INTSXP);
```
CGMossa, line 840: This can't be correct. There is not enough bytes in i32 slice for these to begin with. I think this should be removed.

Ilia-Kosenkov:
I am not in favor of this since it defeats the purpose of having 'almost' safe slices. No extra checks apart from `SEXPTYPE`, which already happens.

CGMossa:
Agreed. I think there is a place for such a conversion, but not embedded as an incidental off-shoot in `AsTypedSlice`.

PR #771, opened 2024-05-22
`Rstr` as `ToVectorValue`

### LINKED ISSUE #770: `ToVectorValue` implemented for `&str` and `String` but not `Rstr`
Hello,

I'm trying to construct a character/string `RMatrix`, however due to trait bounds this doesn't seem to be possible. In order to construct a matrix, it needs to satisfy `Robj: AsTypedSlice<'a, T>` and `T: ToVectorValue`. `AsTypedSlice` is implemented only for `Rstr`, which I understand, but `ToVectorValue` is not implemented for `Rstr`, but is for `&str` and `String`, thereby making it impossible to construct a string matrix.

Would it be possible to implement it for `Rstr`? Given there's implementations for the other string types, I don't see any reason since `Rstr` stores a `SEXP` internally already and that's all `ToVectorValue` seems to require.

Thank you!

CGMossa:
Fixes #770 

- [ ] All `elt`, `set_elt`, even `index` (trait and function) need to be adjusted, as
extracting an element from a stringly-vector requires `STRING_ELT` and `SET_STRING_ELT`. This might cause a larger rewriting..

### DIFF
--- a/extendr-api/src/robj/into_robj/repeat_into_robj.rs
+++ b/extendr-api/src/robj/into_robj/repeat_into_robj.rs
@@ -86,3 +86,49 @@ impl_synonym_type!(Rfloat, f64);
 impl_synonym_type!(&Rfloat, f64);
 impl_synonym_type!(Rint, i32);
 impl_synonym_type!(&Rint, i32);
+
+// since `Rstr` does not implement `Scalar` (due to lifetime conflicts), 
+// it requires manual implementation
+
+impl ToVectorValue for Rstr {
+    fn sexptype() -> SEXPTYPE {
+        SEXPTYPE::STRSXP
+    }
+
+    fn to_sexp(&self) -> SEXP
+    where
+        Self: Sized,
+    {
+        unsafe { self.robj.get() }
+    }
+}
+
+impl ToVectorValue for &Rstr {
+    fn sexptype() -> SEXPTYPE {
+        SEXPTYPE::STRSXP
+    }
+
+    fn to_sexp(&self) -> SEXP
+    where
+        Self: Sized,
+    {
+        unsafe { self.robj.get() }
+    }
+}
+
+impl ToVectorValue for Option<Rstr> {
+    fn sexptype() -> SEXPTYPE {
+        SEXPTYPE::STRSXP
+    }
+
+    fn to_sexp(&self) -> SEXP
+    where
+        Self: Sized,
+    {
+        if let Some(s) = self {
+            unsafe { s.robj.get() }
+        } else {
+            unsafe { R_NaString }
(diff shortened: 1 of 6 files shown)


CGMossa:
I've run into a major design conflict, that I may be able to describe here: Say you've got `RMatrix<T>`. For `T = Rfloat` or even `T = f64`, what we are really saying here, is that the `Rmatrix<f64>` is really a `Doubles` with extra attributes. For `Doubles`, I've defined `Elt<f64>` and `Elt<Rfloat>`. Good! But then how can an `RMatrix<T>` know that it is really a dressed-up `Doubles`? That's not something I've been easily able to inject in the traits system.

CGMossa:
I've made progress on this PR. As now, I merely hand-implement the vector-types to be assumed. Notice how the `transmute` call looks like: Due to `RArray` containing a field more than one (`Robj`), I have to make a subset-type by a tuple-construction.

I've left `TODO`/`FIXME` throughout. I'd be happy with some assistance on this.

(ignore CI failures, as they are related to `faer`)

extendr-api/src/wrapper/mod.rs
```diff
@@ -238,6 +238,11 @@ make_getsexp!(Dataframe<T>, impl<T>);
 //     }
 // }
 
+pub trait Elt<T> {
+    fn elt(&self, index: usize) -> T;
```
JosiahParry, line 242: typically I see these methods more or less like `get()` and `get_unchecked()` do we want to support this where `elt()` should return `Result<T>` and `elt_unchecked()` would return `T`

extendr-api/src/robj/into_robj/repeat_into_robj.rs
```diff
@@ -86,3 +86,49 @@ impl_synonym_type!(Rfloat, f64);
...
+    where
+        Self: Sized,
+    {
+        if let Some(s) = self {
+            unsafe { s.robj.get() }
+        } else {
+            unsafe { R_NaString }
+        }
+    }
+}
```
JosiahParry, line 134: I think this alone can be merged. The rest of the PR is great, but would need a lot of discussion I think. Should we port this over to a smaller PR ?

CGMossa: I don't think this is correct, unfortunately. It looks useful, but I think it is flawed. We need to brainstorm about this more before I can fully commit to an explaination or a strategy on this.

Basically, the `ToVectorValue` stuff assumes certain things about the extendr API, and these assumptions are not complete.. Sorry, I need to write a more thorough explaination of the issues here.. and somewhere more central.

PR #772, opened 2024-05-22
STANDALONE: Removing unnecessary clones

CGMossa:
These conversions are defined on `&Robj` already, so we don't need to clone anyways.


Ilia-Kosenkov (approved):
This always confused me why we needed an extra clone here, if this does not leak anything, than ok.

PR #773, opened 2024-05-27
Error handling, unwinding, unhelpful defaults

CGMossa:


CGMossa:
We essentially have the same code for this, although less flexible, in extendr already (`catch_r_error`). 
The advice from https://yutani.rbind.io/post/dont-panic-we-can-unwind/ is to add the `R_ContinueUnwind` call in the actual C-wrapper, and not the rust code. This might be neccessary, or not, in either case, this will require more involved changes. I'll investigate that later.

extendr-api/src/ownership.rs
```diff
@@ -267,6 +267,16 @@ impl Ownership {
     }
 }
 
+impl Ownership {
+    /// Return the total number of currently protected elements, i.e. those with
+    /// non-zero reference count.
+    pub fn total_protected(&self) -> usize {
+        self.objects
+            .values()
+            .fold(0, |acc, x| if x.refcount != 0 { acc } else { acc + 1 })
```
Ilia-Kosenkov, line 276: I think this is wrong, you count items with zero references, no? Also you can just do filter + count, much more readable.

CGMossa: Nice you're looking at this! I need all the feedback I can receive for this task...

I wrote this to have something in the test... But I realize now, that I'm not actually using this in the test, because the 
test does not use the `ownership` module. So right now, this is actually very redundant.

Next, what's the logic here? I want to count the total number of protected items. A protected item, is an item, who has a reference count that is non-zero, because that means, it is being actively protected. If the reference count is non-zero, then I need to count it, and if the reference count is exactly zero, then I should disregard it.

Well, that is the opposite of what I wrote up there. So thank you for the tip!

PR #776, opened 2024-06-01
Update Makevars for integration test

CGMossa:
We found this file that describes how the linking should look like [`.mk` file for rust compiler tests](https://github.com/rust-lang/rust/blob/bf8fff783ff533c055d276378ada30563312def1/tests/run-make/tools.mk). 

There are other hints in this file, as to what to include, with other platforms.
Hopefully we can include those too..

### DIFF
--- a/tests/extendrtests/src/Makevars.win
+++ b/tests/extendrtests/src/Makevars.win
@@ -3,7 +3,7 @@ TARGET = $(subst 64,x86_64,$(subst 32,i686,$(WIN)))-pc-windows-gnu
 TARGET_DIR = ./rust/target
 LIBDIR = $(TARGET_DIR)/$(TARGET)/debug
 STATLIB = $(LIBDIR)/lextendrtest.a
-PKG_LIBS = -L$(LIBDIR) -lextendrtests -lws2_32 -ladvapi32 -luserenv -lbcrypt -lntdll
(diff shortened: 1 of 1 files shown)


CGMossa:
This is not done. I'd like some help with it, because it needs to assign different libs for different platforms.
I've asked for help from the experts here https://github.com/r-rust/hellorust/issues/21

CGMossa:
The immediately relevant lines are:
https://github.com/rust-lang/rust/blob/bf8fff783ff533c055d276378ada30563312def1/tests/run-make/tools.mk#L137
https://github.com/rust-lang/rust/blob/bf8fff783ff533c055d276378ada30563312def1/tests/run-make/tools.mk#L139
https://github.com/rust-lang/rust/blob/bf8fff783ff533c055d276378ada30563312def1/tests/run-make/tools.mk#L160
https://github.com/rust-lang/rust/blob/bf8fff783ff533c055d276378ada30563312def1/tests/run-make/tools.mk#L174
https://github.com/rust-lang/rust/blob/bf8fff783ff533c055d276378ada30563312def1/tests/run-make/tools.mk#L198
https://github.com/rust-lang/rust/blob/bf8fff783ff533c055d276378ada30563312def1/tests/run-make/tools.mk#L205

CGMossa:
With the introduction of `--print native-static-libs` this is no longer important to do.

PR #778, opened 2024-06-06
dependabot: added `tests/extendrtests` to the directory to keep track of

CGMossa:


CGMossa:
I've tried to enable this on https://github.com/cgmossa/extendr to see if it made an effect or PRs.. and I honestly don't think this `directories` thing is supported. So I'll leave this open, but it needs.. a miracle?

PR #779, opened 2024-06-06
Support `&mut Robj`

CGMossa:
Using `&mut Robj` is _new_ and I think stuff like the following wasn't added.

It needs to be implemented for all "Robj"-types, so that one may return `&mut Robj` from `extendr`-fn and it working out. **You can already do this with `&Robj`.**

## Motivation

Recent change to the API now uses `&mut Robj` as a _return-type_. It would improve user experience if users could do

```rust
#[extendr]
fn generate_model_object() -> Robj {
    Robj::from(1).set_class(["ModelObject"]).unwrap().into() 
    // or `.into_robj()`, but here, the type-inference is obvious to rustc
}
```
instead of what they have to do at the moment, which is
```rust
#[extendr]
fn generate_model_object_old() -> Robj {
    let mut robj = Robj::from(1);
    robj.set_class(["ModelObject"]).unwrap();
    robj
}
```


kbvernon (approved):
At some point, we need to update the docs to elaborate on what this is actually doing, but I think this is good to go!

PR #781, opened 2024-06-07
[error handling]: Ensure all arguments are dropped 

CGMossa:
Another entry in fixing unwinding: In the C-wrappers, all `Robj` derived from `SEXP`s should be moved to the closure, that evaluates the user-provided function. This way, when the closure is finished, either by panic or otherwise, the arguments all dropped.

This actually solves 90% of the problem we have with unwinding.

There are others that does it similarly: See https://github.com/jrobsonchase/easy_ffi/blob/master/src/lib.rs#L128 from the `easy_ffi` crate.

### DIFF
--- a/extendr-macros/src/wrappers.rs
+++ b/extendr-macros/src/wrappers.rs
@@ -215,8 +215,8 @@ pub(crate) fn make_function_wrappers(
                 std::result::Result<Robj, extendr_api::Error>,
                 Box<dyn std::any::Any + Send>
             > = unsafe {
-                #( #convert_args )*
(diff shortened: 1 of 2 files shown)


JosiahParry:
This is such a simple and clean change! This makes quite a bit of sensed. By moving the robj's into the unwind closure, they will be appropriately handled. In this case seeing the macro snapshots has been quite helpful!—in that we can see exactly how the macro generation changed to address this change. In the macro snapshots we see that there is a `move ||` ensuring that everything is appropriately unwound. 

Beautiful!

JosiahParry (approved):

PR #782, opened 2024-06-09
[urgent] fix: relying on specific behavior for `List::iter`

CGMossa:
Discovered by @kbvernon: We removed this behaviour in #761, but it is used in `List::iter`. Basically, a `StrIter` can be initiated without any strings.

This points to `StrIter` being overloaded already, but we have to fix this bug sooner rather than later.

Here's the fix.

### DIFF
--- a/extendr-api/src/iter.rs
+++ b/extendr-api/src/iter.rs
@@ -78,7 +78,7 @@ impl Iterator for StrIter {
             if i >= self.len {
                 None
             } else if TYPEOF(vector) == SEXPTYPE::NILSXP {
-                None
+                Some(<&str>::na())
             } else if TYPEOF(vector) == SEXPTYPE::STRSXP {
                 str_from_strsxp(vector, i)
             } else if TYPEOF(vector) == SEXPTYPE::CHARSXP {
@@ -222,13 +222,20 @@ mod tests {
         with_r(|| {
             let single_charsxp = blank_string();
             let s1: Vec<_> = single_charsxp.as_str_iter().unwrap().collect();
-            // dbg!(&s1);
             let single_charsxp = blank_scalar_string();
             let s2: Vec<_> = single_charsxp.as_str_iter().unwrap().collect();
-            // dbg!(&s2);
             assert_eq!(s1, s2);
             assert_eq!(s1.len(), 1);
             assert_eq!(s2.len(), 1);
+        });
+    }
+
+    #[test]
+    fn test_new_constructor() {
+        with_r(|| {
+            let str_iter = StrIter::new(10);
+
+            assert_eq!(str_iter.collect::<Vec<_>>().len(), 10);
         });
     }
 }
--- a/extendr-api/src/iter.rs
+++ b/extendr-api/src/iter.rs
@@ -37,9 +37,10 @@ impl Default for StrIter {
 impl StrIter {
     /// Make an empty str iterator.
     pub fn new(len: usize) -> Self {
+        let vector = if len == 0 { nil_value() } else { na_string() };
         unsafe {
             Self {
-                vector: ().into(),
+                vector,
                 i: 0,
                 len,
                 levels: R_NilValue,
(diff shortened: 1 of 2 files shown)


JosiahParry:
Whats the scenario where this bug occurs. What I am seeing here is that if we call next on `StrIter` where the thing implementing is a `NULL` then we return `None`. Rather, we want to return a missing value. 

What I'm not tracking is how StrIter is available for a NULL Robj

CGMossa:
The new method may construct a vector of a certain size with NA Strings. This is used when iterating over lists with values, but no names.
See the test for the behavior expected of striter.

CGMossa:
> Whats the scenario where this bug occurs. What I am seeing here is that if we call next on `StrIter` where the thing implementing is a `NULL` then we return `None`. Rather, we want to return a missing value.
> 
> What I'm not tracking is how StrIter is available for a NULL Robj

Reading your reply, I've come up with a different change.

* An iterator over a `NULL` should return `None`.
* An iterator started with `StrIter::new` should contain `NA`, and it should produce those based on provided `len`.
* An iterator of `len = 0` ought to contain `NULL` instead of an `NA`. 

The test and the code now reflects this. Hopefully this addresses your comments @JosiahParry

JosiahParry:
Ah, I think this makes sense. because when you iterate over a list the iterator is `(Rstr, Robj)` where the first tuple element is possibly the name of the element. In most cases this is actually NULL which is why when I iterate through a list its most often done via `(_, Robj)`

extendr-api/src/iter.rs
```diff
@@ -222,13 +225,25 @@ mod tests {
...
+
+    #[test]
+    fn test_new_constructor() {
+        with_r(|| {
+            let str_iter = StrIter::new(10);
+            assert_eq!(str_iter.collect::<Vec<_>>().len(), 10);
+            let str_iter = StrIter::new(0);
+            let str_iter_collect = str_iter.collect::<Vec<_>>();
+            assert_eq!(str_iter_collect.len(), 0);
+            assert!(str_iter_collect.is_empty());
```
JosiahParry, line 244: This one confuses me a little bit. Because the check:

> `let vector = if len == 0 { nil_value() } else { na_string() };`

Wouldn't this result in a length 1 NULL list when `len == 0`?

CGMossa: I don't think. I'm going to answer with a test (on this PR branch of course):

```rs
#[test]
fn iterating_unamed_list() {
    with_r(|| {
        let unamed_list = List::from_values([1, 2, 3, 42, 100]);
        assert_eq!(unamed_list.iter().len(), 5);
        let unamed_list = List::from_values::<[i32; 0]>([]);
        assert_eq!(unamed_list.iter().len(), 0);

        dbg!(unamed_list.iter().collect::<Vec<_>>());
    });
}
```

---- wrapper::list::tests::iterating_unamed_list stdout ----
[extendr-api/src/wrapper/list.rs:450:13] unamed_list.iter().collect::<Vec<_>>() = []
```

`StrIter` is only used with `.iter()`. What do you think? Resolved?

**Comment by CGMossa** [comment]
@JosiahParry I believe this is also ready for review.

When it rains it pours.

File: extendr-api/tests/into_iterator_tests.rs [hunk 2]
```diff
@@ -0,0 +1,14 @@
+use extendr_api::prelude::*;
+use extendr_engine::with_r;
+
+#[test]
+fn iterating_unamed_list() {
+    with_r(|| {
+        let unamed_list = List::from_values([1, 2, 3, 42, 100]);
+        assert_eq!(unamed_list.iter().len(), 5);
```
JosiahParry, line 8:
```suggestion
        assert_eq!(unamed_list.names(), None);
        assert_eq!(unamed_list.iter().len(), 5);
```

extendr-api/tests/into_iterator_tests.rs
```diff
@@ -0,0 +1,20 @@
...
+
+        let empty_strings = unamed_list.iter().map(|(si, _)| si).collect::<Strings>();
+        let na_strs = (0..5).into_iter().map(|_| Rstr::na()).collect::<Strings>();
+        assert_eq!(empty_strings, na_strs);
+
+        let unamed_list = List::from_values::<[i32; 0]>([]);
+        assert_eq!(unamed_list.iter().len(), 0);
+
+        assert!(unamed_list.iter().collect::<Vec<_>>().is_empty());
+    });
```
JosiahParry, line 19: Thank you for adding this test.

CGMossa:
The other R-devel CI failure is due to _warnings_ (which are promoted to errors in our CI) of non-API usage. This is being dealt with in #809. I ask that those failures be ignored for now, as short of disabling errors on warnings, that is the only thing I can do to make it green.

JosiahParry (approved):
Thank you so much for the clear writing! I can confirm that r-devel errors are due to non-API changes. The other issues are as you described. Though I wonder how actionable zulip is. Merging!

Issue #788, opened 2024-06-10
Support for u32 slice removed - workaround


jonlachmann:
In my package I have the following code which takes a dgC-matrix from the Matrix R package and creates a sparse matrix in faer which does not copy the data, but instead creates a reference to it:

```
fn load_dgc_matrix<'a>(mat: S4, x: &'a Robj) -> faer::sparse::SparseColMatRef<'a, u32, f64> {
    use extendr_api::AsTypedSlice;
    let i = mat.get_slot("i").unwrap();
    let p = mat.get_slot("p").unwrap();
    let dim = mat.get_slot("Dim").unwrap();

    let i = i.as_typed_slice().unwrap();
    let p = p.as_typed_slice().unwrap();
    let x = x.as_real_slice().unwrap();
    let dim = dim.as_integer_slice().unwrap();
    let ncols = dim[0] as _;
    let nrows = dim[1] as _;

    // Map the sparse matrix
    let symbolic_1 = faer::sparse::SymbolicSparseColMatRef::new_checked(nrows, ncols, p, None, i);
    let a_mat = faer::sparse::SparseColMatRef::<u32, f64>::new({
            symbolic_1
        },
        x
    );

    return a_mat;
}
```

This does not work in the latest version of extendr as support for u32 slices was removed. The solution (provided by @CGMossa) is to modify the code as follows, it may not be the best solution in terms of type safety etc. but it works. If it would be possible to provide a full function for handling sparse matrices from Matrix, that would be awesome, as that is probably a not too uncommon use case.

```
    let i: &[i32] = i.as_typed_slice().unwrap();
    let i: &[u32] = unsafe { std::mem::transmute(i) };
    let p: &[i32] = p.as_typed_slice().unwrap();
    let p: &[u32] = unsafe { std::mem::transmute(p) };
```

CGMossa:
This was caused by changes made in #767. 

A potential solution for this would be to provide _checked_ conversion, see #789.

Ilia-Kosenkov:
Checked conversion is computationally heavy and introduces iterations over input vector, unlike any other type conversion we have. I'd suggest to just reinterprer `&[i32]` as `&[u32]` with optional runtime check. That way the validation tax will clearly be visible in your code.

FYI @CGMossa bedore you merge anything

CGMossa:
[...]

PR #789, opened 2024-06-10
Add: Checked conversion to `&[u32]` and `Vec<u32>` from integer vector

CGMossa:
#768 was rejected, because adding run-time checks to `AsTypedSlice` would break the promise that providing a slice would entail. 

Due to #788, and since `faer` only support `u32`/`u64`, it seems reasonable to provide safe conversions to atleast `u32`.

I believe `TryFrom` does not signal zero-checked conversion -- actually the opposite.

Therefore, I propose this addition: This conversion has a slight run-time cost for checking non-negative numbers, but does provide access to the underlying values by-reference without copying.

### DIFF
--- a/extendr-api/src/error.rs
+++ b/extendr-api/src/error.rs
@@ -88,6 +88,9 @@ pub enum Error {
 
     #[cfg(feature = "either")]
     EitherError(Box<Error>, Box<Error>),
+
+    // used in &[u32] conversion
+    ExpectedNonNegativeValue,
 }
 
 impl std::fmt::Display for Error {
@@ -191,6 +194,9 @@ impl std::fmt::Display for Error {
                     left_err, right_err
                 )
             }
+            Error::ExpectedNonNegativeValue => write!(
+                f, "expected vector of non-negative values"
+            )
         }
     }
 }
--- a/extendr-api/src/error.rs
+++ b/extendr-api/src/error.rs
@@ -194,9 +194,7 @@ impl std::fmt::Display for Error {
                     left_err, right_err
                 )
             }
-            Error::ExpectedNonNegativeValue => write!(
-                f, "expected vector of non-negative values"
-            )
+            Error::ExpectedNonNegativeValue => write!(f, "expected vector of non-negative values"),
         }
     }
 }
--- a/extendr-api/src/robj/try_from_robj.rs
+++ b/extendr-api/src/robj/try_from_robj.rs
@@ -353,6 +353,34 @@ impl TryFrom<&Robj> for &[f64] {
     }
 }
 
+impl TryFrom<&Robj> for &[u32] {
+    type Error = Error;
+
+    /// Returns a slice by reference, if the underlying `&[i32]` values are
+    /// all non-negative.
+    fn try_from(value: &Robj) -> Result<Self> {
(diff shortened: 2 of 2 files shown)


Ilia-Kosenkov:
I am a bit puzzled why we try to return this feature if we just agreed to remove it and agreed on the justification. R does support only signed 32 bit integer type out of the box, so we can only safely expose `&[i32]`. Since it is Rust, it takes like one line to reinterpret i32 as u32 (in an unsafe manner):
```Rust
fn main() {
    let x = vec![1i32, 2i32, 3i32, -1i32];
    let y: &[u32] = unsafe { std::mem::transmute(&x[..]) };
    
    dbg! { &x , &y};
}

[src/main.rs:5:5] &x = [
    1,
    2,
    3,
    -1,
]
[src/main.rs:5:5] &y = [
    1,
    2,
    3,
    4294967295,
]
```
And if you want runtime check for underflows, it is a separate validation step and IMO not the responsibility of the framework.

CGMossa:
> I am a bit puzzled why we try to return this feature if we just agreed to remove it and agreed on the justification. R does support only signed 32 bit integer type out of the box, so we can only safely expose `&[i32]`. 

We agreed that `AsTypedSlice` was an inappropriate place for a checked-conversion. But `TryFrom` is a place where very many checks and allocations occur. This shouldn't be puzzling at all.

> And if you want runtime check for underflows, it is a separate validation step and IMO not the responsibility of the framework.

That's an opinion. I can be swayed either one way or another, I just think we could provide this convenience without
losing too much on it anyways. @JosiahParry do you have an opinion on this, we can add to the discussion?

Ilia-Kosenkov:
My point is if I do not really understand how it works under the hood, `&[i32]` is basically a no-op for me, but `&[u32]` spins CPU cycles and I am not a fan of such inconsistency. With `Vec<u32>` I actually have no issue since it is a bit different, but I am not entirely convinced about runtime-element-checked slices (which should be as lightweight & transparent as possible).

This then warrant a question, should we work with `&[f32]` in the same manner? Or what about interpreting `&[f64]` as `&[u64]` to support 64 bit integer crate?

CGMossa:
> My point is if I do not really understand how it works under the hood, `&[i32]` is basically a no-op for me, but `&[u32]` spins CPU cycles and I am not a fan of such inconsistency.

I don't think there is a rule that stipulates all slices must be gotten in no-op way. I think this heavily depends on from whence the slice came from. That's why I agreed that `AsTypedSlice` wasn't the place for it. A lot of slices are constructed after going through all elements.. see `CStr`, or how we passed on `&str` until #761 landed. 

> This then warrant a question, should we work with `&[f32]` in the same manner? Or what about interpreting `&[f64]` as `&[u64]` to support 64 bit integer crate?

That's very good questions. Looking at `bytemuck` is where my opinion on this lies right now. Reinterpreting bytes also changes the _size_ of the slice. I think we can't do any conversion (even with checks), that ultimately changes the size of the slice, as that messes up with R's `length` assumption.

For `floats`, I think it is a question of precision, and a question of layout. I don't think we can work with those, unfortunately. 

I think we are getting a little too carried away with this. It is just about writing and providing ergonomic, convenient, and clear abstractions. I don't know what is _too much_ before some sort of proposal PR is written. That's why I've made so many on this specific subject (3 PRs in total).

PR #790, opened 2024-06-10
Added mutable slices from `AsTypedSlice` to `TryFrom`

CGMossa:
* Currently, we have direct mutable reference access facilitated by `AsTypedSlice`, but it isn't neatly available in `#[extendr]`-functions. This PR provides that.
* Note that there are _no mutable access to strings_, as that's not sound at all with R's string interner.

This enables

```rs
#[extendr]
fn middle_zero(integers: &mut [Rint]) {
  let middle = integers.len()/2;
  integers[middle] = 42_i32;
}
```

Instead of today's solution to the above

```rs
#[extendr]
fn middle_zero(mut integers: Robj) {
  let integers: &mut [Rint] = integers.as_typed_slice_mut().unwrap();
  let middle = integers.len()/2;
  integers[middle] = 42_i32;
}
```

Thus, this is a win for ergonomics, to facilities we already allow.


JosiahParry:
I've added integration tests to this PR

JosiahParry:
I'm unsure how to update the snapshots to get CI to pass. Running `cargo extendr devtools-test -a` does not create new snapshots.

JosiahParry:
Otherwise this PR is good!

CGMossa:
> Otherwise this PR is good!

Snapshot tests where updated and this is ready for another review.

JosiahParry (approved):
Nice! I updated the changelog. Note that they were all passing and my skip ci incantation was incorrect. Thank youuu!!!

PR #791, opened 2024-06-10
Update Cargo.toml

CGMossa:
Preparing for the new version

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ rust-version = "1.70"
 
 [workspace.dependencies]
 # When updating extendr's version, this version also needs to be updated
-extendr-macros = { path = "./extendr-macros", version = "0.6.0" }
+extendr-macros = { path = "./extendr-macros", version = "0.7.0" }
 
 # When uncommenting this, do not forget to uncomment the same line in
 # ./tests/extendrtests/src/rust/Cargo.toml, and "Run R integration tests using
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ resolver = "2"
 members = ["extendr-api", "extendr-engine", "extendr-macros", "xtask"]
 
 [workspace.package]
-version = "0.6.0"
+version = "0.7.0"
 authors = [
     "andy-thomason <andy@andythomason.com>",
     "Thomas Down",


Ilia-Kosenkov (approved):
Can we merge it closer to the release please? Still some unresolved PRs

CGMossa:
> Can we merge it closer to the release please? Still some unresolved PRs

No, I need this, because I'd like to see how it behaves with vendoring. Currently, vendoring is broken. And I noticed something _different_ happening with cargo when the version changes.

JosiahParry:
Can we settle for a patch version update in the meantime?

CGMossa:
> Can we settle for a patch version update in the meantime?

I don't know what that looks like. I think we should just update to it to `0.7.0` and hurry up with finalising the release...

JosiahParry:
A `0.6.1` or `0.6.0.9000` would indicate a new numeric version that can be used for testing in the mean time.

CGMossa:
> A `0.6.1` or `0.6.0.9000` would indicate a new numeric version that can be used for testing in the mean time.

I'm sorry for being pedantic about this, but this really not semver-complaint. And it does indicate that the version can be used without any changes whatsoever. It might be the custom in R packages, but that's really not how I see things done in the Rust world. 

I don't want to merge a PR without someone agreeing with me. But we really just ought to increase the version and move on.

JosiahParry:
I agree with Ilia that if we're going to increase the version, the features that you want to have merged should be there before the bump. I'm fine with adding a 0.7.0 so long as it is accompanied by a crates.io release. There's been enough changes and fixes that I think it's worth it. But if the objective is to bump to 0.7.0 to test vendoring then continue to add into the version, I think we wait.

JosiahParry:
Booyah

CGMossa:
🤷 At this point I'm fully ready to give up here.

1. I asked people to test extendr on `master`-branch, because I'd like just some sanity check that we haven't ruined people's lives before publishing a release.
2. One user reported that they can't build at all, because they use the vendored script from `rextendr` and they apparently don't work with the way we have set up things.
3. I researched vendored packages and vendored git packages and basically it should work. However, the setup doesn't copy the `git`-version of extendr.

Then comes this debate: I cannot change the version, before I send up a crates.io release. If we release, we ruin it for everybody at once. And I am fighting cargo's vendor feature, so I have to do what cargo likes. But I can't do that, because you won't let me. 


Closing this, and I will not attempt to fix the issue with `cargo vendor` .

PR #793, opened 2024-06-10
Remove `libR-sys` from extendrtests permanently

### LINKED ISSUE #787: Remove `libR-sys` from `tests/extendrtests`, as it is our r integration test, and we should mimic actual users, which may not include `libR-sys`.

CGMossa:
Fixes #787 

As agreed, we now prefer `R_HOME` in `tests/extendrtests` or if R is available from `PATH`.

It is reasonable enough, to assume that the contributors to extendR, will have configured their terminal to have R available there.

### DIFF
--- a/tests/extendrtests/src/rust/build.rs
+++ b/tests/extendrtests/src/rust/build.rs
@@ -1,8 +1,4 @@
 fn main() {
-    // TODO: I couldn't find any nice way to add the condition based on the R version
-    // except for using libR-sys just for "DEP_R_*" envvars.
-    // let major = std::env::var("DEP_R_R_VERSION_MAJOR").unwrap();
-    // let minor = std::env::var("DEP_R_R_VERSION_MINOR").unwrap();
     let mut major = 0;
     let mut minor = 0;
     let mut _patch = 0;
@@ -33,6 +29,9 @@ fn main() {
         minor = raw_r_version[1];
         _patch = raw_r_version[2];
     }
+    let major = major;
+    let minor = minor;
+    let _patch = _patch;
     // dbg!(major, minor);
 
     if major >= 4 && minor >= 3 {
--- a/tests/extendrtests/src/rust/build.rs
+++ b/tests/extendrtests/src/rust/build.rs
@@ -9,7 +9,7 @@ fn main() {
 
     // prefer `R_HOME` if set, otherwise use the `PATH` available R
     let mut path_to_r = if let Some(r_home_env) = r_home_env {
-        std::process::Command::new(r_home_env.join("R"))
+        std::process::Command::new(r_home_env.join("bin/R"))
     } else {
         std::process::Command::new("R")
     };
--- a/tests/extendrtests/src/rust/build.rs
+++ b/tests/extendrtests/src/rust/build.rs
@@ -1,38 +1,53 @@
+use std::path::PathBuf;
+
 fn main() {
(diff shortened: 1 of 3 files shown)


tests/extendrtests/src/rust/build.rs
```diff
@@ -1,12 +1,56 @@
...
-    let minor = env::var("DEP_R_R_VERSION_MINOR").unwrap();
+    println!("cargo:rerun-if-env-changed=R_HOME");
+    println!("cargo::rerun-if-changed=build.rs");
 
-    if &*major >= "4" && &*minor >= "3" {
+    let r_home_env = std::env::var_os("R_HOME");
+    let r_home_env = r_home_env.map(PathBuf::from);
+
+    // prefer `R_HOME` if set, otherwise use the `PATH` available R
+    let mut r_version = if let Some(r_home_env) = r_home_env {
```
Ilia-Kosenkov, line 11: Why do you call this `r_version`? While it seems like it is a path to R executable

tests/extendrtests/src/rust/build.rs
```diff
@@ -1,12 +1,56 @@
...
+        .split(':')
+        .nth(1)
+        .unwrap()
+        .trim()
+        // ignore commit
+        // only capture the first the major.minor.patch part of the version
+        .split_ascii_whitespace()
+        .next()
+        .unwrap();
+    dbg!(raw_r_version);
```
Ilia-Kosenkov, line 39: Do we need debug statement in 'production' code?

CGMossa: I recently found out, that calls to `dbg!` gets piped to a file `tests/extendrtests/src/rust/target/debug/build/extendrtests-44614d0cd19b0ae0/stderr` (hashing is part of the incremental compilation I think), which means you'd have a file that contains (this doesn't get printed to user, terminal or elsewhere)
```
[build.rs:39:5] raw_r_version = "4.5.0"
```

So that's helpful.

I've removed a ton of these `dbg!` calls and one slipped out, but I think it is nice with some reporting, somewhere neat (doesn't pollute terminal).

tests/extendrtests/src/rust/build.rs
CGMossa: Renamed.

tests/extendrtests/src/rust/build.rs
```diff
@@ -36,11 +36,14 @@ fn main() {
...
         .next()
         .unwrap();
-    dbg!(raw_r_version);
     let raw_r_version: Vec<u32> = raw_r_version
         .split('.')
         .map(|x| x.parse().unwrap())
         .collect();
+    assert!(
+        raw_r_version.len() >= 3,
+        "rust version was not detected properly"
```
Ilia-Kosenkov, line 45: I believe it is 
```suggestion
        "R version was not detected properly"
```
no?

Ilia-Kosenkov (approved):

CGMossa:
While this is approved, it doesn't pass CI.. Somehow, R-cmd-check removes access to R in the build-script:
>  /usr/bin/bash --noprofile --norc -e -o pipefail {0}
🤷‍♂️

CGMossa:
I actually don't know the issue. If any CI hero is around, please take a look.

CGMossa:
Figured it out.

PR #800, opened 2024-06-16
Refactor matrix dimnames 

JosiahParry:
This PR adds integration tests to PR 796. 
Additionally, it changes the definition of the methods introduces as well as adds another.

The methods for get_colnames() and get_rownames() returned NULL on the matrix. I believe the Rf_* functions that were being used were intended for data.frames or something else


JosiahParry:
Didn't realize cargo extendr would change my cargo.toml paths!

JosiahParry:
@CGMossa please review

extendr-api/src/wrapper/matrix.rs
```diff
@@ -118,13 +118,35 @@ where
...
-        let rownames = Robj::from_sexp(unsafe { Rf_GetRowNames(self.get()) });
-        rownames
+        let dims = self.dimnames();
+
+        if dims.is_none() {
+            return ().into_robj();
+        }
+
+        let dims = dims.unwrap().collect::<List>();
+        dims[1].clone()
```
CGMossa, line 149: I don't know why you got `NULL`, but you shouldn't, see this:

```c
/* "GetRowNames" and "GetColNames" are utility routines which
 * locate and return the row names and column names from the
 * dimnames attribute of a matrix.  They are useful because
 * old versions of R used pair-based lists for dimnames
 * whereas recent versions use vector based lists.

 * These are now very old, plus
 * ``When the "dimnames" attribute is
 *   grabbed off an array it is always adjusted to be a vector.''

 They are used in bind.c and subset.c, and advertised in Rinternals.h
*/
SEXP GetRowNames(SEXP dimnames)
{
    if (TYPEOF(dimnames) == VECSXP)
    return VECTOR_ELT(dimnames, 0);
    else
    return R_NilValue;
}

SEXP GetColNames(SEXP dimnames)
{
    if (TYPEOF(dimnames) == VECSXP)
    return VECTOR_ELT(dimnames, 1);
    else
    return R_NilValue;
}
```

JosiahParry: Can you try your existing branch with the integration tests? The integration tests show that they weren't working for me.

CGMossa: I see the issue. These `GetColNames` takes a `dimname`.. Duh.

CGMossa:
Okay, I unfortunatley had to copy them over. But they helped fix the last bit. Sorry, but closing this :(

PR #796, opened 2024-06-12
Add `names<-`,  `dim<-` and `dimnames<-` to `RArray` and friends

CGMossa:

PR #801, opened 2024-06-16
Less ambigious return type for colnames/rownames

CGMossa:
Follow-up to #796

Instead of returning `Robj`, we should aim to be more descriptive of the returned result. For `colnames`/`rownames` it is a bit obvious that it should be `NULL` or a character vector.

_This uncovered something_ that our vector types `Integers`, `Doubles` etc. are not implemented for `From<Robj> for Option<r-vector-type>`. That could be a follow-up to this.

- [ ] Partial dependence on #802.

### DIFF
--- a/extendr-api/src/wrapper/matrix.rs
+++ b/extendr-api/src/wrapper/matrix.rs
@@ -123,7 +123,7 @@ impl<T> RMatrix<T> {
                     let colnames = Robj::from_sexp(maybe_colnames);
                     Some(std::mem::transmute(colnames))
                 }
-                _ => unreachable!(),
+                _ => unreachable!("This should not have occurred. Please report an error at https://github.com/extendr/extendr/issues"),
             }
         }
     }
@@ -136,7 +136,7 @@ impl<T> RMatrix<T> {
                     let rownames = Robj::from_sexp(maybe_rownames);
                     Some(std::mem::transmute(rownames))
                 }
-                _ => unreachable!(),
+                _ => unreachable!("This should not have occurred. Please report an error at https://github.com/extendr/extendr/issues"),
             }
         }
     }
--- a/extendr-api/src/wrapper/matrix.rs
+++ b/extendr-api/src/wrapper/matrix.rs
@@ -114,13 +114,31 @@ where
 }
 
 impl<T> RMatrix<T> {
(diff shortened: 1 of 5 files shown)


kbvernon (approved):
The code and tests look straightforward enough. Additional arguments for this change are
1. It uses std lib.
2. It's still technically returning an `Robj`.

So, I support it. Just need confirmation from someone with write access.

extendr-api/src/wrapper/matrix.rs
```diff
@@ -114,13 +114,31 @@ where
...
+    pub fn get_colnames(&self) -> Option<Strings> {
+        unsafe {
+            let maybe_colnames = Rf_GetColNames(Rf_GetArrayDimnames(self.get()));
+            match TYPEOF(maybe_colnames) {
+                SEXPTYPE::NILSXP => None,
+                SEXPTYPE::STRSXP => {
+                    let colnames = Robj::from_sexp(maybe_colnames);
+                    Some(std::mem::transmute(colnames))
+                }
+                _ => unreachable!(),
```
JosiahParry, line 126:
```suggestion
                _ => unreachable!("This should not have occurred. Please report an error at https://github.com/extendr/extendr/issues"),
```

extendr-api/src/wrapper/matrix.rs
```diff
@@ -114,13 +114,31 @@ where
...
+    pub fn get_rownames(&self) -> Option<Strings> {
+        unsafe {
+            let maybe_rownames = Rf_GetRowNames(Rf_GetArrayDimnames(self.get()));
+            match TYPEOF(maybe_rownames) {
+                SEXPTYPE::NILSXP => None,
+                SEXPTYPE::STRSXP => {
+                    let rownames = Robj::from_sexp(maybe_rownames);
+                    Some(std::mem::transmute(rownames))
+                }
+                _ => unreachable!(),
```
JosiahParry, line 139:
```suggestion
                _ => unreachable!("This should not have occurred. Please report an error at https://github.com/extendr/extendr/issues"),
```

CGMossa:
@JosiahParry Thanks! @kbvernon & I have adressed all of your comments.

JosiahParry (approved):
Thanks for making the change!

PR #802, opened 2024-06-16
Support for `Option<Strings>`, `Option<Doubles>`, and friends 

CGMossa:
One is now able to

```rs
#[extendr]
fn return_me() -> Option<Strings> {
// ...
}
```

And that seems like a simple oversight. 

I've added this for `Strings`, `Complexes`, `Doubles`, `Integers`, `S4`, `ExternalPtr<T>`, `RArray<T,D>` (and thus also for `RMatrix`). It is *not* possible to add for `#[extendr]`-impl blocks as we don't own those types involved (or something like that). 

This oversight comes because we have concentrated about parsing types _from_ R but not necessarily _to_ R, where there we must provide `Into<Robj>` (by way of `From<T>` for a given `T`).


JosiahParry:
It's worth noting here that we are providing a very specific opinion for how returning an `Option<T>` should behave. In these cases we are saying that in the case of `None` we are returning `NULL` and the case of `Some<T>` we are returning `T`

JosiahParry:
Resolved merge conflicts. CI is starting.

JosiahParry (approved):
By adding From<T> for Robj implementations we can now support more return types for #[extendr] annotated functions. 

In the case of None we return `NULL`. 

Love a 0 LOC removed PR. 

Once we get the ok from CI we can merge it.

Issue #799, opened 2024-06-14
Lower the MSRV?


andyquinterom:
I got the following comment from CRAN today in the morning.

> Rust 1.69.0 is only 14 months old -- it is unreasonable to expect OSes
> to have a more recent version (and LTS versions could be a lot older,
> e.g. 1.41.0).

Is it worth an effort from my side to try to lower to MSRV? 

Thinks I would probably need to do:

1. Remove workspace specific DX features that could be replaced with a script.
2. Check for dependencies and their MSRV. I see for example `faer` has a `1.67` MSRV.
3. Extensibly test locally to see if it is possible to lower the MSRV.
4. Add proper CI for testing. 

If this effort is not worth it, perhaps we need to document in the system requirements and the minimum supported R version of `rextendr`.

Currently `rextendr` has the following R requirements.

```
Depends:
    R (>= 4.0)
```

JosiahParry:
I submitted a package to CRAN a few weeks ago with the latest version of extendr (from a commit) and it wasn't an issue. It appears that this change was introduced in https://github.com/extendr/extendr/pull/742. 

Do we know _why_ the MSRV was bumped other than CI was failing?

eitsupi:
The MSRV for extendr has been tested and is declared below.

https://github.com/extendr/extendr/blob/cb85a21a90c3fcb538a40fcea03058454edb3dac/Cargo.toml#L17-L20

I don't think it is worth the maintenance cost for the R package to keep the MSRV at 1.69 for submission to CRAN.
I have heard that this year Cargo will be coming out with edition 2024 which will have MSRV-aware dependency resolution, which will be of great benefit to many users like us, but it may be 10 years before CRAN replaces Fedora 36, so we will not see the benefit for years to come.
I recommend that you give up CRAN in favor of R-universe.

That said, I am aware of several alternatives for submitting to CRAN.

1. The gifski package (not to mention the oldest R package using Rust on CRAN) has a mechanism to replace the Rust source code when it detects old rustc. <https://github.com/r-rust/gifski/blob/a5d830a49f591fbc444be8399e68094f457daa5f/configure#L11-L27>
2. savvy's MSRV is currently below 1.69 (although I believe this is only true for now and does not guarantee CRAN support for the next 10 years). <https://github.com/PRQL/prqlc-r/blob/e22c19b2dfc02e11526c68cb3a7eb9d727bc4e6c/src/rust/Cargo.toml#L2-L5>

CGMossa:
[...]

PR #803, opened 2024-06-17
[PoC] Lower MSRV to 1.69

yutannihilation:
An attempt for https://github.com/extendr/extendr/issues/799

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [workspace]
 resolver = "2"
 
-members = ["extendr-api", "extendr-engine", "extendr-macros"]
+members = ["extendr-api", "extendr-engine", "extendr-macros", "xtask"]
 
 [workspace.package]
 version = "0.6.0"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [workspace]
 resolver = "2"
 
-members = ["extendr-api", "extendr-engine", "extendr-macros"]
+members = ["extendr-api", "extendr-engine", "extendr-macros", "xtask"]
```
CGMossa, line 4: I think we don't want `xtask` to be part of the workspace. Am I wrong?

yutannihilation: I don't know, but the CI fails with this error and I'm just too tired to fix this.

```
cargo extendr r-cmd-check --error-on warning --check-dir extendrtests_check
error: package(s) `xtask` not found in workspace `D:\a\extendr\extendr`
```

yutannihilation:
Okay, I'm closing this as it's confirmed Rust 1.69 is actually supported if only you lower the `workspace.rust-version` (and tweak xtask).

PR #805, opened 2024-06-17
Add flat indexing to `RMatrix` and friends

CGMossa:
It should be possible to index a matrix/array with one index, that treats a matrix $A$ like a $vec(A)$ in terms of column-order. This behavior is available in R:

```r
> A <- pracma::magic(10)
> A
      [,1] [,2] [,3] [,4] [,5] [,6] [,7] [,8] [,9] [,10]
 [1,]   92   99    1    8   15   67   74   51   58    40
 [2,]   98   80    7   14   16   73   55   57   64    41
 [3,]    4   81   88   20   22   54   56   63   70    47
 [4,]   85   87   19   21    3   60   62   69   71    28
 [5,]   86   93   25    2    9   61   68   75   52    34
 [6,]   17   24   76   83   90   42   49   26   33    65
 [7,]   23    5   82   89   91   48   30   32   39    66
 [8,]   79    6   13   95   97   29   31   38   45    72
 [9,]   10   12   94   96   78   35   37   44   46    53
[10,]   11   18  100   77   84   36   43   50   27    59
> A[22]
[1] 7
> A[2, 3]
[1] 7
```

And it should also be available in Rust. This PR implements this.


JosiahParry:
This looks great! We're just missing a test. 

I think this case a single `with_r(||...)` would be sufficient to ensure that the indexing works as anticipated. I think it should `assert_eq!()` that, for example, given a 3x3 matrix that `m[0,1]` is the same as `m[3]`

CGMossa:
Thanks! I've added the test.

Upon implementing the test, something came up: It is not possible to define multiple definitions of `Index`/`IndexMut` for the same type. I then made it now possible to flat index when you interpret the matrix as a `RArray<T, ()>`, i.e. you declare the index type as nothing `()`. 

I've added the appropriate conversions and the test shows how the mechanism would then work.

extendr-api/tests/rarray_tests.rs
```diff
@@ -0,0 +1,19 @@
+//! This test module contains tests for [RArray], [RMatrix], etc.
+//!
+use extendr_api::prelude::*;
+use extendr_engine::with_r;
+
+#[test]
+fn flat_index_test() {
+    with_r(|| {
+        let example = R!("matrix(as.numeric(1:9), byrow=TRUE, ncol = 3, nrow = 3)").unwrap();
+        let test_array = RArray::<f64, ()>::try_from(&example).unwrap();
```
JosiahParry, line 10: In this case, wouldn't it make more sense to make an `RArray<f64,[usize;1]>` and have an index trait for that constraint? 

Alternatively, could we have a `flat_index()` method instead? Or is the `[]` notation super important? I guess from my perspective if you have to cast to `RArray<f64, ()>` what is the difference to just using `as_real_slice()`

CGMossa: It was a big mistake to add this dimension thing to `RArray`/`RMatrix`. It is inspired by `ndarray`. One thing that `ndarray` does in the case of run-time specified indices, is it has https://docs.rs/ndarray/latest/ndarray/fn.IxDyn.html, which is more closer to what we have with R's matrix stuff.

I think we really only know that `D = ()`, as it is a runtime thing, and can be changed, unless you solemnly swear, that you won't change it. But that's not a real rust API. Pinky promises aren't a thing here.

I think we should rectify the situation as much, and introduce the use `()` in the API more and more. For instance, it feels like `Dataframe<T>` can only be used when you have a row-type `T`. But that's not true. There is a lot of advantages from using `Dataframe<()>`, mainly the checks are simpler. 

I think we pre-emptively cluttered the design of these things.. If something should be special cased, it should be this `D` thing, and the column-wise flat index should be the standard.

CGMossa: Yes, I'm definitely thinking those things. 

I think you're right, this is a little too loose patch. A refactor would be more appropriate. 
Let me think about it and propose another PR.

CGMossa: Whatever is in there already is strange. For instance, I couldn't find an impl for accepting `array()` from R. So it is incomplete.

PR #809, opened 2024-06-25
[urgent] Update non-API for CRAN 

CGMossa:
CRAN compliance PR.

Background: CRAN has new requirements for what is non-API. These requirements just came up and became defacto law of the land. They are not tied to any R release. We just have to deal with it.

We pulled non-API list in 
https://github.com/extendr/libR-sys/pull/248

This PR syncs `libR-sys` with this commit.

- Hid previous available functionality behind `non-api` feature. It means, if you want it, you may compile `extendr-api` with `feature = ["non-api"]`. This requires you to build your own bindings by way of `libR-sys` as wel.
- Moved all doc-tests to `non_api_tests`, and in order to make that work, I had to add an entry for it in `extendr-api/Cargo.toml`. Note that the `optional`-module in integration tests doesn't work. And it needs to also have an entry to work.
- I also added `required-features`, but this doesn't run the test _unless_ the feature is provided to the `cargo test`-call. So these tests are present, but won't run. We have to have a broader conversation about testing `non-api`. For now, this is a patch, that allows us to rediscover missing features if users want them.


JosiahParry:
Overall, I am appreciative of these changes. 

Though I wonder if there is utility in keeping the removed doc tests and instead of having them wrap `test!` they can be standard markdown blocks for standard documentation purposes. 

While we should entirely revamp that page, I don't want to lose the examples that are there. 

Can we either: 

1. remove the `test!` macro or
2. Comment these out instead of removing entirely?

CGMossa:
Oh. I don't know. 

Didn't think of either of those options to be honest.
For (2) I'd say some people hate commented-out code in a code-base, and for (1) if they don't actually work, why should they still be present?

The removed doc-test is because these invocations won't work anymore. 

EIther way, I welcome changes to this PR. I'm not planning on doing more here.

extendr-api/src/functions.rs
```diff
@@ -133,14 +135,6 @@ pub fn new_env(parent: Environment, hash: bool, capacity: i32) -> Environment {
 }
 
 /// The base environment; formerly `R_NilValue`
-///
-/// ```
-/// use extendr_api::prelude::*;
-/// test! {
-///     global_env().set_local(sym!(x), "hello");
-///     assert_eq!(base_env().local(sym!(+)), Ok(r!(Primitive::from_string("+"))));
```
JosiahParry, line 141: These tests are removed because they are non-api?

JosiahParry: seee line 4

extendr-api/src/rmacros.rs
```diff
@@ -37,18 +37,6 @@ macro_rules! r {
...
-      test outside of R process, so this doc test is disabled for now. See #211
-      for the details.
-*/
-/// ```no_run
-/// use extendr_api::prelude::*;
-/// test! {
-///     current_env().set_local(sym!(myvar), 1.0);
-///     assert_eq!(var!(myvar), Ok(r!(1.0)));
-/// }
-/// ```
```
JosiahParry, line 51: This test is removed because you cannot modify the base env anymore

extendr-api/src/rmacros.rs
```diff
@@ -59,13 +47,6 @@ macro_rules! var {
 /// Get a global variable.
 ///
 /// Variables with embedded "." may not work.
-/// ```
-/// use extendr_api::prelude::*;
-/// test! {
-///      // The "iris" dataset is a dataframe.
-///      assert_eq!(global!(iris)?.is_frame(), true);
-/// }
```
JosiahParry, line 67: This test is removed because it uses global which is non-api

extendr-api/src/robj/mod.rs
```diff
@@ -802,7 +800,7 @@ make_typed_slice!(Rint, INTEGER, INTSXP);
 make_typed_slice!(f64, REAL, REALSXP);
 make_typed_slice!(Rfloat, REAL, REALSXP);
 make_typed_slice!(u8, RAW, RAWSXP);
-make_typed_slice!(Rstr, STRING_PTR, STRSXP);
```
JosiahParry, line 805: See line here. STRING_PTR is removed from the public API and it _looks like_ that STRING_PTR_R0 and other `_R0` might be user facing now? 

Is this your read as well?
https://github.com/r-devel/r-svn/blob/48e02881502dce9eeeefff1bc422bc0968913c04/src/include/Defn.h#L427

extendr-api/src/robj/rinternals.rs
```diff
@@ -162,24 +162,13 @@ pub trait Rinternals: Types + Conversions {
...
-    /// test! {
-    ///    let iris_dataframe = global_env()
-    ///        .find_var(sym!(iris)).unwrap().eval_promise().unwrap();
-    ///    assert_eq!(iris_dataframe.is_frame(), true);
-    ///    assert_eq!(iris_dataframe.len(), 5);
     ///
-    ///    // Note: this may crash on some versions of windows which don't support unwinding.
-    ///    //assert_eq!(global_env().find_var(sym!(imnotasymbol)), None);
-    /// }
-    /// ```
```
JosiahParry, line 176: removed because global_env() is non-api

extendr-api/src/wrapper/altrep.rs
```diff
@@ -155,7 +155,7 @@ pub trait AltrepImpl: Clone + std::fmt::Debug {
 fn manifest(x: SEXP) -> SEXP {
     single_threaded(|| unsafe {
         Rf_protect(x);
-        let len = XLENGTH_EX(x);
```
JosiahParry, line 158: This appears which i think is an a-ok location? 
https://github.com/r-devel/r-svn/blob/48e02881502dce9eeeefff1bc422bc0968913c04/src/include/Rinternals.h#L1057C34-L1057C34

extendr-api/src/wrapper/environment.rs
```diff
@@ -108,6 +109,7 @@ impl Environment {
         self
     }
 
+    #[cfg(feature = "non-api")]
     /// Iterate over an environment.
     pub fn iter(&self) -> EnvIter {
         unsafe {
```
JosiahParry, line 115: This is a breaking change. Needs to be in changelog

We `iter()` method for an environment is now non-api feature which means that if anyone was using this, it will now no longer be cran compliant and that they need to add a feature flag.

extendr-api/src/wrapper/environment.rs
```diff
@@ -127,6 +129,7 @@ impl Environment {
         }
     }
 
+    #[cfg(feature = "non-api")]
     /// Get the names in an environment.
     /// ```
     /// use extendr_api::prelude::*;
```
JosiahParry, line 135: Doc breaking change for `set_local()` method. Goes in changelog

extendr-api/src/wrapper/environment.rs
```diff
@@ -187,27 +184,6 @@ impl Environment {
...
-///     small_env.set_local(sym!(x), 1);
-///     let names_and_values = small_env.as_environment().unwrap().iter().collect::<Vec<_>>();
-///     assert_eq!(names_and_values, vec![("x", r!(1))]);
-///
-///     let large_env = Environment::new_with_capacity(global_env(), 1000);
-///     large_env.set_local(sym!(x), 1);
-///     let names_and_values = large_env.as_environment().unwrap().iter().collect::<Vec<_>>();
-///     assert_eq!(names_and_values, vec![("x", r!(1))]);
-/// }
-///
```
JosiahParry, line 209: removed because non-api

extendr-api/src/wrapper/function.rs
```diff
@@ -29,6 +29,7 @@ pub struct Function {
 }
 
 impl Function {
+    #[cfg(feature = "non-api")]
```
JosiahParry, line 32: Note that this is a breaking change. Goes in changelog.

extendr-api/src/wrapper/primitive.rs
```diff
@@ -2,21 +2,15 @@ use super::*;
...
 /// Wrapper for creating primitive objects.
 ///
-/// Make a primitive object, or NULL if not available.
-/// ```
-/// use extendr_api::prelude::*;
-/// test! {
-///     let builtin = r!(Primitive::from_string("+"));
-///     let special = r!(Primitive::from_string("if"));
-/// }
-/// ```
```
JosiahParry, line 12: removed because non-api

extendr-api/src/wrapper/primitive.rs
```diff
@@ -2,21 +2,15 @@ use super::*;
...
+/// Make a primitive object, or `NULL` if not available
 ///
 #[derive(PartialEq, Clone)]
 pub struct Primitive {
     pub(crate) robj: Robj,
 }
 
 impl Primitive {
+    #[cfg(feature = "non-api")]
     /// Make a Primitive object from a string.
```
JosiahParry, line 14: goes in changelog Primitive::from_string()

extendr-api/src/wrapper/promise.rs
```diff
@@ -29,6 +29,7 @@ impl Promise {
         })
     }
 
+    #[cfg(feature = "non-api")]
     /// Get the code to be executed from the promise.
     pub fn code(&self) -> Robj {
```
JosiahParry, line 34: Should just note that like everything promise related is now non-api. 

Could this be related to the CVE in R? lol

extendr-api/src/wrapper/promise.rs
```diff
@@ -82,9 +87,13 @@ impl Promise {
...
-        f.debug_struct("Promise")
-            .field("code", &self.code())
-            .field("environment", &self.environment())
-            .finish()
+        let mut result = f.debug_struct("Promise");
+
+        #[cfg(feature = "non-api")]
+        {
+            let result = result.field("code", &self.code());
+            let result = result.field("environment", &self.environment());
```
JosiahParry, line 95: doesnt this need to be `mutate = ` and not `let result ` as the inner result is scoped to that section? Or does the outer ` { } ` get ignored because of the cfg flag?

JosiahParry (changes requested):
I have one question regarding scope of the mutable assignments.

But we have a lot to add to the changelog. And should add receipts from r-devel exchanges as well.

yutannihilation:
> * Hid previous available functionality behind `non-api` feature.

While I agree this is a good idea for a temporary solution, I'd recommend you to re-implement the functionality using the provided APIs (or simply remove them). I think the R core team is cleaning up these "non-API"s, so it's possible there's no time until they actually remove them (I guess not all, but some), which means extendr won't compile with the `non-api` feature.

extendr-api/src/wrapper/promise.rs
CGMossa: `cfg`applies to the `{}`, so we are good in that case :)

extendr-api/src/wrapper/function.rs
JosiahParry: @CGMossa thoughts on this? I'm fine removing it for now. I don't think it is being used in the wild at the moment.

CHANGELOG.md
```diff
@@ -23,6 +23,12 @@
 
 #### Breaking changes
 
+- R-devel Non-API changes:
+  - R's C API is being formalized. While the changes are formalized, non-API functions are hidden behind a feature flag to prevent removal from CRAN.
+  - See [[Rd] clarifying and adjusting the C API for R](https://stat.ethz.ch/pipermail/r-devel/2024-June/083449.html)
+  - [nonAPI.txt](https://github.com/r-devel/r-svn/blob/f36c203d3a53a74d56a81d4f97a68d24993e0652/src/library/tools/R/sotools.R#L564) functions are hidden behind the `non-api` feature flag.
+  - Removed from default include (but may not be limited to):
+    - `global_var()`, `local_var()`, `base_env()`, various `Environment`, `Function`, `Primitive`, and `Promise` methods.
```
JosiahParry, line 31: @CGMossa please review this changelog.

CGMossa: Yes, this work. Please mention CRAN somewhere there, because it seems a little absurd that we are pre-emptively pushing R-devel changes to extendr. Of course, it is because CRAN is rejecting packages that don't comply immediately.

extendr-api/src/robj/mod.rs
CGMossa: Yes. They made functions that return const-ptr instead of mutable-ptr, so this makes sense. The `VECTOR_PTR` version has been deprecated for a while now.

extendr-api/src/wrapper/function.rs
CGMossa: I'm for removing these for now.

CHANGELOG.md
JosiahParry: Yes i've updated it a bit.

extendr-api/src/wrapper/altrep.rs
CGMossa: The link doesn't provide exact location. But yes?

CGMossa (commented):
The `CHANGELOG` works for me.

extendr-api/src/wrapper/environment.rs
CGMossa: @JosiahParry the changelog doesn't contain this.

extendr-api/src/wrapper/environment.rs
CGMossa: This is missing from changelog too

JosiahParry:
Ubuntu CI is failing due to multithreading of the cargo tests which is poisoning the thread safety mutex. This can be fixed by #674 

The windows tests are failing at test-metadata.

CGMossa:
If I rebase, then the meta data tests will succeed.

CGMossa:
> I have one question regarding scope of the mutable assignments.
> 
> But we have a lot to add to the changelog. And should add receipts from r-devel exchanges as well.

I've now rebased. I don't understand what to reply to this. Is it possible to do or not?

extendr-api/src/wrapper/environment.rs
JosiahParry: > ...various `Environment`, `Function`, `Primitive`, and `Promise` methods.

do you want the iter method specifically mentioned here?

CGMossa: No, not at all.

PR #810, opened 2024-06-25
[internal] Adjust path for `extendrtests` 

CGMossa:
The path in `tests/extendrtests/src/rust/Cargo.toml` is very obscure.

I don't know who it serves.

With `xtask`, we adjust this in a process, to make it work with `devtools` and `rcmdcheck`.

Now, I ask that we merge this, so that it works when one is developing extendr, in the repository as workspace/working-directory, and when in vscode,

```json
{
    "rust-analyzer.linkedProjects": [
        "tests/extendrtests/src/rust/Cargo.toml",
    ]
}
```

I ask you to merge this as soon as possible, because I'm tired of things being set in an unhelpful way.


dfalbel (approved):
LGTM, should use `cargo extendr r-cmd-check` instead of checking from RStudio.

PR #811, opened 2024-06-25
Fix unnecessary note when running CMD CHECK

dfalbel:
Changes the URL so we don't get:

```
Found the following (possibly) invalid URLs:
URL: https://opensource.org/license/mit/ (moved to https://opensource.org/license/mit)
    From: README.md
    Status: 200
    Message: OK
```


CGMossa (approved):
Yes
I see this all the time and I fixed it locally a bunch.. But to put a PR was just too much. Thanks!

PR #812, opened 2024-06-28
Lower MSRV for CRAN Compliance

CGMossa:
This PR lowers extendr's MSRV to 1.65 making it compatible with CRAN.

It is not no longer possible to build bindings by way of `cargo`-features.

The feature `non-api` is fully broken, and enabling it won't compile.


.github/workflows/test.yml
```diff
@@ -257,187 +255,3 @@ jobs:
...
-            ci-cargo +$toolchain test test_result_list --manifest-path extendr-api/Cargo.toml --features tests-minimal,result_list $(if($target -ne 'default') {"--target=$target"} ) $env:EXTRA_ARGS '--' --nocapture -ActionName "Test extendr-api \w tests-minimal for $target target with `result_list` feature"
-            ci-cargo +$toolchain test test_result_condition --manifest-path extendr-api/Cargo.toml --features tests-minimal,result_condition $(if($target -ne 'default') {"--target=$target"} ) $env:EXTRA_ARGS '--' --nocapture -ActionName "Test extendr-api \w tests-minimal for $target target with `result_condition` feature"
-                      
-            ci-cargo +$toolchain test --manifest-path extendr-macros/Cargo.toml  $(if($target -ne 'default') {"--target=$target"} ) $env:EXTRA_ARGS '--' --nocapture -ActionName "Test extendr-macros for $target target"
-          
-          }
-        env:
-          BUILD_TARGETS: ${{ join(matrix.config.rust-targets, ',') }}
-          RUST_TOOLCHAIN: ${{ matrix.config.rust-version }}
-          EXTRA_ARGS: ${{ join(matrix.config.extra-args, ',') }}
```
JosiahParry, line 443: These tests are removed because we have removed the `use-bindgen` feature from libR-sys. We will need to reimplement this using an xtask per https://github.com/extendr/libR-sys/issues/189.

This needs to be removed so that we can address CRAN issues within the next 8 days. (Insane).

CHANGELOG.md
```diff
@@ -40,6 +40,8 @@ now hidden behind the feature flag `non-api`. Also, `Promise::from_parts` is mar
 - You can no longer create an `Robj` from a reference `&T`, where `T` is an `extendr`-impl. [[#759]](https://github.com/extendr/extendr/pull/759)
 - You can no longer use `from_robj`, as the trait `FromRobj` as been removed. Instead, use `try_from`.
 - It is no longer possible to access an R integer vector as a `&[u32]` [[#767]](https://github.com/extendr/extendr/pull/767)
+- It is no longer possible to generate bindings as part of the compilation of
+extendr. Feature `non-api` is broken.
```
JosiahParry, line 44: Can we be more specific with this? What do you mean `non-api` is broken? Isn't it that it will cause warnings when used with R-devel (R >=  4.5) and may not be guaranteed to work in the future.

Cargo.toml
```diff
@@ -17,7 +17,6 @@ authors = [
 edition = "2021"
 license = "MIT"
 repository = "https://github.com/extendr/extendr"
-rust-version = "1.70"
```
JosiahParry, line 20: THIS is the hero bit. Need to add to changelog

extendr-api/Cargo.toml
```diff
@@ -8,7 +8,7 @@ license.workspace = true
 repository.workspace = true
 
 # Note: it seems cargo-msrv doesn't support rust-version.workspace = true.
-rust-version = "1.70"
+rust-version = "1.65"
```
JosiahParry, line 11: 🔥

extendr-api/Cargo.toml
```diff
@@ -42,13 +42,14 @@ full-functionality = ["graphics", "either", "ndarray", "faer", "num-complex", "s
 # Parts of the R-API are locked behind non-API, as CRAN frowns upon the presence
 # of non-API items in packages. You may enable this feature, to generate
 # bindings to these non-API items, and their usage in extendr
-non-api = ["libR-sys/non-api"]
+# BEWARE: this feature does not work, as `bindgen` use in extendr is deprecated
+non-api = []
```
JosiahParry, line 46: @CGMossa this needs to be a no-op because we don't have a way to generate the bindings via build.rs for someone who wants to use `libR-sys/non-api` at the moment. We will need to address afterwards https://github.com/extendr/libR-sys/issues/251

CHANGELOG.md
CGMossa: It does not work at all. Using it will not compile extendr. So either we remove all `non-api`-dependent code, or I don't know what to do.

I'm leaning towards deleting everything related to `non-api`.

Cargo.toml
CGMossa: I don't really think so, because extendr 0.6.0 didn't have MSRV 1.70. So it is "back to normal" 😃

extendr-api/Cargo.toml
CGMossa: It can't be a no-op. It just breaks if you add it to the compilation process.

CHANGELOG.md
JosiahParry: No, I think we keep it and back-fill it at a later point with libR-sys generated bindings. Once the C API is actually "stabilized" we can port everything over and then remove the non-api entirely.

JosiahParry, line 44:
```suggestion
extendr. Feature `non-api` is broken and will not compile. Related <https://github.com/extendr/libR-sys/issues/251>
```

CGMossa: Good! So this is resolved.. Or?

extendr-api/tests/extendr_macro.rs
```diff
@@ -130,6 +130,7 @@ fn tests_with_successful_outcomes() {
 // Win32 does not support catch_unwind.
 #[cfg(not(target_arch = "x86"))]
 #[test]
+#[ignore = "panicíng in FFI is now automatically abort instead of UB"]
```
JosiahParry, line 133:
```suggestion
// This behavior is now handled in Rust nightly 1.81 making catch_r_error() unusable.
// For previous versions this is useful. 
// See related: https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/.E2.9C.94.20.5BFFI.5D.20Is.20the.20representation.20of.20.26.5BT.3B.20N.5D.20and.20*const.20.5BT.3BN.5D
// TODO: revisit when formalized
// Win32 does not support catch_unwind.
#[cfg(not(target_arch = "x86"))]
#[test]
#[ignore = "panicking in FFI is now automatically abort instead of undefined behavior"]
```

.github/workflows/test.yml
```diff
@@ -257,187 +255,3 @@ jobs:
...
-        config:
-          - { os: ubuntu-latest,  r: 'devel',   rust-version: 'stable' }
-          - { os: macOS-latest,   r: 'devel',   rust-version: 'stable' }
-          - { os: windows-latest, r: 'release', rust-version: 'stable-msvc',  rust-targets: [ 'x86_64-pc-windows-gnu' ], rtools-version: '44' }
-          - { os: windows-latest, r: 'devel',   rust-version: 'stable-msvc',  rust-targets: [ 'x86_64-pc-windows-gnu' ], rtools-version: '44' }
-          # When the MSVC target is used, since it's cross-compilation from MSVC
-          # to GNU, the doc tests are usually skipped. Adding `-Zdoctest-xcompile`
-          # lets the doctests run, which accordingly require the nightly toolchain.
-          - { os: windows-latest, r: 'release', rust-version: 'nightly-msvc', rust-targets: [ 'x86_64-pc-windows-gnu' ], rtools-version: '44', extra-args: [ '-Zdoctest-xcompile' ] }
-          - { os: windows-latest, r: 'oldrel',  rust-version: 'nightly-msvc', rust-targets: [ 'x86_64-pc-windows-gnu' ], rtools-version: '43', extra-args: [ '-Zdoctest-xcompile' ] }
```
JosiahParry, line 284: Note that these Rust versions `nightly-msvc` with `extra-args: [ '-Zdoctest-xcompile']` are not in the remaining CI configs. Do we want to add these? 

What benefit do we get testing on nightly over release?

CGMossa: Z flags won't work on stable.

JosiahParry: Totally understand. So do we want to include these versions into the matrix config so that we get doc testing?

CGMossa: Yes. Have you looked at the steps in both jobs?

.github/workflows/test.yml
```diff
@@ -19,29 +19,47 @@ jobs:
...
+          - {
+              os: ubuntu-latest,
+              r: "release",
+              rust-version: "stable",
+              check_fmt: true,
+            }
+          - {
+              os: ubuntu-latest,
+              r: "release",
+              rust-version: "nightly-msvc",
```
CGMossa, line 56: on ubuntu, the rust-version should be `"nightly"`
```suggestion
              rust-version: "nightly",
```

.github/workflows/test.yml
```diff
@@ -19,29 +19,47 @@ jobs:
...
+              rust-targets: ["x86_64-unknown-linux-msvc"],
```
CGMossa, line 57:
```suggestion
```

JosiahParry:
Note that by removing the bindgen CI we lost doc tests. The ubuntu + rust nightly test have been adapted to also do doctests. This ensures we still have doctest coverage

JosiahParry (approved):
Well that was a lot! This is much better. Glad we can get the MSRV down. Now we just need to handle single threading PR for 0.7.0 release

PR #817, opened 2024-07-01
JOSS paper citation

CGMossa:
JOSS paper on extendr was just published.
So now we can have a badge and a citation file.


JosiahParry (approved):
LOVE IT

PR #818, opened 2024-07-01
Improve derive macros

CGMossa:
- `#[extendr]`-`impl` assumed that the `Error` available is `extendr_api::Error`. Now, it is explicit.
  - Same with `ExternalPtr`
  - Same with `Result`
- `mut` on `args` is not necessary in `wrappers.rs`

~~I've got a few more improvements that I've found that I want to migrate over...~~
Actually, I'll save that to another PR.

These were discovered when trying to update r-polars to extendr 0.7.0.
https://github.com/pola-rs/r-polars/pull/1154

I think these should be merged ASAP, and we can consider a patch-release. We'll release when a few more issues reveal themselves in the next few weeks.

### DIFF
--- a/extendr-macros/src/extendr_impl.rs
+++ b/extendr-macros/src/extendr_impl.rs
@@ -182,7 +182,7 @@ pub(crate) fn extendr_impl(
         impl TryFrom<Robj> for &#self_ty {
             type Error = extendr_api::Error;
 
-            fn try_from(robj: Robj) -> Result<Self> {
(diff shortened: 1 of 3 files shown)


JosiahParry:
Thanks for this. Waiting for the failing CI to re-run. Once that is done I'll approve this. This is super uncontroversial of a change. This will address any namespace conflicting of the `Error` type.

JosiahParry (approved):
CI looks good.

PR #819, opened 2024-07-02
CHANGELOG

CGMossa:
- Update the version (and seal previous version)
- Escape this part, as it looks weird on the website, see the first entry at this link https://extendr.github.io/changelog.html#changed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@
 
 ### Changed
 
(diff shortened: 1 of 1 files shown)


kbvernon (commented):
If you imagine that the website+pandoc issue will only come up occasionally, so that escaping won't get too annoying, then LGTM.

JosiahParry (approved):
Doc only.

PR #824, opened 2024-07-05
Don't rely on `prelude` in extendr-macros

CGMossa:
There are a few places, where we refer to objects, as if they are in scope, but they may not be.

This was seen on r-polars and is related to #818.


JosiahParry (approved):
Another uncontroversial change here. To recap, this PR explicitly namespaces from `extendr_api` for `Result`, switches `r!()` for `Robj::from()`, adjusts our tests so that `crate` is available as `extendr_api`, as well as fixes some typos along the way, and throughout just prepends `extendr_api::` just about everywhere else

Also important to note is that the macro has moved from `mod.rs` to its own `wrapper_macro.rs` file

PR #827, opened 2024-07-08
add unreachable message

kbvernon:


CGMossa:
Thanks! Every little bit helps

PR #832, opened 2024-08-03
update: Don't export `faer` and `ndarray` and their contents in `prelude`

CGMossa:
Currently, the prelude, i.e. using `use extendr_api::prelude::*;` exports the contents of both `faer` and `ndarray`. There are collisions.
 This PR merely exports the crates, and then the use may decide to include their contents downstream.
 
To obtain previous behavior, one has to do:
```rs
use extendr_api::prelude::*;
use ndarray::*;
use faer::*;
```

### DIFF
--- a/extendr-api/src/optional/ndarray.rs
+++ b/extendr-api/src/optional/ndarray.rs
@@ -4,13 +4,14 @@ Defines conversions between R objects and the [`ndarray`](https://docs.rs/ndarra
 To enable these conversions, you must first enable the `ndarray` feature for extendr:
 ```toml
 [dependencies]
-extendr-api = { version = "0.4", features = ["ndarray"] }
+extendr-api = { version = "0.7", features = ["ndarray"] }
 ```
 
 Specifically, extendr supports the following conversions:
 * [`Robj` → `ArrayView1`](FromRobj#impl-FromRobj<%27a>-for-ArrayView1<%27a%2C%20T>), for when you have an R vector that you want to analyse in Rust:
     ```rust
     use extendr_api::prelude::*;
+    use ndarray::ArrayView1;
 
     #[extendr]
     fn describe_vector(vector: ArrayView1<f64>){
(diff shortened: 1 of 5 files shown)


kbvernon (commented):
Community review here: This makes sense to me. Another motivation might be that extendr conversions for these crates are currently feature-gated. Only suggestion I have is that this would - I believe - technically count as a breaking change, so CHANGELOG.md should probably be updated, too.

CGMossa:
The CI is failing, because we test `rextendr` in our CI, and there is an example that uses `ndarray` in the test suite for `rextendr`. I've made a PR in https://github.com/extendr/rextendr/pull/371 to deal with the change once it gets merged.

extendr-api/src/prelude.rs
```diff
@@ -72,10 +72,10 @@ pub use super::scalar::*;
...
 #[cfg(feature = "ndarray")]
-pub use ::ndarray::*;
+pub use ::ndarray;
 
 #[cfg(feature = "either")]
 pub use ::either::*;
 
 #[cfg(feature = "faer")]
-pub use ::faer::*;
+pub use ::faer;
```
JosiahParry, line 81: Is this also too much? Should we even be reexporting `faer` itself? Ideally we allow faer to be specified in the `Cargo.toml` and the version used by extendr and the one in the `Cargo.toml` are compatible. We should probably have a test for that?

CGMossa: Actually, I think this is necessary for consumers of our conversions. I have helped out people before, where their code did not work, because they had _one version_ of faer in their dependencies, and the extendr-api has another.

This is a known (but not well-known) Rust constraint.

extendr-api/src/optional/ndarray.rs
```diff
@@ -4,13 +4,14 @@ Defines conversions between R objects and the [`ndarray`](https://docs.rs/ndarra
 To enable these conversions, you must first enable the `ndarray` feature for extendr:
 ```toml
 [dependencies]
-extendr-api = { version = "0.4", features = ["ndarray"] }
+extendr-api = { version = "0.7", features = ["ndarray"] }
```
JosiahParry, line 7:
```suggestion
extendr-api = { version = "0.8", features = ["ndarray"] }
```

JosiahParry (approved):
This is good for me. We are removing ndarray and faer from the prelude

extendr-api/src/optional/ndarray.rs
JosiahParry: this is purely cosmetic and most likely won't be noticed.

PR #833, opened 2024-08-03
Add missing impls to `ExternalPtr<T>`

CGMossa:
Missing impls that are useful in `#[extendr]`-impl, that should be accessible via `TryFrom<Robj>` too!

This was caught by @kbvernon and the code in the R test is from them.

Long explanation: The `#[extendr]` macro knows that it needs to convert from `&Robj`/`&mut Robj`. Usually, we don't do that. What we do is we implement things from `TryFrom<&Robj>` and piggyback that on `TryFrom<Robj>`, because `#[extendr]` usually just use `Robj` directly. 

And therefore, we forgot to include these impls.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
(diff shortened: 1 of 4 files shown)


JosiahParry:
Thank you for adding a new test!! I've updated the changelog as well.

JosiahParry (approved):

PR #835, opened 2024-08-03
Added iterator for `Strings`

CGMossa:
We already had this in `StrIter`. The universal conversion tool just needed to be used here.
In summary, this adds `.into_iter()` to `Strings`.

### DIFF
--- a/extendr-api/src/wrapper/strings.rs
+++ b/extendr-api/src/wrapper/strings.rs
@@ -159,3 +159,13 @@ impl From<Option<Strings>> for Robj {
         }
     }
 }
+
+impl IntoIterator for Strings {
+    type Item = &'static str;
+
+    type IntoIter = StrIter;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.robj.as_str_iter().unwrap()
+    }
+}


CGMossa:
Currently, the `.into_iter()` piggybacks on `.iter()` method and we get an `Iterator<Item=&Rstr>`. 
I don't know how useful that is, but it is good for `NA`-awareness amongst other things. Ideally, there should be a natural way to get `Iterator<Item=&str>`, as `str` has many useful functionality associated with it.

Overall, it is not without allocation that one can acquire the equivalent of `character()` in Rust `&[Rstr]`/`&[&str]`, so somehow I think we should also provide a way to acquire `Vec<&str>` from a `Strings`.  

Moving it back to draft for now...

JosiahParry:
This is a good PR i'm moving it to ready to review.

JosiahParry:
I've gone ahead and given this a review. I think what we need to do here is make the iterator return an owned `Rstr`

JosiahParry:
Looking at this further _all_ wrapper types provide a reference when iterating over the elements. In a way I think this is better because its a bit more performative. And when using `into_iter()` we expect to consume the item but I don't think we actually can do that in our case because the objects are owned by R

CGMossa:
@JosiahParry Hello.. So what should we do about this PR?

JosiahParry:
I think I prefer the status quo at present where we get a `&Rstr` as this gives us NA handling. This doesn't give us the NA handling and would be a fairly significant breaking change.

PR #839, opened 2024-08-31
Missing `Drop` impl on `ExternalPtr`s..

CGMossa:
If we drop `ExternalPtr` within Rust code, we don't actually then destroy the data "leaked" to R. This ensures that we do the clean-up properly, when `ExternalPtr`s goes out of scope.

### DIFF
--- a/extendr-api/src/wrapper/externalptr.rs
+++ b/extendr-api/src/wrapper/externalptr.rs
@@ -116,6 +116,23 @@ impl<T> DerefMut for ExternalPtr<T> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


CGMossa:
This PR is incomplete. I believe this is a "flaw" that we somehow need to address. The issue is, if `ExternalPtr` goes out of scope in Rust, it should destroy the data it leaked. But `ExternalPtr` goes out of scope in Rust when it passes a variable onto R.

CGMossa:
This was a flawed interpretation of `ExternalPtr<T>`. It is deallocated by R anyways, and even if it is in Rust only. It is deallocated when `gc()` is called.

PR #845, opened 2024-09-05
Add `PartialEq` for `Rstr` 

CGMossa:
Addition: It is not possible to compare `Rstr` to `Rstr` like the R interpreter does.

This PR adds that comparison. 

The case for this emerges when you want to take advantage of the fact that R has a string interner, which means that equality of strings can fall on the pointer address directly, instead of comparing the actual character bytes.

Use-case: 
Constructing a `HashSet<Rstr>` should be faster (with this) than constructing a `HashSet<&str>`. 
Caveat here is that the `Hash` for `Rstr` needs to be implemented, but this is a crucial bit...


CGMossa:
Okay, so what the changes here enable, is to use R's string-interner, as part of `Rstr`, to make them efficiently comparable.
I believe before it would be _slower_ to compare `Rstr` than it is to compare `&str`, which is unintuitive.

Let me illustrate the full case here: Let us make an function that finds the total number of unique character strings in a `character()`-vector. Currently, we offer two interfaces, the native-adjacent `Rstr` and the Rust native `&str`. 

```rs
use extendr_api::prelude::*;

#[extendr]
fn to_unique_rstr(r_char_vec: Strings) -> usize {
    let capacity = r_char_vec.len();
    let mut seen: Vec<&Rstr> = Vec::with_capacity(capacity);
    let mut n_unique = 0;
    for word in r_char_vec.iter() {
        if !seen.contains(&word) {
            n_unique += 1;
            seen.push(word);
        }
    }
    n_unique
}

#[extendr]
fn to_unique_str(r_char_vec: StrIter) -> usize {
    let capacity = r_char_vec.len();
    let mut seen: Vec<&str> = Vec::with_capacity(capacity);
    let mut n_unique = 0;
    for word in r_char_vec {
        if !seen.contains(&word) {
            n_unique += 1;
            seen.push(word);
        }
    }
    n_unique
}
```

These two implementations are **identical**, except for their different string type.
But their performance is certainly not identical:

```r
library(magrittr)
library(tidyverse)
devtools::load_all(path = "tests/extendrtests")

text_character_vector <-
  lorem::ipsum_words(1e6, collapse = FALSE)

to_unique_rstr(text_character_vector)
to_unique_str(text_character_vector)


bench::mark(
  # dplyr = dplyr::n_distinct(text_character_vector),
  rstr = to_unique_rstr(text_character_vector),
  str = to_unique_str(text_character_vector),
  check = TRUE
) -> bm_n_unique

bm_n_unique %>% 
  select(expression, is.atomic)
# # A tibble: 2 × 9
# expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time
# <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm>
# 1 rstr          539ms    539ms      1.86        0B        0     1     0      539ms
# 2 str           840ms    840ms      1.19        0B        0     1     0      840ms
```

Thus, it is twice as fast to "compare" `Rstr` than it is to compare `&str`.

extendr-api/src/iter.rs
```diff
@@ -145,6 +145,14 @@ impl TryFrom<&Robj> for StrIter {
     }
 }
 
+impl TryFrom<Robj> for StrIter {
```
JosiahParry, line 148: Can we add a doc comment here? I'm not sure where this may be used or really how.

extendr-api/src/wrapper/rstr.rs
```diff
@@ -100,13 +100,9 @@ impl Deref for Rstr {
...
 }
 
-impl<T> PartialEq<T> for Rstr
-where
-    T: AsRef<str>,
-{
-    /// Compare a `Rstr` with a `Rstr`.
-    fn eq(&self, other: &T) -> bool {
-        self.as_str() == other.as_ref()
+impl PartialEq<Rstr> for Rstr {
```
JosiahParry, line 103:
```suggestion
/// Utilize SEXP for comparison of [`Rstr`]
/// Deferring to R's comparison results in faster comparisons.
impl PartialEq<Rstr> for Rstr {
```

extendr-api/src/wrapper/rstr.rs
```diff
@@ -117,6 +113,27 @@ impl PartialEq<str> for Rstr {
     }
 }
 
+impl PartialEq<Rstr> for &str {
```
JosiahParry, line 116:
```suggestion
// casts Rstr to a string slice for comparison
// less efficient than comparing two [Rstr]
impl PartialEq<Rstr> for &str {
```

tests/extendrtests/notebooks/001_create_unique_vectors.R
JosiahParry, line 1: i think this file is accidentally committed?

JosiahParry (changes requested):

JosiahParry (approved):
Once we get CI to pass let's merge this. This is a huge performance boost.

JosiahParry:
CI is not passing due to the macro snapshot tests. The macro expansion snapshots do not appear to be identical across distro / rust version matrix.

I do not see how to fix this for the PR.

JosiahParry:
I'm going to merge this PR the current failing issues are due to an issue in R-devel https://bugs.r-project.org/show_bug.cgi?id=18806.

I have a patch to address this but it is being ignored.

PR #850, opened 2024-09-20
feat: impl Hash and Eq for Robj

avhz:
Simply add `#[derive(Eq, Hash)]` for `Robj`.

### DIFF
--- a/extendr-api/src/robj/mod.rs
+++ b/extendr-api/src/robj/mod.rs
@@ -104,6 +104,7 @@ mod tests;
 /// panics and segfaults. We will take great trouble to ensure that this
 /// is true.
 ///
+#[derive(Eq, Hash)]
 pub struct Robj {
     inner: SEXP,
 }


CGMossa:
This requires some thought.. Robj is a pointer essentially. So the hash of that is what?

avhz:
I'll take a closer look over the weekend, but I don't see an issue with hashing a pointer.

CGMossa:
I'd appreciate a review on this PR, https://github.com/extendr/extendr/pull/845. Maybe my concerns will be more visible after.

JosiahParry:
I'm reviewing #845 right now and it led me independently to implement `Hash` via deriving just like this. This derives the Hash of the pointer. This is really quite fast. Based on the examples in #845 we can count the number of unique values via hash set like so using the pointer address of an `[Rstr]`

```rs
#[extendr]
fn to_unique_hs_const(x: Strings) -> usize {
    let mut hs = HashSet::new();
    for word in x.into_iter() {
        hs.insert(word.as_ptr());
    }
    hs.len()
}
```

. If we implement `Hash` on `Robj` and `Eq` for `Rstr` the performance is faster and lets us write 

```rs
#[extendr]
fn to_unique_hs_rstr(x: Strings) -> usize {
    let mut hs = HashSet::new();
    for word in x.into_iter() {
        hs.insert(word);
    }
    hs.len()
}
```

JosiahParry (dismissed):

JosiahParry:
I think we can derive `Hash` no problem, but what I'm less convinced about right now is deriving `Eq` this will compare the pointer addresses (if i understand correctly). This will not compare whether the data are identical. I can see misuse coming from the use of `Eq` implemented on an `Robj`.

@avhz would you be okay if we removed `Eq` and merged only the `Hash` trait?

CGMossa:
I'm sorry, but I disagree on the `Hash` part.

For the `Eq` part, it essentially expands to

```rs
impl core::cmp::Eq for Robj {}
```
And since `PartialEq` is implemented for `Robj`, this works out fine.
The way `impl PartialEq for Robj` is implemented, is using an equality check from the R-API. This ensures that equality is checked regardless of the underlying type, see `R_compute_identical` of the R-API.

However, the `Hash` trait is a different issue. We need some general way to hash `Robj`s. You can't just rely on the pointer for that. Hashing two identical vectors should yield the same hash, but they would have different pointers.

avhz:
Hi everyone, just revisiting this after quite a long time. 

Can we just hash the inner SEXP instead of deriving ?

CGMossa:
I have thought about it some more. I must say, that I need to get back to you on it.. I discovered that strings can be deallocated, and reallocated. I thought R 'leaked' strings, but it doesn't! Now, I need to know if `Hash` is about the value (yes...) or the instance (??).

JosiahParry:
@CGMossa I believe that the hash must be deterministic

avhz:
I also realise that hashing the pointer and not the actual contents would essentially allow duplicate keys.

Roughly (abusing R + Rust syntax):

```rust
HashMap.from([
    (c(1,2), "Bob"),
    (c(1,2), "Jane"),
])
```

This would work if we only hashed the pointers, but it's obviously nonsense.

andy-thomason:
Finding an Robj in a hashmap isn't going to work unless the hash actually hashes
the contents of the Robj.

This means that your implementation of Hash must have a match statement that returns the hash of the `sexptype()`,
the attributes and the contents vector `as_typed_slice()`

So a Character and Bytes must both hash the bytes but be distinct.

The rule a == b => hash(a) == hash(b) must be followed.

CGMossa:
I think you're right. Plus, we need a runtime error when trying to hash externalptrs.. That's very unfortunate, but those can't be hashed, when calling through the opaque Robj.

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

CGMossa:
Hello again. I've rebased and pushed some code, as to how one could implement the hashing of the opaque, run-time typed `Robj`. There are currently no tests, and I'm unsure of many things about this implementation. But it is the thoughts that are described in this thread.
Before this piece of code could be considered, it needs
- Tests
- Figure out if `state.finish()` is necessary (I don't think so).
- Whether the remainder types, that has a `todo!` at the moment can be hashed or not.

PR #852, opened 2024-10-08
Add type-checking to `#[extendr]`-impl

CGMossa:
I'm finally back to working on extendr.

The idea of type-checking `#[extendr]`-impl/`ExternalPtr<T>` was brought up again.
I recall that I made a fairly clever strategy to embed `std::any::TypeId`, to use as a mechanism
for checking Rust types retrieved from R.
Comparing `TypeId` should be more than sufficient.

Instead of embedding it in all sorts of convoluted ways, I just figured that we can
embed it as a `Raw`-blob, and comparing this with other `Raw`-blob is good enough for purpose.
We could of course make this more efficient, but not by all that much.

**This PR is not done, and won't be done unless design decisions are made on how this should work**

Therefore, I'd like some discussion on this, and thoughts, and I will try to implement these ideas in this PR.

## Example output

```r
> MyClass$new()
 [1] 66 73 05 dc d3 37 0a 0f d8 88 c5 b2 cb d6 1a ba
<pointer: 0x1583e6880>
attr(,"rust_type_id")
 [1] 66 73 05 dc d3 37 0a 0f d8 88 c5 b2 cb d6 1a ba
attr(,"class")
[1] "MyClass"
```

### DIFF
--- a/extendr-api/src/wrapper/externalptr.rs
+++ b/extendr-api/src/wrapper/externalptr.rs
@@ -271,9 +271,10 @@ impl<T: 'static> TryFrom<&mut Robj> for &mut ExternalPtr<T> {
         }
         let raw_type_id = value.get_attrib("rust_type_id").unwrap();
         let raw_type_id = raw_type_id.as_raw_slice().unwrap();
+        // these lines extract the type ID of `T` and store is as a slice for us to compare against
         let expected_type_id = std::any::TypeId::of::<T>();
-        let expected_type_id = unsafe { std::mem::transmute::<_, [u8; 16]>(expected_type_id) };
-        let expected_type_id = expected_type_id.as_slice();
+        let expected_type_id = unsafe { std::mem::transmute::<_, [u8; 16]>(expected_type_id) }.as_slice();
+        // now we compare the expected type ID with what we found. If they do not match, we return an error
         if expected_type_id != raw_type_id {
             return Err(Error::ExpectedExternalPtrType(
                 value.clone(),
--- a/extendr-api/src/wrapper/externalptr.rs
+++ b/extendr-api/src/wrapper/externalptr.rs
@@ -238,10 +238,9 @@ impl<T: 'static> TryFrom<&Robj> for &ExternalPtr<T> {
         }
 
         // type checking that could be optional
-        if !value.has_attrib("rust_type_id") {
+        let Some(raw_type_id) = value.get_attrib("rust_type_id") else {
             return Err(Error::ExpectedExternalPtr(value.clone()));
-        }
-        let raw_type_id = value.get_attrib("rust_type_id").unwrap();
+        };
         let raw_type_id = raw_type_id.as_raw_slice().unwrap();
         let expected_type_id = std::any::TypeId::of::<T>();
         let expected_type_id = unsafe { std::mem::transmute::<_, [u8; 16]>(expected_type_id) };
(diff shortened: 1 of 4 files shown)


extendr-api/src/wrapper/externalptr.rs
```diff
@@ -109,14 +109,14 @@ impl<T> Deref for ExternalPtr<T> {
...
-impl<T> DerefMut for ExternalPtr<T> {
+impl<T: 'static> DerefMut for ExternalPtr<T> {
     /// This allows us to treat the Robj as if it is the mutable type T.
     fn deref_mut(&mut self) -> &mut Self::Target {
         self.addr_mut()
     }
 }
 
-impl<T> ExternalPtr<T> {
+impl<T: 'static> ExternalPtr<T> {
```
JosiahParry, line 119: I'm worried about these bounds here—though I'm not entirely sure what it implies. Does the static bound mean that the _type_ or the struct must be alive for the duration of the process?

CGMossa: The issue is, that the bounds come from `std::any::TypeId::of`. So they are not optional. But they were present before, I just removed them, when we didn't use the `std::any::type_name` or the other `TypeId` thing.

I'm also not sure what `T: 'a` means. I don't think it means anything in particular here either.

CGMossa: I asked about it in the Rust discord and I got many interesting answers.
link: https://discord.com/channels/442252698964721669/443150878111694848/1293353783178428417

> It basically means that the type doesn't hold any borrows that are non-'static lifetime
This includes any type that doesn't borrow at all

> Or in practical terms, it means that you can hold on to that value as long as you like without fear that anything it borrows might go away
Which for example is important for threads, since they can run long after the calling function has returned

> for TypeId specifically, it's important because lifetimes are erased before types are given ids 
so to prevent you from freely casting &'a T to &'static T, Any and TypeId require 'static

CGMossa: I believe what it simply means, is that TypeId contains the information of lifetimes as well as types, and thus, it becomes part of the typeid.

JosiahParry: I followed up with this conversation and it makes sense. Thank you!

extendr-api/src/wrapper/externalptr.rs
```diff
@@ -222,29 +229,63 @@ impl<T> ExternalPtr<T> {
...
     fn try_from(value: &mut Robj) -> Result<Self> {
         if !value.is_external_pointer() {
             return Err(Error::ExpectedExternalPtr(value.clone()));
         }
+
+        // type checking that could be optional
+        if !value.has_attrib("rust_type_id") {
+            return Err(Error::ExpectedExternalPtr(value.clone()));
+        }
+        let raw_type_id = value.get_attrib("rust_type_id").unwrap();
```
JosiahParry, line 272: Could this be simplified as 
```suggestion
        let Some(raw_type_id) = value.get_attribu("rust_type_id") else {
           return Err(Error::ExpectedExternalPtr(value.clone()));
        };
```

extendr-api/src/wrapper/externalptr.rs
```diff
@@ -222,29 +229,63 @@ impl<T> ExternalPtr<T> {
...
+        let raw_type_id = raw_type_id.as_raw_slice().unwrap();
+        let expected_type_id = std::any::TypeId::of::<T>();
+        let expected_type_id = unsafe { std::mem::transmute::<_, [u8; 16]>(expected_type_id) };
+        let expected_type_id = expected_type_id.as_slice();
+        if expected_type_id != raw_type_id {
```
JosiahParry, line 277:
```suggestion
        // these lines extract the type ID of `T` and store is as a slice for us to compare against
        let expected_type_id = std::any::TypeId::of::<T>();
        let expected_type_id = unsafe { std::mem::transmute::<_, [u8; 16]>(expected_type_id) }.as_slice();
        // now we compare the expected type ID with what we found. If they do not match, we return an error
        if expected_type_id != raw_type_id {
```

CGMossa:
The core idea here does not work. I have to close it.

The issue is `TypeId` changes definition across Rust compilers, and I have to hard-code
how many bytes to expect for `TypeId`.

PR #853, opened 2024-10-09
Try again: Type-checking `ExternalPtr<T>` by downcasting..

CGMossa:


CGMossa:
I finally see the issue here. We always thought of `ExternalPtr` as a facility for `#[extendr]`-impl, but `ExternalPtr` needs the type-checking to stand on its own. Let me show the example that illustrates this. Imagine you've got these two functions

https://github.com/extendr/extendr/blob/caf485e27ddfffff5e1cc228d97d49faff8f395a/tests/extendrtests/src/rust/src/externalptr.rs#L92-L100

And one may accidentally use this 
```r
 externalptr_numeric <- create_numeric_externalptr(as.numeric(1:10)) 
 sum_integer_externalptr(externalptr_numeric) 
```


I'm unsure what happens today, unfortunately. I suspect we do not error, before using this information,
in which then a panic ensues unpredictably.

JosiahParry:
This compiles and runs well in {b64} this permits the use of `ExternalPtr<Engine>` as a function argument instead of `Robj` previously there would be a messy panic. 
image

JosiahParry (approved):

PR #854, opened 2024-11-26
Generalize `HashMap` conversion

CGMossa:
A named list can be thought of as a `HashMap<S, Robj>` where `S` is a string type, and `Robj` is an opaque type representing arbitrary R data. One may liken a `list()` to only having say `Doubles`s, where in this case, the desired representation would be `HashMap<S, Doubles>`. While this may seem as a limited example, it probably covers a lot of use cases for `list()`. This PR provides this conversion on `Robj` and `List` through the `TryFrom`-interface. It can also be used in `#[extendr]`-`fn`.

### DIFF
--- a/extendr-api/src/robj/try_from_robj.rs
+++ b/extendr-api/src/robj/try_from_robj.rs
@@ -468,15 +468,6 @@ impl TryFrom<&Robj> for Rcplx {
 // Convert TryFrom<&Robj> into TryFrom<Robj>. Sadly, we are unable to make a blanket
 // conversion using GetSexp with the current version of Rust.
 macro_rules! impl_try_from_robj {
-    () => {};
-    (&mut [$type:ty], $($rest:tt)*) => {
-        impl_try_from_robj!(&mut [$type]);
-        impl_try_from_robj!($($rest)*);
-    };
-    ($type:ty, $($rest:tt)*) => {
-        impl_try_from_robj!($type);
-        impl_try_from_robj!($($rest)*);
-    };
     (&mut [$type:ty]) => {
         impl TryFrom<Robj> for &mut [$type] {
             type Error = Error;
@@ -506,9 +497,8 @@ macro_rules! impl_try_from_robj {
             }
         }
     };
-
-    ($type:ty) => {
-        impl TryFrom<Robj> for $type {
+    ($(@generics<$generics:tt>)? $type:ty $(where $($where_clause:tt)*)?) => {
+        impl$(<$generics>)? TryFrom<Robj> for $type $(where $($where_clause)*)? {
             type Error = Error;
 
             fn try_from(robj: Robj) -> Result<Self> {
@@ -516,7 +506,7 @@ macro_rules! impl_try_from_robj {
             }
         }
 
-        impl TryFrom<&Robj> for Option<$type> {
+        impl$(<$generics>)? TryFrom<&Robj> for Option<$type> $(where $($where_clause)*)? {
             type Error = Error;
 
             fn try_from(robj: &Robj) -> Result<Self> {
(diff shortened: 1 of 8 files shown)


extendr-api/src/robj/try_from_robj.rs
```diff
@@ -537,45 +527,112 @@ macro_rules! impl_try_from_robj {
...
+impl_try_from_robj!(&mut [Rint]);
+impl_try_from_robj!(&mut [Rfloat]);
+impl_try_from_robj!(&mut [Rbool]);
+impl_try_from_robj!(&mut [Rcplx]);
+impl_try_from_robj!(&mut [u8]);
+impl_try_from_robj!(&mut [i32]);
+impl_try_from_robj!(&mut [f64]);
+
+impl_try_from_robj!(&str);
+impl_try_from_robj!(String);
```
JosiahParry, line 579: @CGMossa This appears to be a formatting change of the previous single macro invocation. is that correct?

CGMossa: Yes. Instead of providing all the types at once, it was now necessary to call the macro on each type separately.

extendr-api/src/wrapper/list.rs
```diff
@@ -88,12 +88,17 @@ impl List {
     ///     assert_eq!(names, vec!["a", "b"]);
     /// }
     /// ```
-    pub fn from_hashmap<K>(val: HashMap<K, Robj>) -> Result<Self>
+    pub fn from_hashmap<K, V>(val: HashMap<K, V>) -> Result<Self>
     where
+        V: IntoRobj,
```
JosiahParry, line 93: Should this be `TryFrom`?

CGMossa: I don't think so. All Rust types can be made into R types, although not uniquely. So the conversion is not falliable. 

What I should have done is this for clarity

```diff
diff --git a/extendr-api/src/wrapper/list.rs b/extendr-api/src/wrapper/list.rs
index 4f3a0bbc7..87a53e476 100644
--- a/extendr-api/src/wrapper/list.rs
+++ b/extendr-api/src/wrapper/list.rs
@@ -90,13 +90,11 @@ impl List {
     /// ```
     pub fn from_hashmap<K, V>(val: HashMap<K, V>) -> Result<Self>
     where
-        V: IntoRobj,
+        V: Into<Robj>,
         K: Into<String>,
     {
-        let (names, values): (Vec<_>, Vec<_>) = val
-            .into_iter()
-            .map(|(k, v)| (k.into(), v.into_robj()))
-            .unzip();
+        let (names, values): (Vec<_>, Vec<_>) =
+            val.into_iter().map(|(k, v)| (k.into(), v.into())).unzip();
         let mut res: Self = Self::from_values(values);
         res.set_names(names)?;
         Ok(res)

```

JosiahParry: Can we make a follow up issue to trash this? I think it would be good. In a test case I'm about to push i have 

```rs

impl TryFrom<Robj> for Point {
    type Error = Error;

    fn try_from(value: Robj) -> std::result::Result<Self, Self::Error> {
        let inner_vec = Doubles::try_from(value)?;
        let x = inner_vec[0].inner();
        let y = inner_vec[1].inner();
        Ok(Point { x, y })
    }
}

impl From<Point> for Doubles {
    fn from(value: Point) -> Self {
        Doubles::from_values([value.x, value.y])
    }
}

impl From<Point> for Robj {
    fn from(value: Point) -> Self {
        Robj::from(Doubles::from(value))
    }
}
#[extendr]
fn test_hm_custom_try_from(mut x: HashMap<&str, Point>) -> List {
    x.insert("inserted_value", Point {x : 3.0, y: 0.1415});
    List::from_hashmap(x).unwrap()
}
```

which works quite well

JosiahParry (approved):
I've added tests to this PR and assuming CI is amenable to these changes I think it is ready for merge.

CGMossa:
@JosiahParry This one is also seems to be lost to CI mania.

PR #855, opened 2024-11-26
Soundness: `Robj` must be transparent

CGMossa:
It is possible to use `std::mem::transmute` to go from a `SEXP` to, say, `&Robj`/`&mut Robj`. This is indeed necessary to make some API possible. See `Index` on `List` as an example. A follow-up PR for `Index` on `List` will come..

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@
 - `TryFrom<Robj>` is now implemented for `&ExternalPtr<T>` and `&mut ExternalPtr<T>` <https://github.com/extendr/extendr/pull/833>
 - `Option<T>` can now be returned from `#[extendr]` annotated functions for extendr_api types. `None` is translated into `NULL` value in R [[#802]](https://github.com/extendr/extendr/pull/802)
 
-
 ### Changed
 
 - Enhancement: string comparisons are now implemented for `Rstr` using R's string intern mechanism. Making string comparisons _much_ faster. [[#845]](https://github.com/extendr/extendr/pull/845)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -20,6 +19,9 @@ return `Option<Strings>` instead of opaque `Robj`.
 
 ### Deprecated
 
+- Removed `from_sexp_ref` from public API. This is a completely unsound function, and uses of it must not
```
Ilia-Kosenkov, line 22: Can you rephrase this please? I think something is not right in this sentence.

JosiahParry: What about: 

```
- Removed `from_sexp_ref` from the public API. `from_sexp_ref()` is an unsafe method and is intended for use only within `extendr-api`.

Comment (JosiahParry) [inline CHANGELOG.md:22]:
```suggestion
- Removed `from_sexp_ref()` from the public API. `from_sexp_ref()` is an unsafe method and is for internal use only.
```

CHANGELOG.md
```diff
@@ -20,6 +19,9 @@ return `Option<Strings>` instead of opaque `Robj`.
...
+used otherwise. [[#855]](https://github.com/extendr/extendr/pull/855)
```
JosiahParry, line 23:
```suggestion
```

JosiahParry (approved):
Tests that fail are on r-devel due to BDR's new check-rust function. We have an existing patch but it is being ignored. https://bugs.r-project.org/show_bug.cgi?id=18806

This is good to merge now. Thanks for the help everyone

PR #856, opened 2024-11-26
Add conversions into `[T;N]`

CGMossa:
Work in progress


JosiahParry:
Thanks! I'm curious, what do you see the use case is for u8? Would this be from `Raws`?

CGMossa:
Yes! U8 is bytes, same as raw in R.

extendr-api/src/error.rs
```diff
@@ -170,6 +172,8 @@ impl std::fmt::Display for Error {
                 write!(f, "It is only possible to return a reference to self.")
             }
             Error::NoGraphicsDevices(_robj) => write!(f, "No graphics devices active."),
+            // this is very unlikely to occur, and it would just say: Rust error: could not convert slice to array
+            Error::TryFromSliceError(std_error) => write!(f, "Rust error: {}", std_error),
```
JosiahParry, line 176: I think we could improve this error message. Particularly if we want to encourage users to use arrays in function arguments. 

```rs
#[extendr]
/// @export
fn round_trip_array_f64(x: [f64; 4]) -> [f64; 4] {
    x
}
```

Supplying this with a vector length < 4 returns the error 
```
Error in `round_trip_array_f64()`:
! could not convert slice to array
```
which isn't clear from an R user perspective.

CGMossa: I'll fix that. I didn't think that error would ever appear.. I don't know why I thought that.. I am sorry for the obvious fail here.

JosiahParry: All good! Do you think we can do a length check or something similar?

CGMossa: Yeah. Let me get back to you. I might have lost a commit because I could swear I did this... Or something similar... 😅

I'll ping you later on this.

JosiahParry:
@CGMossa I've added tests for the tuple array conversions. Please @ me when you're able to change the error message then I can give it the 👍🏼

CGMossa:
Yes! Can you please avoid exporting these test functions. It is not necessary to run tests, and then we avoid issues with lack of doc strings.

JosiahParry (approved):
@CGMossa I've added a type check here: 

```rust
if value.len() != N {
    return Err(Error::ExpectedLength(N));
}
```

which gives us the standard error message. 

I've removed the `/// @export` and have updated the snapshots. Once CI is looking good (as good as it can due to changes in devel and cargo msrv) we can merge.

CGMossa:
Thank you. And yes, that's what I would've done.

CGMossa:
Confirming (manually) that MSRV is preserved:
```sh
extendr-main on  from_array_and_tuple [$⇡] is 📦 v0.7.1 via 🦀 v1.83.0 
❯ cargo msrv --path extendr-api verify
  [Meta]   cargo-msrv 0.17.1                                                                                                                               

Compatibility Check #1: Rust 1.65.0
  [OK]     Is compatible 
```

PR #857, opened 2024-11-26
Added conversions to tuples of up to size 12

CGMossa:
Spoiler:
image

The R construct `list()` is a very special. It is hard to find an equivalent Rust type.
First, it is a contiguous data-layout, but it is the pointers, rather than actual data. The Rust equivalent could be `Vec<Box<dyn std::any::Any>>`. That is to say, it stores a list owned pointers, to any kind of data. I could make other Rust suggestions that would resemble `list()` better, but this one would do for now.

The issue is the heterogeneity of the stored elements. They do not have to be the same thing.
A fixed sized version of this is Rust tuples. A tuple of a certain size can contain any type, e.g. `(i32, i32, f64)`. Is equivalent to the unnamed list that contains `list(integer(), integer(), numeric())`. 

This PR adds conversions from `list(...)` to tuples. The tuple is of size 12, as this is chosen number by the `std` as well. 

It is important to say, that if `Robj` is any other type, like `integers(n=4), that does not mean the conversion would then target `(i32, i32, i32, i32)`. `Robj` must be `list()`/[`List`](https://extendr.github.io/extendr/extendr_api/wrapper/list/struct.List.html). If you're looking for such a conversion, see #856 instead, as that is more appropriate.

- [x] ~~Things to do in a follow-up PR: All the extra functionality provided by `impl_try_from_robj`, such as `Option<(T0, ..., Tn)>`, etc.~~ This was necessary to make the PR work with `#[extendr]`-macro, so it is also applied.


JosiahParry:
One interesting thing that I ran into using this is that it requires that the `TryFrom<&Robj>` and does not support when an owned TryFrom impl is provided. 

Should we also support when `TryFrom<Robj>` is present? 

```rust
impl TryFrom<&Robj> for Point {
    type Error = extendr_api::Error;
    fn try_from(value: &Robj) -> std::result::Result<Self, Self::Error> {
        let dbl_vec = Doubles::try_from(value)?;
        let x = dbl_vec[0].inner();
        let y = dbl_vec[1].inner();
        Ok(Point {x, y})

    }
}

/// @export
#[extendr]
fn sum_points(x: (Point, Point)) -> Doubles {
    let Point { x:x1, y:y1 } = x.0;
    let Point { x:x2, y:y2} = x.1;

    Doubles::from_values([x1+x2, y1+y2])
}
```

CGMossa:
I already did that.. Right?

CGMossa:
I understand now what you meant. One thing I've noticed about the extendr-api, is that it implements all conversions on `&Robj`,  and merely invoke the `TryFrom<&Robj>` method in `TryFrom<Robj>`-impl.
I don't think we can offer the flexibility that this particular impl requires `TryFrom<Robj>` instead of requiring the `TryFrom<&Robj>`. Furthermore, we `#[extendr]`-`fn` requires `TryFrom<Robj>`.

I've been contemplating adding another macro that can be used to assist with this boilerplate for users. But that has to be in follow-up PRs, and when I'm more confident as to what the constraints are. 

@JosiahParry

JosiahParry:
> "Furthermore, we #[extendr]-fn requires TryFrom<Robj>."

Exactly, this feels like a weird differentiation. I'm not sure how to rectify it. 

One thing that can be done—I think—is implement `impl<T: TryFrom<Robj>> TryFrom<&Robj> for T` or something like that?

CGMossa:
It's a bit more convoluted. I am all for unifying things, but I don't see a path forward for this, at the moment.

JosiahParry:
i've pushed a change to add testthat tests to ensure that this conversion works with a TryFrom implementaiton and that the length error works as anticipated

JosiahParry:
Oops! Need to update snapshots.

CGMossa:
Perfect! CI runs now, for the most part. MSRV is an error outside of this PR, and R-devel is problematic due to.. R-devel treatment of rust projects.

PR #858, opened 2024-11-26
Added `Index`/`IndexMut` for `List`.

CGMossa:


extendr-api/src/wrapper/list.rs
```diff
@@ -175,6 +175,42 @@ impl List {
...
+        } else {
+            unsafe {
+                let sexp = VECTOR_ELT(self.robj.get(), i as R_xlen_t);
+                let robj_ref = std::mem::transmute::<SEXP, &Robj>(sexp);
+                // Despite the surrounding list being protected, we ought to protect
+                // the extracted elements, just in case that the list goes out of scope,
+                // but the element extracted is still in play.
+                // Although the lifetime annotation ought to catch this issue at compile time.
+                ownership::protect(sexp);
+                Ok(robj_ref)
```
JosiahParry, line 191: When the references `Robj` goes out of scope, the `sexp` will be decremented right? My only worry is that this will increment the protection and never decrement it

CGMossa: The transmute is avoiding the protection call that is seen in from_sexp.

So by adding it here, I believe I've covered it.

CGMossa: I do think we can get away with it not being there. The List itself is protected. The element is not. But the reference will only be valid as long as the reference to the list is still present. But I'm not sure that rust lifetimes truly translate to Robj protection just yet...

So I'd vote for also protecting the elements, of protected lists, for now.

JosiahParry: I'm fine with additional protection so long as that it is possible for it to be unprotected. My only worry is that we increment the ownership Rc without there being a way for the reference to be decremented so it is held in memory longer than needed. Though I'm unsure if that would be the case here.

CGMossa:
Currently, there is a way to index a `List`, and you will get an object back. There is no `Index`/`IndexMut`, but somehow, the `Deref`-trait comes into play here. I believe the `[index]` is inducing a `&Self`.

PR #860, opened 2024-12-03
Improve_macro_docs

CGMossa:


JosiahParry (dismissed):
We should likely have a sprint to add doc examples **and** use them as tests rather than ignoring them. 

It would be good for users to see these in the docs ad be able to use them as real examples.

JosiahParry (approved):

PR #861, opened 2024-12-13
Hotfix: Mishandling `NULL` pointer

CGMossa:
In #853, type-checking was added to `ExternalPtr<T>`. Previously, there were issues with processing `NULL` pointers, that we covered through a test. This test resided in `test-classes.R` of `extendrtests`. While the test still logically made sense after #853, the test would _panic_. For some reason, the `panic!` (due to an `unwrap()`) would not fail the test. We believe that this was unnoticed, because there was no signal of failure, outside the print log. The print log of R tests are not visible in the CI.

Now, the case of `NULL` pointer is handled in the type-checking code. It was an oversight, made possible by me being overly naive. But now, it is handled properly.

This is not a critical issue, but many exotic extendr code relies on passing of `NULL` pointers, so in a sense, it is critical to advanced use-cases.

### DIFF
--- a/extendr-api/src/wrapper/externalptr.rs
+++ b/extendr-api/src/wrapper/externalptr.rs
@@ -247,7 +247,7 @@ impl<T: 'static> TryFrom<&Robj> for &ExternalPtr<T> {
                 .external_ptr_addr::<Box<dyn Any>>()
                 .cast_const()
                 .as_ref()
-                .unwrap()
+                .ok_or_else(|| Error::ExpectedExternalNonNullPtr(value.clone()))?
         };
 
         if boxed_ptr.downcast_ref::<T>().is_none() {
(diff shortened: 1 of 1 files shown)


JosiahParry:
This looks good. Any idea why the cargo msrv check is failing?

CGMossa:
I don't know what is going on with CI. It works locally. The new version of `cargo-msrv` have changed their print

```sh
extendr-main on  nightly_crash [$] is 📦 v0.7.1 via 🦀 v1.83.0 took 1m45s 
❯ cargo extendr msrv             
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.08s
     Running `target/debug/xtask msrv`
$ cargo-msrv --path extendr-api verify
  [Meta]   cargo-msrv 0.17.1                                                                                                       

Compatibility Check #1: Rust 1.65.0
  [OK]     Is compatible 

extendr-main on  nightly_crash [$] is 📦 v0.7.1 via 🦀 v1.83.0 took 2s 
❯ echo $?
0

```

The local version that still worked is
```sh

extendr-main on  nightly_crash [$] is 📦 v0.7.1 via 🦀 v1.83.0 took 8s 
❯ cargo msrv --version
cargo-msrv 0.15.1
```

CGMossa:
@JosiahParry I've updated the top post here. I think this should be merged regardless of the MSRV CI failure.

JosiahParry (approved):

PR #862, opened 2024-12-14
CI: fix MSRV broken

CGMossa:
In the new version of `cargo-msrv`,  there are many changes, and one of them is, that `--output-format=minimal` is the preferred output for CI usage.

But that does not fix the failure on CI.

Locally, the output is

```sh
 "Verify Rust MSRV"
::group::Verify Rust MSRV
Running cargo msrv --path extendr-api --output-format minimal verify
true
::endgroup::
```

However, in CI, it appears to be `false`.  I don't know why.


CGMossa:
I'm given up on this.

PR #865, opened 2024-12-18
Replaced the macro-based implementation, to one relying on trait-based (with `AsTypedSlice`)

CGMossa:


CGMossa:
I believe this is a net positive change. I will revisit this when appropriate.

PR #871, opened 2025-01-03
feat: `new_na_filled()` associated method

JosiahParry:
This PR adds associated methods to instantiate vectors that are filled with NA values.

### DIFF
--- a/extendr-api/src/wrapper/macros.rs
+++ b/extendr-api/src/wrapper/macros.rs
@@ -38,9 +38,16 @@ macro_rules! gen_vector_wrapper_impl {
                 #[doc = "   assert_eq!(vec.len(), 10);"]
                 #[doc = "}"]
                 #[doc = "```"]
+                /// Constructs an empty vector of size `len` with default values
                 pub fn new(len: usize) -> $type {
                     // TODO: Check if impacts performance.
                     let iter = (0..len).map(|_| <$primitive_type>::default());
+                    <$type>::from_values(iter)
+                }
+
+                /// Constructs a new vector of size `len` with `NA` values
+                pub fn new_na_filled(len: usize) -> $type {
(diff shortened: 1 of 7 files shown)


extendr-api/src/wrapper/macros.rs
```diff
@@ -38,11 +38,18 @@ macro_rules! gen_vector_wrapper_impl {
                 #[doc = "   assert_eq!(vec.len(), 10);"]
                 #[doc = "}"]
                 #[doc = "```"]
+                /// Constructs an empty vector of size `len` with default values
                 pub fn new(len: usize) -> $type {
                     // TODO: Check if impacts performance.
                     let iter = (0..len).map(|_| <$primitive_type>::default());
                     <$type>::from_values(iter)
                 }
```
JosiahParry, line 46: as an aside: @CGMossa should this use [`Rf_allocVector()`](https://docs.rs/libR-sys/latest/libR_sys/fn.Rf_allocVector.html) instead?

CGMossa: In short, yes.

Long answer: This roundabout way of doing this initialisation is quite confusing, but it was _meant_ to be a gateway to implementing
iterators as altrep objects. That vision never materialised, and actually it has hurt us somewhat...

If you follow the calls, eventually you'll find something that is less efficient than just called `allocVector` directly. In fact, you'll
find `allocVector`, but then on top, you'll also see that each element will be set individually. Which is not redundant...

Also, there is precedent for such a functionality, see https://github.com/extendr/extendr/blob/6250cb73441b6d3b5b9ceba82d20daa558daa5fc/extendr-api/src/wrapper/matrix.rs#L101-L113

JosiahParry: I'll rename and update the PR thank you for finding that.

Issue #538, opened 2023-04-27
Allow multiple impls for the same struct in different files


daniellga:
Would it be possible for you to implement multiple `impl` to the same struct in different files? Here are the errors I get.

lib.rs
```
use extendr_api::prelude::*;

mod test1;
mod test2;

extendr_module! {
    mod library_name;
    use test1;
    use test2;
}

```

test1.rs
```
use extendr_api::prelude::*;

pub struct Lol;

#[extendr]
impl Lol {
    fn ttt() {}
}

extendr_module! {
    mod test1;
    impl Lol;
}

```

test2.rs
```
use extendr_api::prelude::*;

use crate::test1::Lol;

#[extendr]
impl Lol {
    fn zzz() {}
}

extendr_module! {
    mod test2;
    impl Lol;
}
```

Error:
```
error[E0119]: conflicting implementations of trait `extendr_api::FromRobj<'_>` for type `&Lol`
 --> src/test2.rs:5:1
  |
5 | #[extendr]
  | ^^^^^^^^^^ conflicting implementation for `&Lol`
  |
 ::: src/test1.rs:5:1
  |
5 | #[extendr]
  | ---------- first implementation here
  |
  = note: this error originates in the attribute macro `extendr` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0119]: conflicting implementations of trait `extendr_api::FromRobj<'_>` for type `&mut Lol`
 --> src/test2.rs:5:1
  |
5 | #[extendr]
  | ^^^^^^^^^^ conflicting implementation for `&mut Lol`
  |
 ::: src/test1.rs:5:1
  |
5 | #[extendr]
  | ---------- first implementation here
  |
  = note: this error originates in the attribute macro `extendr` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0119]: conflicting implementations of trait `From<Lol>` for type `extendr_api::Robj`
 --> src/test2.rs:5:1
  |
5 | #[extendr]
  | ^^^^^^^^^^ conflicting implementation for `extendr_api::Robj`
  |
 ::: src/test1.rs:5:1
  |
5 | #[extendr]
  | ---------- first implementation here
  |
  = note: this error originates in the attribute macro `extendr` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0119]: conflicting implementations of trait `From<&Lol>` for type `extendr_api::Robj`
 --> src/test2.rs:5:1
  |
5 | #[extendr]
[...]

PR #878, opened 2025-02-15
Allowing multiple `impl Foo` blocks with `#[extendr]`

jtlandis:
This PR aims to address #538. As a recap, the `#[extendr]` module implements the conversions to an Robj when used on an `impl <T>` block. This causes a conflicts in implementations when trying to have multiple `impl` blocks tagged with `#[extendr]` across the code base.

The approach taken was to introduce a new optional key in `#[extendr(impl_only = true)]` which will exclude implementing the conversions for that block. Additionally, changes were made to `extendr-api/scr/metadata.rs` so that the `struct Impl` type is aware of if the methods are the only thing required and thus the wrappers do not overwrite each other.

This makes the following syntax valid 

```rust

use extendr_api::prelude::*;

struct Counter {
    n: i32,
}

mod create {
    use super::*;

    #[extendr]
    impl Counter {
        fn new() -> Self {
            Self { n: 0 }
        }
    }

    extendr_module! {
      mod create;
      impl Counter;
    }
}

mod increment {
    use super::*;

    #[extendr(impl_only = true)]
    impl Counter {
        fn get(&self) -> i32 {
            self.n
        }

        fn increment(&mut self) {
            self.n += 1
        }
    }

    extendr_module! {
      mod increment;
      impl Counter;
    }
}

extendr_module! {
  mod counter;
  use create;
  use increment;
}
```

The above would be invalid if the user attempted to tag, `#[extendr]` within the same mod. Thus the convention from this PR would be one `#[extendr]` tag  per impl block of a given type per module.

@CGMossa has also suggested removing the conversions from the impl blocks and instead requiring them on the struct itself. This PR does not take this approach, but I wouldn't mind updating the PR to do so.

### DIFF
--- a/extendr-macros/src/extendr_options.rs
+++ b/extendr-macros/src/extendr_options.rs
@@ -5,6 +5,7 @@ pub(crate) struct ExtendrOptions {
     pub r_name: Option<String>,
     pub mod_name: Option<String>,
     pub use_rng: bool,
+    pub impl_only: bool,
 }
 
 impl ExtendrOptions {
@@ -14,6 +15,7 @@ impl ExtendrOptions {
     ///
     /// - `r_name = "name"` which specifies the name of the wrapper on the R-side.
     /// - `use_rng = bool` ensures the RNG-state is pulled and pushed
+    /// - `impl_only = bool` create wrappers for impl functions only
     ///
     pub fn parse(&mut self, meta: ParseNestedMeta) -> syn::parse::Result<()> {
         let value = meta.value()?;
@@ -45,6 +47,14 @@ impl ExtendrOptions {
                     Ok(())
                 } else {
                     Err(value.error("`use_rng` must be `true` or `false`"))
+                }
+            }
+            "impl_only" => {
+                if let Ok(LitBool { value, .. }) = value.parse() {
+                    self.impl_only = value;
+                    Ok(())
+                } else {
+                    Err(value.error("`impl_only` must be `true` or `false`"))
                 }
             }
             _ => Err(syn::Error::new_spanned(meta.path, "Unexpected key")),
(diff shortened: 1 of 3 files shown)


jtlandis:
I have been fiddling around with taking the approach @CGMossa indicated except for creating  any label key to the `#[extendr]` proc_macro.

If the devs would prefer that, I can open a different PR and close this one. 👍

CGMossa:
Hello! Thanks for the PR! I will respond eventually, I'm just tied up these days..

I think it is unlikely that we will go with this method. I can explain, but will need some time to write up the example, where this ends up with a strong limitation..

I'm very happy to see the PR though and I'm also happy that my words are read, regardless of how incoherent they are.


I never finished the label strategy, but I think that one might work! If you can devise a PR for that, that would be great. I think it is important that a test is present for such PRs, but I will help making them, with you, if you like.

Cheers

CGMossa:
I thought that nested modules cannot access private fields. But I was wrong. This should be reviewed and probably merged.
Can you make a test inside of `tests/extendrtests/` please?

extendr-macros/src/extendr_options.rs
```diff
@@ -14,6 +15,7 @@ impl ExtendrOptions {
     ///
     /// - `r_name = "name"` which specifies the name of the wrapper on the R-side.
     /// - `use_rng = bool` ensures the RNG-state is pulled and pushed
+    /// - `impl_only = bool` create wrappers for impl functions only
```
JosiahParry, line 18: What would the alternative be here? If this is false what do you get?

jtlandis:
@JosiahParry Just to make sure that I don't waste your time, #881 is in my opinion a better version of this PR, and also walks back the feature of having "impl_only" as an option in the proc macro.

JosiahParry:
@jtlandis hah thanks. I thought i was in this PR. should we close this one in favor of #881 ?

jtlandis:
I believe so, thanks!

PR #879, opened 2025-02-16
Chore/add from iterator implementations

Ilia-Kosenkov:
Housekeeping: add missing `FromIterator` implementation for vector types.
Allows for, e.g.,
```rust
let ints : Integers = (0..3).collect();
```

### DIFF
--- a/extendr-api/src/wrapper/macros.rs
+++ b/extendr-api/src/wrapper/macros.rs
@@ -148,7 +148,7 @@ macro_rules! gen_vector_wrapper_impl {
                 $type { robj }
             }
         }
-        
+
          impl FromIterator<$primitive_type> for $type {
             /// A more generalised iterator collector for small vectors.
             /// Generates a non-ALTREP vector.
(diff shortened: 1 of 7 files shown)


extendr-api/src/wrapper/macros.rs
```diff
@@ -151,4 +151,53 @@ macro_rules! gen_vector_wrapper_impl {
...
+                let mut robj = Robj::alloc_vector($sexp, values.len());
+                let dest: &mut [$underlying_type] = robj.as_typed_slice_mut().unwrap();
+
+                for (d, v) in dest.iter_mut().zip(values) {
+                    $assignment(d, *v)
+                }
+
+                $type { robj }
+            }
+        }
```
JosiahParry, line 198: elegant 👏🏼

JosiahParry (approved):

extendr-api/src/wrapper/macros.rs
Ilia-Kosenkov: I stole & copy-pasted existing solution so no credit goes to me 🤷

JosiahParry:
Failures are expected on r-devel until https://bugs.r-project.org/show_bug.cgi?id=18806 is addressed or we suppress the `Downloading {crate name}` message

PR #881, opened 2025-02-20
Streamlining `#[extendr]` macro for impl blocks and structs

jtlandis:
Hello Devs, This PR aims to address #538 and walks back some of the logic placed in #878.

The general idea is the same, to allow multiple `impl <T>` blocks take the `#[extendr]` macro.

This PR dose the following:

* Removed `extendr` proc macro logic for creating conversions to an external pointer from the `Item::Impl` branch to its on branch of `Item::Struct`. This forces users to annotate structs with `#[extendr]` if they wish to have their type created as an external pointer.
  With this change, we no longer need to annotate impl blocks with `[extendr(impl_only = true)]` as in #878 .
* multiple impl blocks can be annotated so long as they are contained in their own modules.
* To ensure all impl blocks are properly wrapped in R, this PR adapts `extendr-api/src/metadata.rs` to handle impls contained in the metadata based on the unique names of `impl.name`. 

In comparison to #878 , here is how the same code would look.

```rust

use extendr_api::prelude::*;

//now required if impl block function 
// takes Self or returns Self
#[extendr] 
struct Counter {
    n: i32,
}

mod create {
    use super::*;

    #[extendr]
    impl Counter {
        fn new() -> Self {
            Self { n: 0 }
        }
    }

    extendr_module! {
      mod create;
      impl Counter;
    }
}

mod increment {
    use super::*;

    #[extendr]
    impl Counter {
        fn get(&self) -> i32 {
            self.n
        }

        fn increment(&mut self) {
            self.n += 1
        }
    }

    extendr_module! {
      mod increment;
      impl Counter;
    }
}

extendr_module! {
  mod counter;
  use create;
  use increment;
}
```

There are still some potential improvements. This change will technically not make it possible to create conversions for enum structs whereas before, this was simply handled by annotating the `impl <T>` for a type `T`. Shall I go ahead and include the functionality for `enum` as well? would there be any other types we should include?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 - Conversion of a `Robj` that contains a `list()`/`List` to a compatible tuple `(T0, ..., Tn)`, where `n` is atmost 12 [[#857]](https://github.com/extendr/extendr/pull/857)
 - Added conversions of `[T;N]` where `T` is `Rint`, `Rfloat`, `Rbool`, `Rcplx`, `u8`,
   `i32`, and `f64`. [[#856]](https://github.com/extendr/extendr/pull/856)
+- `FromIterator<T>` and `FromIterator<&T>` for vector wrapper where `T` represents a matching underlying type (`bool`,
+  `i32`, `f64`, `c64`) [[#879]](https://github.com/extendr/extendr/pull/879)
 
 ### Changed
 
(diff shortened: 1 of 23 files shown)


CGMossa:
Yes, please add compatibility to `enum` as well. We don't want to lose functionality from the functionality available before.
Also, this touches code that I have recently also dug into. Are all changes present necessary?

PR #882, opened 2025-02-20
Extendr proc refactor - multiple `impl` blocks and struct converions

jtlandis:
I decided to clean up my PR branches as my code had lingering changes from #878 within the `extendr_module` that did not relate to the core of the PR.

This PR addresses #538 and is a cleanup of #878 and #881 .

This PR also includes enum support, and tests have been added to `extendrtests`.

To be explicit, this PR does the following:

```rust

use extendr_api::prelude::*;

//now required if impl block function 
// takes Self or returns Self
#[extendr] 
struct Counter {
    n: i32,
}

mod create {
    use super::*;

    #[extendr]
    impl Counter {
        fn new() -> Self {
            Self { n: 0 }
        }
    }

    extendr_module! {
      mod create;
      impl Counter;
    }
}

mod increment {
    use super::*;

    #[extendr]
    impl Counter {
        fn get(&self) -> i32 {
            self.n
        }

        fn increment(&mut self) {
            self.n += 1
        }
    }

    extendr_module! {
      mod increment;
      impl Counter;
    }
}

extendr_module! {
  mod counter;
  use create;
  use increment;
}
```


JosiahParry:
Note that this PR will require that we update the user guide here: https://github.com/extendr/extendr.github.io/blob/main/user-guide/type-mapping/extendr-macro.qmd

JosiahParry:
Testing this with [tomledit](https://github.com/JosiahParry/tomledit) and it immediately throws an error 
image
I'm not familiar with proc-macros, is there any way we can make the compiler suggest adding `#[extendr]` to the struct definition? 

Adding it to the struct definition allows the package to be compiled and runs as intended.

One thing I'll also note is that this PR also allows annotating a struct with `#[extendr]` _without_ an `impl` block allows you to return the type—which is something I've been wanting! 

For example the following works _without_ using an ExternalPtr on the struct type or having an empty impl block. I'm all for that. 

```rs
#[extendr]
#[derive(Clone)]
pub struct Toml(pub DocumentMut);

#[extendr]
fn test_out() -> Toml {
    Toml::new()
}


extendr_module! {
    mod tomledit;
    fn test_out;
}
```

Moving impls out and into their own (private) modules also works. 

This is looking really good to me

jtlandis:
After spending some time researching (and begging chat-gpt for guidance) it seems that there may be a solution with [`#[rustc_on_unimplemented]`](https://doc.rust-lang.org/error_codes/E0230.html).

I dont really know what that would look like to implement in extendr though.

JosiahParry:
Super nice find! I don't think its required to fix that here. `ToVectorValue` is a bit of a pain-point that can be handled later. 

It is also worth noting in the docs that if using `#[extendr]` on a struct / enum then custom `TryFrom<>` or `From<>` implementations **cannot** be used.

@CGMossa do you have any thoughts on this PR?

JosiahParry (approved):
I'm giving this my stamp of approval. I would like one more before we merge this. It is a breaking change (with a very simple resolution). 

But on the whole I cannot see any reason why this _shouldn't_ be merged. There have been folks who have been vocal that this is their largest hold up on using extendr. This resolves that issue.

CGMossa:
I would say merge it. We should just make an issue, that dictate that I review these changes, and amend them myself, before we cut a crates.io release.

That seems very reasonable. And I've had a cursory read of the code.... It is better than several of my prototypes.. But proc macro code is impossible to review, without getting your hands dirty...

Approved!

jtlandis:
Before we do, should I add stuff to the docs and the change log?

JosiahParry:
Sounds like a plan. 

@jtlandis could you update the changelog? You can use [skip ci] in the message. 

- make an issue in extendr.github.io 
- Then can you also make a review issue for mossa in this repo?
- 
that should cover it

jtlandis:
@JosiahParry I have a few questions:

1. Should I make a new heading for "Breaking Changes" within the Changelog, or just put this under the Added heading?
2. I am unsure where the docs/repo is for extendr.github.io, are the docs built from [rextendr](https://github.com/extendr/rextendr)?

JosiahParry:
1. You got it! 
2. Here: https://github.com/extendr/extendr.github.io/issues/ !

PR #883, opened 2025-02-25
chore: avoid r-devel check

JosiahParry:
This PR :

- makes `cargo build` quiet in extendrtests so that CI for r-devel can finally pass. 
- adds `tools/msrv.R` to report rustc and cargo versions for r-devel
- addresses 0x1 error in 0 length vectors in r-devel

### DIFF
--- a/tests/extendrtests/.gitignore
+++ b/tests/extendrtests/.gitignore
@@ -41,5 +41,3 @@ vignettes/*.pdf
 # check dir
 check/
 src/rust/vendor
-src/Makevars
-src/Makevars.win
--- a/tests/extendrtests/.gitignore
+++ b/tests/extendrtests/.gitignore
@@ -40,3 +40,6 @@ vignettes/*.pdf
 
 # check dir
 check/
+src/rust/vendor
+src/Makevars
+src/Makevars.win
--- a/tests/extendrtests/src/extendrtests-win.def
+++ b/tests/extendrtests/src/extendrtests-win.def
@@ -1,0 +1,2 @@
+EXPORTS
+R_init_extendrtests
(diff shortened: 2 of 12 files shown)


CGMossa:
I must say, while I think these changes are good, they also reflect a potential adjustments needed on rextendr
The templates there doesn't have quiet as an arg for the build script. Or maybe they do.

It's never enough, but I think this will do. I will approve soon

JosiahParry:
The `--quiet` flag is used in lieu of using vendored dependencies here. The CRAN check for vendoring for Rust crates literally is a `grepl("Downloading crate", ...)` or something like that.

tests/extendrtests/.gitignore
```diff
@@ -40,3 +40,6 @@ vignettes/*.pdf
 
 # check dir
 check/
+src/rust/vendor
+src/Makevars
+src/Makevars.win
```
JosiahParry, line 45: this needs to be removed.

tests/extendrtests/tools/msrv.R
```diff
@@ -0,0 +1,116 @@
+# read the DESCRIPTION file
+desc <- read.dcf("DESCRIPTION")
+
```
JosiahParry, line 3: this comes from rextendr

tests/extendrtests/src/extendrtests-win.def
```diff
@@ -0,0 +1,2 @@
+EXPORTS
+R_init_extendrtests
```
JosiahParry, line 2: this comes from `rextendr::use_extendr()`

tests/extendrtests/.gitignore
JosiahParry, line 45:
```suggestion
```

CGMossa (approved):
LGTM

Issue #848, opened 2024-09-11
 Request for ndarray 0.16.1 Support


scm-mark:
I’m currently working on a project that uses the extendr-api crate alongside other dependencies that require ndarray v0.16.1. However, I’ve run into compatibility issues because it seems extendr-api v0.7.1 still depends on ndarray v0.15.6. This results in type mismatches when working with ndarray arrays across different crates.

scm-dan:
I am also having the same issue. @CGMossa

CGMossa:
Hello @scm-dan and @scm-mark.
Unfortunately, extendr-api supports ndarray `0.15.3`, see https://github.com/extendr/extendr/blob/39ac2e8ce8fda7c86df5bd69be37190f49e9cf48/extendr-api/Cargo.toml#L20

Now, the issue you're probably facing, is that you're importing `ndarray` and `extendr-api` separately in your crate. This is the core problem. You ought to use the `ndarray` that is used in `extendr-api`.
The `extendr_api::prelude` exports `ndarray` items. So if you remove your own imported `ndarray`, and simply `use extendr_api::prelude::*`, you'll have the `ndarray` component you'd like.

The part were we re-export `ndarray` is here:https://github.com/extendr/extendr/blob/39ac2e8ce8fda7c86df5bd69be37190f49e9cf48/extendr-api/src/prelude.rs#L74-L75

Unfortunately, this is a limitation **by Rust, and not extendr**, and we sort of have to live with it.

We do have a PR to make this a little more explicit https://github.com/extendr/extendr/pull/832, but the results are the same in the end. If you like the proposed change, please comment on it, as right now, it is not getting merged, because we don't have a feeling of what users really want.

PR #885, opened 2025-02-27
RFC: Adds mutually exclusive support for ndarray 0.16 and 0.15

andyquinterom:
I have a request for comments. I was in need of using `ndarray` 0.16 for interoperate with some other crates. I see this is also something other users have encountered (#848).

This is a solution that could work. Keeps support for versión `0.15` of `ndarray` while adding support for `0.16`.

- To use 0.16 you would enable feature `ndarray_0_16`.
- To use 0.15 you would enable feature `ndarray_0_15`.

## Drawbacks

This adds a lot more code to maintain. Perhaps this is a valid temporary solution in case we want to update to 0.16 but maintain support for a previous version for a set period of time.

### DIFF
--- a/extendr-api/Cargo.toml
+++ b/extendr-api/Cargo.toml
@@ -17,7 +17,8 @@ once_cell = "1"
 paste = "1.0.5"
 either = { version = "1.8.1", optional = true }
 libc = { version = "0.2", optional = true }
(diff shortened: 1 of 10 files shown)


CGMossa:
I've decided to simply upgrade `ndarray` to `0.16`. Please see #907.

PR #898, opened 2025-03-24
Replace `dim` vector with constant generic parameter `const DIM: usize`

wenjie1991:
Initiated by @CGMossa, PR focuses on refactoring the `RArray` struct and related traits and implementations to use const generics for dimensions instead of `dim` field. This change simplifies the code by replacing dimension arrays with const generic parameters.

Key changes include:

### Refactoring to use const generics:

* `extendr-api/src/robj/into_robj.rs`: Modified `collect_rarray` method to use const generics for dimensions.
* `extendr-api/src/wrapper/matrix.rs`: Updated `RArray` struct and its methods to use const generics instead of dimension arrays. This includes changes to `RArray` struct definition, `from_parts`, `new_array`, and various indexing methods.

### Update dimension handling:

* `extendr-api/src/wrapper/matrix.rs`: Added `get_dim` method to retrieve dimensions as a vector and updated methods to use this new method for dimension-related operations, to replace the previous `dim` field.

### Updating trait implementations:

* `extendr-api/src/wrapper/matrix.rs`: Updated `Offset`, `Index`, `IndexMut`, `Deref`, `DerefMut`, and `From` trait implementations to work with const generics.


CGMossa:
Hello @wenjie1991. Thank you for this input. I've taken the liberty to make direct changes to this PR. I would like your opinion on these changes. First, there was something with rebasing this PR to `master` that caused issues. I think you created this PR together with commits from another PR.

I've also changed the `DIM` to `NDIM`, as that is a more clear description of what the `const` generic parameter is representing. I have also changed the method `RArray::dim` to `RArray::ndim`. If users want access to the dimensions, they can use the already implemented `RArray::get_dim` (which you also use in your PR).

extendr-api/src/wrapper/matrix.rs
```diff
@@ -159,53 +186,35 @@ trait Offset<D> {
...
-            panic!("RMatrix3D index: submatrix overflow");
-        }
-        (index[0] - BASE) + self.dim[0] * (index[1] - BASE + self.dim[1] * (index[2] - BASE))
+        index
+            .iter()
+            .zip(dims.iter())
+            .rev()
+            .enumerate()
+            .fold(0, |acc, (i, (&idx, &dim))| {
+                println!("idx: {}, dim: {}", idx, dim);
```
JosiahParry, line 202:
```suggestion
```

extendr-api/src/wrapper/matrix.rs
```diff
@@ -159,53 +186,35 @@ trait Offset<D> {
...
+                if idx - BASE >= dim {
+                    panic!("array index: dimension {} overflow (0-based dimension)", i);
```
JosiahParry, line 204: same as above. we should err for no panics and instead use `throw_r_error()` here

extendr-api/src/wrapper/matrix.rs
```diff
@@ -159,53 +186,35 @@ trait Offset<D> {
...
     }
 }
 
-impl<T, D> RArray<T, D>
+impl<T, const NDIM: usize> RArray<T, NDIM>
 where
     Robj: for<'a> AsTypedSlice<'a, T>,
 {
-    pub fn from_parts(robj: Robj, dim: D) -> Self {
+    pub fn from_parts(robj: Robj, _ndim: [usize; NDIM]) -> Self {
```
JosiahParry, line 215: is the `_ndim` argument required here? The argument name has changed significantly. I think it would be better for consistency for this to be still `dim`.

extendr-api/src/wrapper/matrix.rs
```diff
@@ -220,9 +229,9 @@ where
         self.as_typed_slice_mut().unwrap()
     }
 
-    /// Get the dimensions for this array.
-    pub fn dim(&self) -> &D {
-        &self.dim
```
JosiahParry, line 225: This is a breaking change. We should keep this method.

JosiahParry:
There is a significant amount of breaking changes in this PR. We should update the changelog to reflect all of the breaking changes. 

Additionally, we have `panic!()` being used in here. We should likely offer an `offset() -> Result<usize>` and `offset_unchecked() -> usize` to opt into the behavior we have now. 

We should also not change the method and argument names if possible. Or, if there is a reason to do so we should add deprecation flags instead.

extendr-api/src/wrapper/matrix.rs
CGMossa: We have a `get_dim`. I don't know why we had a `dim` before, and who used it. But alternatively, we inject a call to `get_dim` in here.

CGMossa: I can't even mimic the previous behavior, as we no longer have with a const array anymore, that is defined by Rust and used by Rust. Rather, we are not using R's `dim` array.

extendr-api/src/wrapper/matrix.rs
CGMossa: I've changed this.

extendr-api/src/wrapper/matrix.rs
```diff
@@ -159,53 +186,35 @@ trait Offset<D> {
...
-        if index[0] - BASE > self.dim[0] {
-            panic!("matrix index: row overflow");
-        }
-        if index[1] - BASE > self.dim[1] {
-            panic!("matrix index: column overflow");
+    fn offset(&self, index: [usize; NDIM]) -> usize {
+        let dims = self.get_dim();
+        if index.len() != dims.len() {
+            panic!("array index: dimension mismatch");
         }
```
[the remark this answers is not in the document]
CGMossa: I think this need to be a follow-up. This is exactly as the previous design.

CGMossa:
Okay. I'll leave the remainder of these comments to @wenjie1991. Let us know what you think.

CGMossa:
@wenjie1991 Can you try and address these things soon? We are planning to cut a new extendr release, in order to fix the packages, that are on CRAN.

extendr-api/src/wrapper/matrix.rs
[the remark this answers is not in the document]
wenjie1991: Hi @JosiahParry why we need to mark "unchecked" in the name?

extendr-api/src/wrapper/matrix.rs
wenjie1991: Thanks, I'll update this

extendr-api/src/wrapper/matrix.rs
wenjie1991: Hi @CGMossa,
Since `dim()` is a public function, I think we should keep the original name and output.
So, let's have `dim()` to return the n-dimension. I will revert the change from `ndim()` back to `dim().
At least in the current version.

CGMossa:
No. You need to use get_dim to preserve the previous function and output.

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

JosiahParry:
This is fine with me for now! We should definitely create _unchecked() functions to identify when we are not validating bounds and a panic can occur. 

Ideally we’d have an offset() -> Result<T> where we do appropriate checks and an offset_unchecked() -> T where we dont

wenjie1991:
> This is fine with me for now! We should definitely create _unchecked() functions to identify when we are not validating bounds and a panic can occur.
> 
> Ideally we’d have an offset() -> Result where we do appropriate checks and an offset_unchecked() -> T where we dont

I agree; returning a Result aligns with Rust's style. 
We can implement it once we determine the proper use of the Result.

extendr-api/src/wrapper/matrix.rs
```diff
@@ -220,9 +229,15 @@ where
...
-    /// Get the dimensions for this array.
-    pub fn dim(&self) -> &D {
-        &self.dim
+    /// Returns the number of dimensions.
+    pub fn ndim(&self) -> usize {
+        NDIM
+    }
+
+    /// Returns the dimensions of the array.
+    // TODO:: Need to deprecate this and use `get_dim` instead.
```
CGMossa, line 238: Personally, I'm fine with two methods to get the same piece of data. I don't think it needs deprecation. Other areas have similar inconsistencies...

CGMossa:
> This requires changelog entries. Can we update the changelog and then merge?

There is a changelog entry. Were you thinking of more of a notice?

JosiahParry:
Oops! But yes, this is a breaking change. This shouldn't go under "Added." 

We should describe what the breaking change is and how to fix it.

wenjie1991:
The change log has been updated, highlighting the "breaking changes."

JosiahParry (approved):

PR #911, opened 2025-04-09
Clear warnings of the `faer` feature

CGMossa:
While exploring something about the `faer` feature, I saw these warnings, and just fixed them (applied suggestions).


JosiahParry:
Much prefer getting rid of an `expect()` or `unwrap()` or `panic!()` wherever possible. Thank you.

JosiahParry (approved):

Issue #422, opened 2022-06-26
Export method documentation


FelixHenninger:
Dear `extendr` authors,

thanks a lot for all that you do! I'm a happy user having a blast learning to wrap Rust libraries, and `extendr` is an absolute joy to work with. I'm writing today because of a minor quibble I noticed, namely that documentation for `impl` methods is lost somehow.

For example, in the test library:

* `MySubmoduleClass` provides top- and method-level [documentation in the code](https://github.com/extendr/extendr/blob/master/tests/extendrtests/src/rust/src/submodule.rs#L16-L45=)
* The top-level documentation, but not that for the methods, is [output to `extendr-wrappers.R`](https://github.com/extendr/extendr/blob/master/tests/extendrtests/R/extendr-wrappers.R#L127-L142=).
* As a result (I think) the documentation is also not present in the [generated `.Rd` files](https://github.com/extendr/extendr/blob/master/tests/extendrtests/man/MySubmoduleClass.Rd#L4-L23=)

I haven't been able to figure out exactly how the conversion works; whether the documentation is generated through `extendr`, for example, or through `roxygen`. If this is a local issue that you think can be fixed by a beginner, I'd love to investigate further and possibly contribute a fix, and would appreciate any pointers you could give me.

Many kind regards, and thanks again for all of your efforts!

-Felix

FelixHenninger:
I've been looking into this some more, collecting a few notes in hopes that they come in useful:

* In `extendr-api/src/metadata.rs`, it appears that the function `write_function_wrapper` [writes documentation](https://github.com/extendr/extendr/blob/master/extendr-api/src/metadata.rs#L151=) for a function, but `write_method_wrapper` [skips that step](https://github.com/extendr/extendr/blob/master/extendr-api/src/metadata.rs#L214-L217=).
* Both functions appear to operate on the [`Func` struct](https://github.com/extendr/extendr/blob/master/extendr-api/src/metadata.rs#L17-L28=), which _should_ include the docstring as best I can tell. I haven't figured out yet how this is generated from the Rust code -- my best guess is the [black magic in `extendr_module.rs`](https://github.com/extendr/extendr/blob/d13d215804a6b704150e45fd856238bf09395d0c/extendr-macros/src/extendr_module.rs#L46-L52=), but I'm not sure.

I think I'm going to try my hand at debugging this, but any hints are still super-welcome (and are bound to save a newb hours of head-desking).

JosiahParry:
[...]

PR #913, opened 2025-04-11
feat: documenting methods from impl blocks

albersonmiranda:
Currently, only documentation above impl blocks (impl-level) are sent to R. that happens because the #[extendr] macro operates on the Rust item it is applied to. If the macro is applied to an impl block, it processes the attributes and documentation of the impl block itself (via syn::ItemImpl), not the individual methods (fn) within the block.

Documentation comments placed above individual methods inside the impl block are not part of the impl block's attributes. Instead, they are part of the attributes of the methods themselves. Since the macro does not explicitly process the attributes of individual methods, their documentation is not extracted or passed to R.

For struct-level, there is `syn::ItemStruct` that might be helpful.

This PR aims to get method-level documentation and pass them to R.

- [x] Title (see https://github.com/albersonmiranda/fiorefactor for a proof of concept)
- [ ] Args
- [ ] Examples?

### DIFF
--- a/extendr-macros/src/extendr_impl.rs
+++ b/extendr-macros/src/extendr_impl.rs
@@ -137,8 +137,38 @@ pub(crate) fn extendr_impl(
     let self_ty = item_impl.self_ty.as_ref();
     let self_ty_name = wrappers::type_name(self_ty);
     let prefix = format!("{}__", self_ty_name);
-    let mut method_meta_names = Vec::new();
+
+    // Get the outer (impl-level) documentation.
+    // That's not ideal. We should be able to get the struct level docstring.
+    // But for now, we just use the impl docstring.
+    // So, for usage, document struct in one of the impl blocks.
     let doc_string = wrappers::get_doc_string(&item_impl.attrs);
+    
+    // Now get the method-level (fns) docstrings.
+    let mut method_docs = Vec::new();
+    for impl_item in &item_impl.items {
+        if let syn::ImplItem::Fn(method) = impl_item {
+            let mdoc = wrappers::get_doc_string(&method.attrs);
+            if !mdoc.is_empty() {
+                method_docs.push((method.sig.ident.to_string(), mdoc));
+            }
+        }
+    }
+    // Build a Methods section
+    // It actually creates a method section for each impl block, but Roxygen
+    // won't complain about that.
+    let methods_section = if !method_docs.is_empty() {
+        let mut section = String::new();
+        section.push_str("\n @section Methods:");
+        for (name, doc) in method_docs {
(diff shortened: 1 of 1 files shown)


JosiahParry:
When installing fiorefactor there are a number of warnings like `Warning: bad markup (extra space?) at Test.Rd:55:2` I wonder if that is something we can address?

albersonmiranda:
@JosiahParry I suspect we need to enclose subsection content in curly braces.

Currently:
```
#' \subsection{details}
#' Something interesting.
```

After:
```
#' \subsection{details}{
#' Something interesting.
#' }
```

Should be an easy fix

albersonmiranda:
@CGMossa What was missing is documenting the extendrtests R package. Now all tests are green!

albersonmiranda:
@CGMossa @JosiahParry Is there something else I should address here? I've setup an [example](https://webr.sh/#code=eJyFUsFq3DAQPfTmn6iaDawNu7Y3lB5Ml5xyyKEUQnOtmbVm12IljZHkdei1X9B%2FyK1%2F0Hv%2FqyO7JS5hE4OQmPfm6c2zfnx%2FtGDw570NKmiUm%2FzusYPQ%2FipaMlgMuKt7j66Y4xIC%2FH7zVlkfQOu8g%2BYIB%2FRpIsRyr8jhHppAbrnigsOOvNiKJqKMtyF0vioK0Dt0nqxRDqyE3K17q05cwlziaWydsaMKUwbwJid3KJYMZ0mWJIvFQrSoO9GxA6GJjl4ciOQ7wUhyfUtm4khCL1QQA7mjvx7BZCE8Bl4oEz7k8ZBuyjKqijhiYiA49VBLZcTHtbgqyzJRNqAzKBUErANb9zyqIusjY%2BLHQcHnkvqdxtSD6XjbVKxcrsST5NerVQxHQ4Mcz5e7%2B5ssi1NbRwMXnohjsSH9X5FnD8Tp150j2Y8WXnHwvhy%2F6sO0z628aGRz7n5OqXEIQdmDAGFxELe268P6cx94E59Iok4UjdnNXkVV8U%2B5ZHp0GZ8e617cPIwmp6aLeOP5oLfnsdj4LJbts9Jf82TYZ3QfsGmtakBzDfd71Si0wUfvlxOJb%2FlHqeeUdBQ6Kd%2BDVt9YauzpHBtMsz9H3xw2) based on [this test package](https://github.com/albersonmiranda/fiorefactor/tree/main/src/rust/src) and it seems good!

JosiahParry (approved):
I think this is in a place that is good enough for now and can be refined in the future. It closes a very long standing issue https://github.com/extendr/extendr/issues/422

PR #922, opened 2025-05-02
Adding ser/de to `Complexes` and `Strings`

### LINKED ISSUE #876: serialize extendr Strings
Serializing to `Strings` with `serde` doesn't seem to be supported atm:
```rust
use extendr_api::prelude::*;
use serde::{Deserialize, Serialize};

// this works
#[derive(Serialize)]
MyStruct {
  x: Vec<Rstr>
}

// this fails
#[derive(Serialize)]
MyStruct {
  x: Strings
}
```
Rust-analyzer says this:
```rust
the trait bound `extendr_api::Strings: Serialize` is not satisfied
for local types consider adding `#[derive(serde::Serialize)]` to your `extendr_api::Strings` type
for types from other crates check whether the crate offers a `serde` feature flag
```
And, for reference, this is in my TOML:
```toml
[dependencies]
extendr-api = { version = "*", features = ["serde"] }
```

CGMossa:
Fixes #876

While #876 seemed like a good-first-issue, I believe it requires too much for newcomers. Here's my take on it.

- `Rcomplex` is the direct, independent of enabled features, way to serialize `Complexes`.

- [ ] Missing `Strings` serialization

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,3 +29,4 @@ repository = "https://github.com/extendr/extendr"
 # When updating extendr's version, this version also needs to be updated
 extendr-macros = { path = "./extendr-macros", version = "0.8.0" }
 extendr-ffi = { path = "./extendr-ffi", version = "0.8.0" }
+serde = { version = "1.0.219", features = ["derive"] }
--- a/extendr-api/Cargo.toml
+++ b/extendr-api/Cargo.toml
@@ -19,7 +19,7 @@ either = { version = "1.8.1", optional = true }
 libc = { version = "0.2", optional = true }
 ndarray = { version = "0.16.1", optional = true }
 num-complex = { version = "0.4.6", optional = true }
-serde = { version = "1.0.219", features = ["derive"], optional = true }
+serde = { workspace = true, optional = true }
 faer = { version = "0.20", optional = true }
 
 [dev-dependencies]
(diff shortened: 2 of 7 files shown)


extendr-api/src/serializer.rs
```diff
@@ -673,6 +674,17 @@ impl ser::Serialize for Doubles {
     }
 }
 
+impl serde::Serialize for Strings {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        use crate::AsTypedSlice;
+        let strings_slice: &[Rstr] = self.robj.as_typed_slice().unwrap();
```
JosiahParry, line 683: given that we're returning a `Result<>` can we `.map_err()` into `S::Error` and then `?` and same here for `Complexes`

CGMossa: The unwrap never fails, as we know it must have this underlying type from the invocation. So I don't see a need to map to any specific error.

CGMossa: But let me look closelier....

JosiahParry:
@kbvernon this looks great! I have one request—can we replace `.ok_or_else()` with `.map_err()`? `.ok_or_else()` is eagerly evaluated meaning it _always_ runs. It will incur a runtime cost whereas `.map_err()` will let us modify the error if there is one. Some I think something like: 

```rust
self
  .robj
  .as_typed_slice()
  .map_err(|_| <S::Error as ser::Error>::custom("Failed to get slice of Rstr."))?
  .serialize(serializer)
```
would be the perfect solution

kbvernon:
I tried that first, but `.map_err()` works with `Result` and `as_typed_slice()` returns an `Option`.

JosiahParry:
eeek i was wrong. this IS good.

PR #923, opened 2025-05-02
Bump MSRV to 1.81

### LINKED ISSUE #921: Incorrect MSRV
Currently, we use `cargo-msrv` to verify that `extendr-api` is within a certain MSRV.
However, we do not check, if `extendr-api` with _features_ follow this MSRV.

First, we must rectify this situation by:
- Pushing a patch release?

The CI has to be amended with something that checks the MSRV for `extendr-api`, and the features separately,
as it can be very daunting to figure out what has broken MSRV requirement. Inspiration is:
```shell
cargo-msrv --path extendr-api verify -- cargo check --features full-functionality
```
- Add arguments to `xtask`, that calls `cargo-msrv` with the given cargo features
- Add those extra checks to CI

CGMossa:
Fixes #921 

- Bump `extendr-api`'s MSRV to `1.81`
- Now, the embedded `faer` version does satisfy MSRV
- Added `cargo extendr msrv --features` argument, to ensure that we routinely check the MSRV of extendr-api and its optionals.

Together with this PR, a slew of CI improvements made it through.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,7 +15,6 @@ jobs:
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: nightly
-      - uses: r-lib/actions/setup-r@v2
       - name: Docs
         run: cargo doc --workspace --no-deps --document-private-items --features full-functionality
         env:
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -29,9 +29,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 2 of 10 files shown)


JosiahParry:
Okay, so here's my thinking on this: 

The MSRV for `faer` is `>= 1.81`. If one wants to use the faer feature they **inherit** the MSRV of faer. The faer feature is not _core_ extendr functionality and I don't think it should dictate the MSRV of the entire project. 

I think we keep the MSRV as is but use the new xtask CLI feature introduced in this PR to test the MSRV when features are used. 

I think, though, that we are safe to bump MSRV to `1.71` and use the unwind feature you'd like there. That, I think, is a very justified bump of the MSRV.

PR #927, opened 2025-05-20
Remove build-print dependency

JosiahParry:
This PR removes build-print dev dependency for extendr-ffi. 

As build-print is MIT licensed, I've moved the macro definitions into the crate. This is because they set their rust edition to 2024.


CGMossa (approved):

CGMossa:
This works now. I think it needed a rebase too. R-devel is failing, but I vote that we ignore this for now. We don't know if R-devel made a mistake in their current version, or if we are at fault..

PR #942, opened 2025-07-12
Patch release 0.8.1

CGMossa:
These are the changes made in the patch release. The changes here are less than what's on `master`.


.github/workflows/trigger_rextendr_ci.yaml
```diff
@@ -0,0 +1,19 @@
+name: Trigger rextendr CI
+on:
```
Ilia-Kosenkov, line 2: A nitpick, but I think something is wrong with the indentation here.

JosiahParry:
@CGMossa think we can close here?

PR #946, opened 2025-09-05
feat: add macro attribute `invisibly`

albersonmiranda:
This PR modifies wrapper generation to add invisible() for Result<()> type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Added
 
+- Added `#[extendr(invisible)]` attribute to allow functions to return invisibly in R [[#946]](https://github.com/extendr/extendr/pull/946)
+
 ### Changed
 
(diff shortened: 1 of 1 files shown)


albersonmiranda:
@JosiahParry You're right, I was only thinking of my use case `Result<()>`. I'll scratch that out. Now, about NULL `()`, if I remove it it would introduce a breaking change. Is that OK?

CGMossa:
I like the attribute macro. I don't think it needs to be a `= true` though. It ought to be either `#[extendr(invisible)]` or when it is not there, it should be enough. 

I think `()` resulting in `invisible()` is a good default, and `Result<()>` doing the same make a lot of sense. The issue is that _being explicit is good in Rust_, but it is not the norm in R. I vote retain the current design choice, even if it goes against our modern perspective..

PR #955, opened 2025-10-23
Added `Nullable<()>` to conversions

CGMossa:
Added `Nullable<()>` implementations. In the case of non-`NULL` input, the conversion throws an error. Also, added `TryFrom` for `Option<()>`. 

Alongside this
- snapshots have been updated, using `cargo +1.82 extendr devtools-test -a`. The latest rustc version yields different snapshots, and thus we have to use an older version to check snapshot tests. I used one that I had installed previously, so it wouldn't invoke a new installation of an old toolchain.
- Fixed a warning in `xtask`, but this is not user-facing

### DIFF
--- a/extendr-macros/Cargo.toml
+++ b/extendr-macros/Cargo.toml
@@ -1,19 +1,19 @@
 [package]
 name = "extendr-macros"
-description = "Generate bindings from R to Rust."
 version.workspace = true
 authors.workspace = true
 edition.workspace = true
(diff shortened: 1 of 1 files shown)


JosiahParry:
See https://github.com/dtolnay/syn/issues/1924

checking if the last patch version works locally.

JosiahParry (approved):
Fixed msrv by pinning syn

PR #973, opened 2025-10-27
fix: remove tracebacks from errors 

JosiahParry:
This PR modifies the error behavior when returning an Err() to R from Rust. 

By default extendr now uses the Display method of an Error to display the message. 

If `EXTENDR_BACKTRACE=true` then the full traceback is displayed. This is great because it now gives us the ability to error chain without getting nasty tracebacks in the R console. 

image

image


JosiahParry:
The Error trait uses Display for human readable outputs. But nonetheless, this is exciting thank you! 

https://doc.rust-lang.org/std/error/trait.Error.html

PR #974, opened 2025-10-29
fix: implement ToVectorValue for Rstr

JosiahParry:
This PR implements ToVectorValue for Rstr closing https://github.com/extendr/extendr/issues/770

### DIFF
--- a/extendr-macros/tests/cases/case-extendr-macro.rs
+++ b/extendr-macros/tests/cases/case-extendr-macro.rs
@@ -28,12 +28,6 @@ impl FooStruct {
 }
 
 #[extendr]
-default impl FooStruct {}
-
-#[extendr]
-unsafe impl FooStruct {}
-
-#[extendr]
 impl<const N: usize> FooStruct {}
 
 struct FooStructWithParam<A> {
--- a/extendr-macros/tests/cases/case-extendr-macro.stderr
+++ b/extendr-macros/tests/cases/case-extendr-macro.stderr
@@ -40,52 +40,40 @@ error: found &self in non-impl function - have you missed the #[extendr] before 
 27 |     fn misplaced_macro(&self) {}
    |                        ^^^^^
 
-error: default not allowed in #[extendr] impl
+error: const params not allowed in #[extendr] impl
   --> tests/cases/case-extendr-macro.rs:31:1
    |
-31 | default impl FooStruct {}
-   | ^^^^^^^^^^^^^^^^^^^^^^^^^
-
-error: unsafe not allowed in #[extendr] impl
-  --> tests/cases/case-extendr-macro.rs:34:1
-   |
-34 | unsafe impl FooStruct {}
-   | ^^^^^^^^^^^^^^^^^^^^^^^^
-
-error: const params not allowed in #[extendr] impl
-  --> tests/cases/case-extendr-macro.rs:37:1
-   |
-37 | impl<const N: usize> FooStruct {}
+31 | impl<const N: usize> FooStruct {}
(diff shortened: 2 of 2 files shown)


CGMossa:
I have to take a look at this more thoroughly. I was sure there was something prohibiting doing this.

extendr-api/src/robj/into_robj.rs
```diff
@@ -426,6 +426,103 @@ macro_rules! impl_str_tvv {
...
+    }
+}
+
+// These traits all derive from GetSexp with default implementations
+impl Length for Rstr {}
+impl Types for Rstr {}
+impl Conversions for Rstr {}
+impl Rinternals for Rstr {}
+impl Slices for Rstr {}
+impl Operators for Rstr {}
```
JosiahParry: this frustrated me lol! Why don't we have blanket impls for these?

jtlandis:
Thank you @JosiahParry for taking the time to put this together. After looking at it it is now more clear why this wasn't done before.

I personally do not know how major this breaking change would be. I suppose it comes down to how different is a literal `CHAREXP` and a length 1 `STREXPR`...

If a user knows that `Rstr` is a `CHAREXP`, and then they call `Robj::From(rstr_val)`, I suppose it would be a bit unexpected to have the Rtype change.

Is there ever a world in which we remove the blanket implementation  `From<T: ToVectorValue> for Robj` and include that in some macro for the basic scalars?

To be transparent, I also don't know if removing that blanket implementation will also have unexpected consequences.

jtlandis:
I like the changes that you've made. Removing the `make_conversion!(Rstr, ...)` macro in favor of the manual `TryFrom<Robj> for Rstr` is imo the correct direction!

If we want to entertain a scenario in which we do not bring in any breaking changes to the From conversion api, an alternative would be to make the blanket implementation more restrictive:

```rust
/// empty marker
pub trait NonRobj {}

impl<T> From<T> for Robj
where
    T: ToVectorValue + NonRobj,
{
    fn from(scalar: T) -> Self {
        Some(scalar).into_iter().collect_robj()
    }
}

impl NonRobj for Rint {}
impl NonRobj for i32 {}
/// etc ...
```

This would allow structs to implement a way to be collected into a vector with `ToVectorValue` without having to also implement `From<T> for Robj`.  So for the case of `Rstr` we would just not implement `NonRobj` and manually implment the From impls

```rust
impl From<Rstr> for Robj {
    /// Make an robj from a wrapper.
    fn from(val: Rstr) -> Self {
        val.robj
    }
}

// We can convert a reference to any wrapper to a Robj by cloning the robj pointer
impl From<&Rstr> for Robj {
    /// Make an robj from a wrapper.
    fn from(val: &Rstr) -> Self {
        val.robj.to_owned()
    }
}
```

It seems to have worked well on [my fork](https://github.com/jtlandis/extendr/commit/b8ac2842bafce472c7d32f6d38e5e03ad2f9837b)

JosiahParry:
i don't see the marker trait on your fork. Adding the manual impls alone doesn't give us ToVectorValue for Rstr which means we'd still not be able to create character matrices unfortunately. 

But I think its okay to have breaking changes right now while the API is getting cleaned up and solidified. I think this is one of the goals of the next few months—clean it allll up and make it much easier to contribute and focus on nice to have functionality.

CGMossa:
Oh yeah, I don't want charsxp to R. If you can assure me that this is the case, I'll approve.

extendr-api/src/robj/into_robj.rs
CGMossa: I did this once.. and it was glorious, but that branch didn't make it to PRs.

PR #980, opened 2025-11-05
[Internals] Added MSRV checks of optional features 

CGMossa:
This superceeds #923

This adds msrv checks to the optional features of `extendr-api`. These tests
are done in CI but they are not a blocker for approval. extendr should have an MSRV, but the dependencies that people might choose, ought not to necessarily follow the strict MSRV policy of extendr.

Then, this bumps the extendr-api MSRV from 1.65 to 1.68. This is due to some changes that we are simply unable to discern why has happened, but `quote` now depends on it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ authors = [
   "Michael Milton <michael.r.milton@gmail.com>",
 ]
 edition = "2021"
-rust-version = "1.65"
+rust-version = "1.68"
 repository = "https://github.com/extendr/extendr"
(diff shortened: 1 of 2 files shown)


JosiahParry:
Well, I suppose this is more a philosophic discussion than anything. 

extendr's MSRV shouldn't be dictated by optional features, right? If you want to use faer, you now take on the MSRV of faer. I think the MSRV should **only** be checked for core features—results list for one. _maybe_ serde? but that puts us at the mercy of serde

PR #989, opened 2025-11-21
fix: update invisibility logic for function return types

### LINKED ISSUE #988: bug: all `-> Result` returns invisibly
An unfortunate mishap in #946 is that all `Result` types return invisibly. That's not ideal, as the default invisible behavior ought to be limited to `Result<(),_>`, `Result<()>`, and `Option<()>`.

Possible culprit are the following lines: 

https://github.com/extendr/extendr/blob/baf58405b2c2a9f969124c294a001b4fa551feec/extendr-api/src/metadata.rs#L198-L202

albersonmiranda:
This PR updates the default invisibility logic in `metadata.rs` to only apply to unit return types `()`.

Closes #988

### DIFF
--- a/tests/extendrtests/R/extendr-wrappers.R
+++ b/tests/extendrtests/R/extendr-wrappers.R
@@ -111,7 +111,7 @@ test_hm_string <- function(x) .Call(wrap__test_hm_string, x)
 
 test_hm_i32 <- function(x) .Call(wrap__test_hm_i32, x)
 
-test_try_from_hm <- function(x) invisible(.Call(wrap__test_try_from_hm, x))
+test_try_from_hm <- function(x) .Call(wrap__test_try_from_hm, x)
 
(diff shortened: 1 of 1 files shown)


CGMossa:
Could you please restrict the behavior of returning invisibly due to the use of `#[extendr(invisible)]` only?

albersonmiranda:
> Could you please restrict the behavior of returning invisibly due to the use of `#[extendr(invisible)]` only?

Yes, but then the following happens:
image

which may be undesired. The two paths are:

1. No invisible by default: the above happens and `#extendr[invisible = true]` should be set for every method that returns nothing.
2. Keep current implementation of invisible NULL by default: if a user wants to return NULL visibly, then it must add `#extendr[invisible = false]` for that function.

Thoughts? @CGMossa @JosiahParry

CGMossa:
@JosiahParry note this though:
```r
+ gx <- function() {
+   x <- 1L;
+ }
> gx()
> (gx())
[1] 1
>
```

PR #1000, opened 2025-11-26
enforce type-strictness for references

CGMossa:
This PR does several things.

It removes `TryFrom<Robj> for &/&mut T` as that appears to be unsound under the Rust memory model. We are discussing it in discord, but it is true, that the current API, relies on the "run-time" assumption, that R protects the `Robj` that is passed down to us. But those we create ourselves are not protected, and if one runs the `gc()` within an `extendr`-`fn`, an unsound state might be attained.

# Type-safe conversions

But the main benefit of this PR is the addition of type-safe conversions. Currently, if you ask for `i32` in a `extendr`-`fn` signature, what you get, is that anything _coercible_ to a `i32`, will be coerced for you. These are reasonable assumptions, but you still might want the direct, un-transformed conversion. Now, these are available to you through `&i32`.

```rs
#[extendr]
fn sum_many_args_same_type(x: &i32, y: &i32, z: &mut i32) {
    let _result = *x + *y + *z;
}
```

will not respond kindly to being passed `sum_many_args_same_type(1,1,1)`, as these are `double`/`numeric`, and not `integer`s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,7 @@
 ### Added
 
 - Implements `CanBeNa` for `u8` and `Rcomplex` [[#1000]](https://github.com/extendr/extendr/pull/1000)
-- Added type-strict conversions through `&/&mut T`, where `T` is a R native data type namely
-  `i32`, `f64`, `Rint`, `Rfloat`, `Rcplx`, `Rbool`, and `u8`. [[#1000]](https://github.com/extendr/extendr/pull/1000)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Added
 
+- Added type-strict conversions through `&/&mut T`, where `T` is a R native data type namely
```
JosiahParry:
```suggestion
- Implements `CanBeNa` for `u8` and `Rcomplex` [[#1000]](https://github.com/extendr/extendr/pull/1000)
- Added type-strict conversions through `&/&mut T`, where `T` is a R native data type namely
```

JosiahParry (changes requested):

CGMossa:
This PR adds the ability to ask for an R native type without having it coerced by extendr-api before receiving it. The typed references has type information, and that can be of benefit. Plus, it gives a path for the _strict_ passthrough of R types.

Issue #962, opened 2025-10-25
refactor: rename TryFromRobj -> TryFromList


JosiahParry:
The TryFromRobj works strictly with a list. We should rename the trait accordingly and add a deprecation for TryFromRobj

PR #1001, opened 2025-11-26
Add argument name to error messages!

CGMossa:
Given a function with many, of the same type non-opaque type:
```rs
#[extendr]
fn sum_many_args_same_type(x: &i32, y: &i32, z: &i32) {
    let _result = *x + *y + *z;
}
```
if this function is passed invalid arguments, it is not possible to see which arguments failed.

```r
> sum_many_args_same_type(1,1,1)
Error in `sum_many_args_same_type()`:
! Expected Integers got Doubles
> sum_many_args_same_type(1L,1,1)
Error in `sum_many_args_same_type()`:
! Expected Integers got Doubles
> sum_many_args_same_type(1L,1L,1)
Error in `sum_many_args_same_type()`:
! Expected Integers got Doubles
> sum_many_args_same_type(1L,1L,1L)
NULL
```

This PR ensures that the error message is indeed informative,

```r
> sum_many_args_same_type(1,1,1)
Error in `sum_many_args_same_type()`:
! failed to convert argument `x` from R: Expected Integers got Doubles
> sum_many_args_same_type(1L,1,1)
Error in `sum_many_args_same_type()`:
! failed to convert argument `y` from R: Expected Integers got Doubles
> sum_many_args_same_type(1L,1L,1)
Error in `sum_many_args_same_type()`:
! failed to convert argument `z` from R: Expected Integers got Doubles
> sum_many_args_same_type(1L,1L,1L)
NULL
```

### Misc

This PR depends on the proposed PR #1000 for type-safe runtime conversions.
Also, together with this PR, there is a bunch of improvements to the workflow.. and lint solves, etc.


extendr-api/src/robj/try_from_robj.rs
```diff
@@ -102,6 +103,123 @@ impl_try_from_scalar_integer!(isize);
...
+    let re = v.re().inner();
+    let im = v.im().inner();
+    re.is_nan() || im.is_nan() || re.is_na() || im.is_na()
+});
+impl_scalar_borrow!(c64, ExpectedComplex, true, |v: &c64| {
+    let rc: Rcplx = (*v).into();
+    let re = rc.re().inner();
+    let im = rc.im().inner();
+    re.is_nan() || im.is_nan() || re.is_na() || im.is_na()
+});
```
jtlandis: should `impl_scalar_borrow!(c64, ...` be false for allow_na? It is a rust struct built from two `f64` values and `impl_scalar_borrow!` for `f64` is false.

jtlandis (commented):
@CGMossa 

I've taken a look at the PR and I have a few comments. Initially I was a bit confused on how to use the macro. I originally thought it was wrapped into `#[extendr]`, but after some digging I realized it was for `#[derive(TryFromRobj)]`.

There are two components to this PR that I have questions regarding the project's philosophy, and the first one may be out of scope for this PR.

## Interaction between both `#[extendr]` and `#[derive(TryFromRobj)]`

both of these provide some `TryFrom<&Robj> for T` implementation, except `#[extendr]` provides the `&T` and `#[derive(TryFromRobj)]` provides `T`.

```rust
#[extendr]
#[derive(TryFromRobj)]
struct Foo(f64, i32);
```

In the above case, `#[extendr]` expands to:

```rust
//...
impl TryFrom<&Robj> for &Foo {
    type Error = extendr_api::Error;
    fn try_from(robj: &Robj) -> extendr_api::Result<Self> {
        use extendr_api::ExternalPtr;
        unsafe {
            let external_ptr: &ExternalPtr<Foo> = robj.try_into()?;
            external_ptr.try_addr()
        }
    }
}
//...
```

And the `#[derive(TryFromRobj)]` portion expands to:

```rust
// ...
impl std::convert::TryFrom<&extendr_api::Robj> for Foo {
    type Error = extendr_api::Error;
    fn try_from(value: &extendr_api::Robj) -> extendr_api::Result<Self> {
        Ok(Foo {
            0: value
                .dollar(".0")?
                .try_into()
                .map_err(|error| extendr_api::error::Error::Other(
                    ::alloc::__export::must_use({
                        ::alloc::fmt::format(
                            format_args!(
                                "failed to convert tuple field `{0}` on `{1}`: {2}",
                                ".0",
                                "Foo",
                                error,
                            ),
                        )
                    }),
                ))?,
          // ....
          // more of the same for field 1
        })
    }
}
```

These two ways of getting `Foo` from R are vastly different. The first case assumes we want a reference to an already initialized `Foo` struct, whereas the second is providing a way to construct Foo provided the R object has the right types.

I just wanted to make sure this is the true intention of it.

## The usage of `Robj::dollar()` for tuple structs

Again this gets into a bit of the projects philosophy on these things, but as it is written, unnamed lists cannot be used to construct `Foo` even if the R object has the right types. Stealing tests from other parts of the project, consider this function:

```rust
#[extendr]
fn sum_many_args_same_type(x: &i32, y: Foo, z: &i32) {
    let _result = *x + y.1 * z;
}
```

within R we get the following:

```r
> sum_many_args_same_type(1L, list(1.1, 2L), 3L). # correct types, just not named
Error in sum_many_args_same_type(1L, list(1.1, 2L), 3L) : 
  failed to convert argument `y` from R: failed to convert tuple field `.0` on `Foo`: Expected non zero length
> sum_many_args_same_type(1L, list(`.0` = 1.1, `.1` = 2L), 3L) #name the list .0
NULL
> sum_many_args_same_type(1L, list(`.0` = "", `.1` = 2L), 3L)  #named properly, but improper data
Error in sum_many_args_same_type(1L, list(`.0` = "", `.1` = 2L), 3L) :
  failed to convert argument `y` from R: failed to convert tuple field `.0` on `Foo`: Expected Numeric, got Strings 
```

My initial assumption was that named structs in Rust would expect named lists in R, and that a tuple struct in rust would expect any list, but grab values by index. Again, this is probably just a matter of philosophy. Requiring R objects being converted to a tuple struct to have those names is verbose, but definitely more intentional on the user's side.

JosiahParry:
> "These two ways of getting Foo from R are vastly different. The first case assumes we want a reference to an already initialized Foo struct, whereas the second is providing a way to construct Foo provided the R object has the right types."

See https://github.com/extendr/extendr/issues/962 the trait needs to be renamed to `TryFromList`—as you nailed the distinction. The goal for `#[extendr]` as I understand it, is to create external pointers for a struct / object in rust. We require a reference here as it will always be owned by Rust and not R. 

Then the `TryFromList` is a way to **construct** the struct from an R list—thus owned data

extendr-api/src/robj/try_from_robj.rs
CGMossa: Good catch!

PR #1002, opened 2025-11-27
Let's fix multithreaded access to R API

CGMossa:
I'm finally attacking the single-threaded access to the R API. I've now removed all `-- --test-threads=1`, meaning we are running tests in parallel.

This PR adds a static state quantity that can be reset (because R's error longjump'd away from `Drop::drop`) at will. It is not invoked at the moment, but it could be an addition to the R wrappers to reset the start.

In subsequent PRs, I will introduce an unwind protect guard, which will then use this reset mechanism in `cleanfun`.

Depends on

- [x] #1016

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Added
 
+- Implements `CanBeNa` for `u8` and `Rcomplex` [[#1000]](https://github.com/extendr/extendr/pull/1000)
+- References to R native types are strictly checked `&/&mut T`, where `T` is a R native data type namely `i32`, `f64`, `Rint`, `Rfloat`, `Rcplx`, `Rbool`, and `u8`. We cannot perform coercion on borrowed data only owned data. [[#1000]](https://github.com/extendr/extendr/pull/1000)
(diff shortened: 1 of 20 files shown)


extendr-api/src/thread_safety.rs
```diff
@@ -3,58 +3,98 @@ use crate::*;
 use extendr_ffi::{
     R_MakeUnwindCont, R_UnwindProtect, Rboolean, Rf_error, Rf_protect, Rf_unprotect,
 };
-use std::cell::Cell;
-use std::sync::Mutex;
+use std::cell::RefCell;
+use std::sync::{Mutex, MutexGuard};
 
-/// A global lock, that should represent the global lock on the R-API.
-/// It is not tied to an actual instance of R.
```
JosiahParry:
```suggestion
/// A global lock, that should represent the global lock on the R-API.
/// It is not tied to an actual instance of R.
```

JosiahParry: These comments are helpful

CGMossa:
The issue is the deferred allocation that happens when we use `N:M` (which is altrep (compact)) together with a read-only assumption that R won't allocate, but reading a `seq_len`/`N:M` results in an allocation even when we adhere to read-only lock.

PR #1005, opened 2025-11-27
refactor: `extendr_engine::with_r` returns a Result

CGMossa:
We like `?`, but using `test!` is not without friction. A reasonable compromise is `with_r_result`.

Recent example:

```rs
#[test]
fn use_robj_as_key() -> Result<()> {
    with_r_result(|| {
        let env1: Robj = R!("local({ e <- new.env(parent = emptyenv()); e$a <- 1L; e })")?;
        let env2: Robj = R!("local({ e <- new.env(parent = emptyenv()); e$a <- 1L; e })")?;

        let mut set = HashSet::new();
        set.insert(env1);
        assert!(!set.insert(env2));
        Ok(())
    })
}
```

it allows for `?` without having to specify the `-> Result<_,_>` in too many places

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,10 +6,11 @@
         "extendr-macros/Cargo.toml",
         "tests/extendrtests/src/rust/Cargo.toml"
     ],
-    // enable to debug proc-macros
-    // "rust-analyzer.cargo.extraEnv": {
-    //     "RUSTUP_TOOLCHAIN": "nightly",
-    //     "RUSTFLAGS": "-Zmacro-backtrace",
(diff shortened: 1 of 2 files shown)


JosiahParry:
Actually, what if we just replaced with_r() to return a result and not have a separate variant?

JosiahParry (changes requested):
there are only 16 instances of with_r( so I think we shuold bite the bullet and make that change. I'll do so.

JosiahParry (approved):
there are only 16 instances of with_r( so I think we shuold bite the bullet and make that change. I'll do so.

PR #1009, opened 2025-11-27
Enable only symbol lookup for Rust wrappers

CGMossa:
This PR enables symbol lookup of Rust wrappers by default. One may call a C-function either by quoting the name, or using a symbol, i.e. `.Call(C_wrapper_foo)`, or `.Call("C_wrapper_foo")`. In `extendr-api` we enable the ability to emit wrappers that use either invocation, but the common standard is to use the symbol lookup, as is today. There will be less work done by `R_FindSymbol` if this is enabled, see 

https://github.com/r-devel/r-svn/blob/trunk/src/main/Rdynload.c#L1185-L1201

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 
 ### Changed
 
+- Force symbol-based lookup of native entry points in packages built with extendr (`R_forceSymbols(TRUE)`); `.Call("wrap__…", …)` strings must now be replaced with the symbol form `.Call(wrap__…, …)`. <https://github.com/extendr/extendr/pull/1009>
 - **Breaking**: bumps MSRV to 1.77 <https://github.com/extendr/extendr/pull/1075>
 - **Breaking**: deprecates and removed `global_env()`, `base_env()`, and `empty_env()` from the `prelude`  <https://github.com/extendr/extendr/pull/1075>
 - **Breaking**: non-API items `global_var()`, `local_var()`, `global!()` have been removed  <https://github.com/extendr/extendr/pull/1075>
--- a/extendr-api/src/lib.rs
+++ b/extendr-api/src/lib.rs
@@ -461,9 +461,11 @@ pub unsafe fn register_call_methods(info: *mut extendr_ffi::DllInfo, metadata: M
         std::ptr::null(),
     );
 
-    // This seems to allow both symbols and strings,
+    // Force symbol-based lookup of C entry points (.Call(symbol, ...) instead of
+    // .Call("symbol", ...)) — skips R_FindSymbol's string scan. See:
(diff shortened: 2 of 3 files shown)


tests/extendrtests/R/extendr-wrappers.R
```diff
@@ -1,10 +1,9 @@
-# Generated by extendr: Do not edit by hand
-
 # nolint start
 
+# Generated by extendr: Do not edit by hand
 #
 # This file was created with the following call:
-#   .Call("wrap__make_extendrtests_wrappers", use_symbols = TRUE, package_name = "extendrtests")
+#   .Call(wrap__make_extendrtests_wrappers, use_symbols = TRUE, package_name = "extendrtests")
```
JosiahParry: i like this

JosiahParry (changes requested):
While I really want this feature this amount of customization embedded into extendrtests is quite worrisome. It essentially an entire rewrite of the documentation and build process completely eliding rextendr.

If these are things we want we should definitely be making PRs to rextendr instead of extendr for this.

Can we revert the `document.R`, `install.libs.R`, `extendrtests-wrappers.c`, `justfile`, and `test-template.yml`?

CGMossa:
Rebased onto current `main` and addressed the review by reverting the scaffolding pieces you flagged. The branch now contains only the feature itself:

- `extendr-api/src/lib.rs` — flip `R_forceSymbols(info, FALSE)` → `TRUE`
- `tests/extendrtests/tests/testthat/test-wrappers.R` — adjust the in-tree self-test from `.Call("wrap__make_extendrtests_wrappers", …)` to the symbol form `.Call(wrap__make_extendrtests_wrappers, …)`, since with `R_forceSymbols(TRUE)` the string form is no longer accepted
- `CHANGELOG.md` entry

Reverted in full (now identical to `origin/main`):

- `tests/extendrtests/R/document.R` and `tests/extendrtests/man/document.Rd`
- `tests/extendrtests/src/install.libs.R`
- `tests/extendrtests/src/extendrtests-wrappers.c`
- `tests/extendrtests/src/Makevars`, `Makevars.win`, `.Rbuildignore`, `entrypoint.c`, `src/rust/build.rs`
- `justfile`
- `.github/workflows/test-template.yml`, `post-merge.yml`, `pr-devel-check.yml`
- `.gitignore`

Net diff vs `origin/main` is 3 files / 6 insertions / 3 deletions. `cargo test -p extendr-api --features=full-functionality` is green (446 passed) and `devtools::test()` against `extendrtests` is green (355 passed) locally.

If you want symbol-only wrapper generation as a separate change, that's a {rextendr} concern — happy to open a PR there.

PR #1011, opened 2025-11-27
[internal] Use the same directory for all targets

CGMossa:
The extendr repo consists of 4 crates (`tests/extendrtests`, `extendr-ffi`, `extendr-engine` and `extendr-api`), and these all produce their own target directories, because they are relative. This setting ensures that it all becomes on folder with rust analyzer.

### DIFF
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -4,13 +4,13 @@ on:
   workflow_dispatch:
   push:
     branches:
-      - master
+      - main
     paths:
       - "**.rs"
       - "**/Cargo.toml"
   pull_request:
     branches:
-      - master
+      - main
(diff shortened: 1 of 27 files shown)


JosiahParry:
Actually, I think this is not needed at all since all crates are within a workspace they share `./target`. 

I had `extendr-ffi/target` but i think that was just a leftover. Delete all of the miscelanous target directories.

Run `just check`. Then run 

```shell
extendr ⚡ find . -type d -name "target"
./target
./tests/extendrtests/target
./tests/extendrtests/src/rust/target
```

You'll see that they share the same /target directory

CGMossa:
Thanks Josiah. I appreciate the understanding.

PR #1014, opened 2025-11-27
Remove secret exports of `HashMap` and `CString`.

CGMossa:
These exports are annotated with `doc(hidden)`. They leave a way to refer to certain items in macros. Fortunately, we don't need these anymore, and they have been undocumented, so they pose no breaking change challenge.


JosiahParry (approved):
ew. thank you for this

PR #1017, opened 2025-11-28
Replace `as *const _` / `as *mut _` with safer variants 

### LINKED ISSUE #969: chore: use .cast() instead of as for pointers
We should remove graphics anyways, but this casting can be done more succinctly by using https://doc.rust-lang.org/stable/std/primitive.pointer.html#method.cast-1 and friends.

_Originally posted by @CGMossa in https://github.com/extendr/extendr/pull/965#discussion_r2463208754_

CGMossa:
fixes #969

Increased MSRV to 1.76. Seeing as CRAN's MSRV is 1.81, I'm taking a very cautious increase here. The reason for this increase is then we can construct pointers via `std::ptr::from_ref` and  `std::ptr::from_mut`.

Furthermore, some inconsistencies were erased here with use of dataptr, by force casting `*const _` to `*mut const`, etc.

### DIFF
--- a/extendr-api/src/graphics/mod.rs
+++ b/extendr-api/src/graphics/mod.rs
@@ -117,7 +117,7 @@ pub struct TextMetric {
 /// A row-major array of pixels. One pixel is 32-bit, whose each byte represents
 /// alpha, blue, green, and red in the order.
 #[derive(Clone, Debug, PartialEq)]
-pub struct Raster<P: AsRef<[u32]>> {
+pub struct Raster<P: AsRef<[u32]> + AsMut<[u32]>> {
     pub pixels: P,
     pub width: usize,
 }
@@ -405,7 +405,7 @@ impl Context {
     }
 
     pub(crate) fn context(&self) -> pGEcontext {
-        &self.context as *const extendr_ffi::R_GE_gcontext as *mut extendr_ffi::R_GE_gcontext
+        std::ptr::from_ref(&self.context)
     }
 
     // Affine transform.
@@ -551,7 +551,7 @@ impl Device {
     }
 
     /// Start a new page. The page color can be set in advance.
-    pub fn new_page(&self, gc: &Context) {
+    pub fn new_page(&self, gc: &mut Context) {
         unsafe { GENewPage(gc.context(), self.inner()) }
     }
 
@@ -563,7 +563,7 @@ impl Device {
     }
 
     /// Draw a stroked line. gc.color() is the stroke color.
-    pub fn line(&self, from: (f64, f64), to: (f64, f64), gc: &Context) {
+    pub fn line(&self, from: (f64, f64), to: (f64, f64), gc: &mut Context) {
         let from = gc.t(from);
         let to = gc.t(to);
         unsafe { GELine(from.0, from.1, to.0, to.1, gc.context(), self.inner()) }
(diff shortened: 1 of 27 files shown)


extendr-api/src/graphics/mod.rs
```diff
@@ -787,31 +787,31 @@ impl Device {
...
+                std::ptr::from_mut(&mut res.width),
                 self.inner(),
             );
             gc.tmetric(res)
         }
     }
 
     /// Get the width of a unicode string.
-    pub fn text_width<T: AsRef<str>>(&self, text: T, gc: &Context) -> f64 {
+    pub fn text_width<T: AsRef<str>>(&self, text: T, gc: &mut Context) -> f64 {
```
JosiahParry: note: breaking change

extendr-api/src/io/save.rs
```diff
@@ -27,7 +26,7 @@ impl<W: Write> OutStream<W> {
         hook: Option<WriteHook>,
     ) -> Box<OutStream<W>> {
         unsafe extern "C" fn outchar<W: Write>(arg1: R_outpstream_t, arg2: ::std::os::raw::c_int) {
-            let writer = &mut *((*arg1).data as *mut W);
+            let writer = (*arg1).data.cast::<W>().as_mut().unwrap();
```
JosiahParry: same as above except in this case we don't return a result.

extendr-api/src/io/save.rs
```diff
@@ -37,8 +36,8 @@ impl<W: Write> OutStream<W> {
             arg2: *mut ::std::os::raw::c_void,
             arg3: ::std::os::raw::c_int,
         ) {
-            let writer = &mut *((*arg1).data as *mut W);
-            let b = std::slice::from_raw_parts(arg2 as *mut u8, arg3 as usize);
+            let writer = (*arg1).data.cast::<W>().as_mut().unwrap();
```
JosiahParry: gotta have an epic for removing .unwrap()

extendr-engine/src/lib.rs
```diff
@@ -72,12 +72,32 @@ use std::sync::Once;
...
     };
 }
 
+#[cfg(all(target_os = "windows", target_arch = "x86"))]
+static mut R_ARGV: [*mut raw::c_char; 6] = [
+    cstr_mut!("R"),
+    cstr_mut!("--arch=i386"),
+    cstr_mut!("--slave"),
+    cstr_mut!("--no-save"),
+    cstr_mut!("--vanilla"),
```
JosiahParry: possibly breaking change here.

extendr-engine/src/lib.rs
```diff
@@ -72,12 +72,32 @@ use std::sync::Once;
...
+    std::ptr::null_mut(),
+];
+
+#[cfg(not(all(target_os = "windows", target_arch = "x86")))]
+static mut R_ARGV: [*mut raw::c_char; 5] = [
+    cstr_mut!("R"),
+    cstr_mut!("--slave"),
+    cstr_mut!("--no-save"),
+    cstr_mut!("--vanilla"),
+    std::ptr::null_mut(),
```
JosiahParry: any reason why the 5th argument is a null pointer?

extendr-macros/src/wrappers.rs
```diff
@@ -302,7 +302,9 @@ pub(crate) fn make_function_wrappers(
                 mod_name: #c_name_str,
                 args: args,
                 return_type: #return_type_string,
-                func_ptr: #wrap_name as * const u8,
+                func_ptr: unsafe { *std::ptr::from_ref(&#wrap_name) }
+                    as *const ()
+                    as *const u8,
```
JosiahParry: what is this doing? is it taking the wrap_name and then casting as a null pointer then casting to a normal pointer?

JosiahParry (changes requested):
We need to document breaking changes

PR #1028, opened 2025-12-22
chore!: remove result from the prelude

JosiahParry:
This PR removes Result from the prelude. This is a breaking change but a very welcome one i think.


JosiahParry:
The one windows test is failing due to pak being unavailable

Issue #1031, opened 2026-01-07
Feature request: Add From<anyhow::Error> implementation to error handling.


Yunuuuu:
It’s common for users to work with errors from the `anyhow` crate. It would be beneficial to have an optional feature that allows the error module to seamlessly convert from `anyhow::Error`. I can create a PR to implement this if this feature is okay.

Yunuuuu:
Hi extendr-team! Thanks for the update! The ability to add custom error types sounds like a fantastic improvement. I’ve proposed the PR and really appreciate your willingness to review it thoroughly. I’ve also run the tests on my machine, and everything works well.

Yunuuuu:
Thanks, that’s awesome! I’ve tested it, but when I enable the `"result_list"` feature, it doesn't work as expected.

Compiling the package gave me following error:
```
error[E0277]: the trait bound `extendr_api::Robj: From<Result<(), anyhow::Error>>` is not satisfied
```

JosiahParry:
I'm going to close this as this feature exists in the main branch :)

PR #1032, opened 2026-01-08
feat(anyhow): add conversion support from `anyhow::Error` to `extendr_api::Error`

Yunuuuu:
fix https://github.com/extendr/extendr/issues/1031

This PR adds support for converting `anyhow::Error` to `extendr_api::Error`.

Since `anyhow::Error` doesn’t implement the `PartialEq` trait, it can’t be directly wrapped in `extendr_api::Error` by creating a new enum variant like `Error::Anyhow` (which I would have preferred, as it would allow conversion back to the original `anyhow::Error` without data loss). Instead, I’ve opted to convert the `anyhow::Error` to a string and wrap it in `Error::Other`. For the conversion to a string, I use the `{:?}` format, which fully supports `anyhow::Context` and backtrace.
<https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations>

A simple test has been added in `tests/extendrtests`, and the documentation for the feature gates has been updated to reflect the new anyhow feature.

### DIFF
--- a/extendr-api/Cargo.toml
+++ b/extendr-api/Cargo.toml
@@ -20,6 +20,7 @@ num-complex = { version = "0.4.6", optional = true }
 serde = { version = "1.0.219", features = ["derive"], optional = true }
 faer = { version = "0.20", optional = true }
 readonly = "0.2.13"
+anyhow = { version = "*", optional = true }
 
 [dev-dependencies]
 extendr-engine = { path = "../extendr-engine" }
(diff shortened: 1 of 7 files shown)


JosiahParry:
Is there a reason why the conversion is necessary? You can currently return -> anyhow::Result

CGMossa:
I need to review this PR specifically. There are some thing missing..
Please give me a few days.

Yunuuuu:
> I need to review this PR specifically. There are some thing missing.. Please give me a few days.

@CGMossa Thanks for letting me know! I appreciate your time and will await your review. Feel free to reach out if you need any further information or clarification.

PR #1047, opened 2026-02-26
wrappers as string constants

kbvernon:
## **Overview**
This is an attempt to implement a wrapper strategy @CGMossa has developed in his miniextendr branch. The basic strategy is this: 

1. Use the `#[extendr]` proc-macro to generate string constants that have the full wrappers in them. The logic here will depend on whether it's extendr-function or extendr-impl.
2. Use `extendr_module!()` to collect those wrapper constants and store them in modules.
3. In Makevars, run a small Rust binary that `use`s those modules, processes their string values, and writes them to `src/extendr-wrappers.R`.
4. In Makevars, compare `R/extendr-wrappers.R` to `src/extendr-wrappers.R`. If there are changes, overwrite `R/extendr-wrappers.R`.

Only 1. and 2. should require updates to the proc-macros in extendr - hence this pr.

I believe that 3. and 4. are all scaffolding that can be generated by `rextendr::use_extendr()`. We would generate a `.rs` file maybe `document.rs`, store it in `src/rust` alongside `Cargo.toml`, which would also get a `[[bin]]` to, so `cargo run --bin document` is actually what gets called in Makevars.

The primary advantages of this strategy are

1. No more dynamic loading in rextendr. In fact, `rextendr::document()` may be superfluous with these changes.
2. Following 1., improved developer experience as updates will be incremental and handled by cargo.
3. As such, this should be considered part of the extendr-push, currently supported by the R Consortium, to defer to cargo as much as possible.

## **What is in this pr?**
Four files are changed here:

- `wrappers.rs` - changed `wrappers: &mut Vec<ItemFn>` to `wrappers: Vec<syn::Item>` so the output can hold both functions and consts, and added logic to emit a `pub const R_WRAPPER_<FN>: &str` for each standalone function.
- `extendr_function.rs` - updated the local wrappers vec declaration from `Vec<ItemFn>` to `Vec<syn::Item>` to match the new signature. 
- `extendr_impl.rs` - also gains the logic to emit a `pub const R_WRAPPERS_IMPL_<CLASS>: &str` containing the full env-based class wrapper (new.env, method assignments, and $.Class boilerplate) for each extendr-impl.
- `extendr_module.rs` - added two new compile-time consts to the `extendr_module!` output: `R_WRAPPERS_<MOD>: &[&str]` aggregating the per-function wrapper consts and `R_WRAPPERS_IMPLS_<MOD>: &[&str]` aggregating the per-impl wrapper consts.

## **Reviewer considerations**
Most of these changes are surprisingly minimal. The one exception is the string processing in `extendr_impl.rs`. There are currently a lot of `let` assignments to hold temporary strings, which suggests that there may be a more elegant solution. The other thing to note is that this should not affect extendr's current behavior in any way. It only adds output from the macro expansion.


extendr-macros/src/extendr_impl.rs
```diff
@@ -378,6 +379,96 @@ pub(crate) fn extendr_impl(
...
+            let mname = {
+                let s = method.sig.ident.to_string();
+                s.strip_prefix("r#").unwrap_or(&s).to_string()
+            };
+            let r_name_r = if mname.starts_with('_') {
+                format!("`{}`", mname)
+            } else {
+                mname.clone()
+            };
+            let wrap = format!("wrap__{}__{}", self_ty_name, mname);
```
CGMossa: you ought to use the wrap__ `const` that is stored somewhere

PR #1055, opened 2026-03-20
fix: rethrow panic without prepending unwrap message

### LINKED ISSUE #1054: bug: anyhow returns traceback
See https://github.com/extendr/extendr/commit/5563f11c5622b4e60f4f5a1814700e826adc9470

> Hi, it looks like this was later reverted (in 5563f11). I'm confused as to what the recommendation is to have clean errors come through on the R side. 
> 
> Ideally we have a rust function returning Result, which translates to an error in R. I've read through #278, and tried with/without result_condition feature. It seems the wrappers won't automatically call stop for you when using result_condition, so you then need to wrap every call-site with a condition check. Not sure if I'm missing something?
> 
> This change made sense to me - is there some way to regain the ability to have clean errors now?
> 
> Thanks! 

 _Originally posted by @aquasync in [#72](https://github.com/extendr/extendr.github.io/issues/72#issuecomment-4095907882)_

JosiahParry:
Partially closes: https://github.com/extendr/extendr/issues/1054

Previously we were calling `.unwrap().into()` instead, we match on the result and don't unwrap but instead panic with the error message itself.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,16 +9,6 @@ members = [
 
 [workspace.package]
 version = "0.8.1"
-authors = [
-  "andy-thomason <andy@andythomason.com>",
-  "Thomas Down",
-  "Mossa Merhi Reimert <mossa@sund.ku.dk>",
-  "Josiah Parry <josiah.parry@gmail.com>",
-  "Claus O. Wilke <wilke@austin.utexas.edu>",
-  "Hiroaki Yutani",
-  "Ilia A. Kosenkov <ilia.kosenkov@outlook.com>",
-  "Michael Milton <michael.r.milton@gmail.com>",
-]
 edition = "2021"
 rust-version = "1.71"
 repository = "https://github.com/extendr/extendr"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ authors = [
   "Michael Milton <michael.r.milton@gmail.com>",
 ]
 edition = "2021"
-rust-version = "1.68"
+rust-version = "1.71"
 repository = "https://github.com/extendr/extendr"
 license = "MIT"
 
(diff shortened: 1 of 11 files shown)


extendr-api/src/robj/into_robj.rs
```diff
@@ -68,7 +68,10 @@ where
     E: std::fmt::Debug + std::fmt::Display,
 {
     fn from(res: std::result::Result<T, E>) -> Self {
-        res.unwrap().into()
+        match res {
+            Ok(val) => val.into(),
+            Err(err) => panic!("{}", err),
+        }
```
JosiahParry, line 74: @CGMossa this is the only change we needed to get your beautiful handling

Cargo.toml
```diff
@@ -20,7 +20,7 @@ authors = [
   "Michael Milton <michael.r.milton@gmail.com>",
 ]
 edition = "2021"
-rust-version = "1.68"
+rust-version = "1.71"
```
JosiahParry, line 13: @CGMossa bumping MSRV

JosiahParry:
@CGMossa when you get a chance: 

This PR changes the panic message, bumps MSRV, fixes compiler error for non-API entrypoints, and removes deprecated `authors` field from Cargo.toml

CGMossa (approved):

CGMossa:
Merging this but I am certain I already did something that fixes this more appropriately. I'll need to follow this up.

JosiahParry:
Your fix is already merged. The issue is twofold: 

1. it requires entrypoint.c registration whcih was added to extendrtests but not to the rextendr templates so no packages got the update
2. the unwrap error messages forced added a `"... on Result::unwrap()"` 


This just changes the message by creating a new panic that doesn't come from the .unwrap() instead it is called directly using the e.to_string() message making it _prettier_ 💅🏼

PR #1082, opened 2026-04-29
feat: create and return condition objects 

### LINKED ISSUE #1080: feat: return condition objects
Condition objects are useful and can be created in base R using a variety of: 

- simpleCondition / simpleMessage / simpleWarning /  simpleError
- errorCondition / warningCondition

Then R lang also provides:

- cnd()
- error_cnd()
- warning_cnd()
- message_cnd()

These provide a way to **create** condition objects. rlang emphasizes adding custom classes **and** tracebacks. 

I think, for an MVP we can focus on support: 

- message
- class
- call

The traceback is quite involved and i think for "expert use only". 

I suggest we introduce a builder approach creating a condition struct. 

In my opinion, I think we should try and have a Rust-first representation of the condition that we then turn into an R type. These should be small so the overhead should be minimal. I'm torn between these different representations. 

```rust
// this is transparent fro a real Condition object
struct Condition(List)

// this stores R native types but is converted to R list during From<Condition> for Robj
struct Condition {
  message: Strings
  class: Strings,
  call: Option<Environment>
}

// pure Rust structure
// enum to discriminate between condition types
enum Condition {
   Condition,
    Message,
    Warning,
    Error,
}

// this gets turned into an List on conversion to Robj
struct RCondition {
  message: Vec<String>,
  class: Option<Vec<String>>,
  call: Option<Environment>,
  condition: Condition
}
```

Most importantly, I think we should have a builder type experience that lets us build these like so: 

```rust
  let cnd = ConditionBuilder::new()
      .set_condition(Condition::Message)
      .set_message("this is a custom message")
      .set_class(["class1", "class2"])
      .build();
```

## Out of scope

This issue is for **designing** the struct / experience for **creating** condition objects. This does not include how to **signal** them.

There are 3 different scenarios we must address:

1. **returning** condition objects (the focus of this issue)
2. early return from a function with a custom error condition (the intent of newly added `abort!()`)
3. Signal a non-error condition without an early return **and** continue computation

JosiahParry:
## Summary

  Adds Rust-native types for creating, inspecting, and round-tripping R
  condition objects, with no dependency on rlang.

  Four new types in `extendr_api::conditions`:

  - `ConditionKind`: enum discriminating `condition`, `message`, `warning`,  `error`
  - `Condition`: Rust-native struct with `String`/`Vec` fields. The primary  working type.
  - `RCondition(List)`:  validated R condition at the boundary. Use as  function argument/return type.
  - `ConditionBuilder`: builder for `Condition` via  `ConditionBuilder::default()`

Condition objects support `rlang` condition objects out of the box via the fields: `message`, `class`, `call`, `parent`, and `trace`. rlang conditions (`error_cnd`, `warning_cnd`, `message_cnd`) are round tripped correctly maintaining `trace` and `parent`.

Also adds `From<Strings> for Vec<String>` and `From<Strings> for StrIter` for modest quality of life enhancement.

Closes #1080 

## Example

```rust
use extendr_api::{prelude::*, conditions::{Condition, RCondition}};

#[extendr]
fn roundtrip_condition(x: RCondition) -> RCondition {
    let cnd = Condition::try_from(x).expect("failed to parse condition");
    RCondition::from(cnd)
}

 let cnd = ConditionBuilder::default()
      .set_kind(ConditionKind::Error)
      .set_message(["bad thing"])
      .set_class(["my_err"])
      .build();
  let rcnd = RCondition::from(cnd.clone());
  let c2 = Condition::try_from(rcnd)?;
  assert_eq!(cnd, c2);
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Added
 
+- `Condition`, `RCondition`, `ConditionKind`, and `ConditionBuilder` in `extendr_api::conditions` for creating, converting, and round-tripping R condition objects. `Condition` is the Rust-native representation; `RCondition` is the R-boundary type wrapping a validated `List`. Conversions between all four types are provided via `From`/`TryFrom`. <https://github.com/extendr/extendr/pull/1082>
+- `From<Strings> for Vec<String>` and `From<Strings> for StrIter` conversions. <https://github.com/extendr/extendr/pull/1082>
 - `stop()` is an alias for `throw_r_error()` 
 - `abort!()` and `warn!()` macros for rlang-style error and warning conditions with no rlang dependency. Supports optional body bullets and caller environment attribution: `abort!("msg", &["detail"], call = env)` and `abort!("msg", call = env)`. <https://github.com/extendr/extendr/pull/1075>
 - `Environment::call()` retrieves the call associated with an environment by matching against the R call stack, mirroring `rlang::frame_call()` <https://github.com/extendr/extendr/pull/1075>
--- a/extendr-api/src/conditions/mod.rs
+++ b/extendr-api/src/conditions/mod.rs
@@ -1,3 +1,195 @@
+use crate::{
+    AsStrIter, Attributes, Environment, Error, IntoRobj, Language, List, Operators, Result, Robj,
+    Rstr, StrIter, Strings,
+};
+
+/// Discriminates the kind of R condition being constructed.
+#[derive(Copy, Debug, Clone, PartialEq)]
+pub enum ConditionKind {
+    Condition,
+    Message,
+    Warning,
+    Error,
+}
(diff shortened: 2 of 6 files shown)


JosiahParry:
I think a natural follow up to this is an Error variant that is an `RCondition` that, when present, is signalled. Doing so would like folks have a workflow roughly like: 

```rust
fn fx() -> Result<()> {

  let Ok(v) = x else {
    return Err(ConditionBuilder::default().set_condition(ConditionKind::Error).build());
  }

}
```

CGMossa (approved):
this is a work in progress; so we gotta give it some air before it can really be evaluated.

But for starts, for this to be used in `Result`, it would be better that it was a trait, and that trait could somehow build itself, through the wrappers.

PR #1098, opened 2026-06-26
feat: implement From<Option<&str>> for Rstr

### LINKED ISSUE #1096: feat: impl From<Option<&str>> for Rstr
repro:

```rust
Rstr::from(Some("value"))
```

We should be able to convert from an Option here to Rstr in every case. None -> NA

brancengregory:
## Summary

feat: implement From<Option<&str>> for Rstr

Get the value or NA string out of an Option<&str>

## Checklist

- [x] format via `just fmt`
- [x] tested with `just test`
- [] integration tests added (new features only)
- [x] CHANGELOG.md updated

## Resolves

- Closes #1096

### DIFF
--- a/extendr-api/src/wrapper/rstr.rs
+++ b/extendr-api/src/wrapper/rstr.rs
@@ -125,6 +125,16 @@ impl From<Option<String>> for Rstr {
     }
 }
 
+impl From<Option<&str>> for Rstr {
+    fn from(value: Option<&str>) -> Self {
+        if let Some(string_ref) = value {
+            Self::from(string_ref)
+        } else {
+            Self { robj: na_string() }
+        }
+    }
+}
+
 impl Deref for Rstr {
     type Target = str;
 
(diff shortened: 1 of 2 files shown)


extendr-api/src/wrapper/rstr.rs
```diff
@@ -125,6 +125,16 @@ impl From<Option<String>> for Rstr {
     }
 }
```
kbvernon, line 126: Do you mind changing this to `Self::from(value.as_deref())`? It'll get forwarded to your new impl that way.

JosiahParry (approved):
Nice and quick! Thank you!

kbvernon (dismissed):
Thanks, brancen! Looks great, just that one small edit, please.

extendr-api/src/wrapper/rstr.rs
kbvernon: Actually, scratch that. Let's not worry about it.

brancengregory:
Yay, thanks y'all!

PR #1100, opened 2026-06-27
refactor: deprecate parse in favor of Expressions::from_str

### LINKED ISSUE #1081: refactor: replace `parse()` with `impl FromStr for Expressions`
instead of the current `parse()` function we should replace it with `Expressions::from_str()` which would use a std lib trait `FromStr`. This is more idiomatic and modern rust which we should be aiming for. 

The code would be: 

```rust
impl FromStr for Expressions {
    type Err = Error;

    fn from_str(code: &str) -> Result<Expressions> {
        single_threaded(|| unsafe {
            use extendr_ffi::{ParseStatus, R_NilValue, R_ParseVector};
            let mut status = ParseStatus::PARSE_NULL;
            let status_ptr = (&mut status) as *mut _;
            let codeobj: Robj = code.into();
            let parsed = Robj::from_sexp(R_ParseVector(codeobj.get(), -1, status_ptr, R_NilValue));
            match status {
                ParseStatus::PARSE_OK => parsed.try_into(),
                _ => Err(Error::ParseError {
                    status,
                    code: code.into(),
                }),
            }
        })
    }
}
```

Then `parse()` becomes: 

```rust
pub fn parse(code: &str) -> Result<Expressions> {
    Expressions::from_str(code)
}
```

and we need a deprecation warning on this too.

brancengregory:
## Summary

refactor: deprecate parse function in favor of Expressions::from_str

## Checklist

- [x] format via `just fmt`
- [x] tested with `just test`
- [ ] integration tests added (new features only)
- [x] CHANGELOG.md updated

## Resolves

- Closes #1081

### DIFF
--- a/extendr-api/src/prelude.rs
+++ b/extendr-api/src/prelude.rs
@@ -4,10 +4,11 @@
 //! using deprecated features.
 
 pub use super::error::Error;
+pub use super::functions::parse;
 pub use super::functions::{
     base_namespace, blank_scalar_string, blank_string, eval_string, eval_string_with_params,
     find_namespace, find_namespaced_function, global_function, na_string, new_env, nil_value,
-    parse, srcref,
+    srcref,
 };
 pub use super::{
     print_r_error, print_r_output, CanBeNA, Rtype, FALSE, NA_INTEGER, NA_LOGICAL, NA_REAL,
@@ -58,6 +59,7 @@ pub use extendr_macros::{
     call, extendr, extendr_module, pairlist, IntoDataFrameRow, IntoList, Rraw, TryFromList, R,
 };
 pub use std::convert::{TryFrom, TryInto};
+pub use std::str::FromStr;
 
 #[cfg(feature = "ndarray")]
 pub use ::ndarray;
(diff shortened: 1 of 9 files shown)


brancengregory:
FYI, I made the deprecation `since = "0.10.0"` to match other deprecations even though this is on the 1.0 milestone

extendr-api/src/prelude.rs
```diff
@@ -4,10 +4,12 @@
 //! using deprecated features.
 
 pub use super::error::Error;
+#[allow(deprecated)]
+pub use super::functions::parse;
```
JosiahParry: Lets delete this. There should be no `allow()` here.

JosiahParry (changes requested):
@brancengregory thank you sooooo much! I have one tiny little nit—can you remove the allow(deprecated) clippy lint?

brancengregory:
Yep on it!

brancengregory:
@JosiahParry Done!

JosiahParry (approved):

PR #1104, opened 2026-07-02
feat: add matrix iteration methods

### LINKED ISSUE #738: Create `row_iter()` and `col_iter()` methods for `RMatrix`
Right now I am working through how to iterate through the rows of an `n x 2` matrix to create point structs from each row.

It would be very nice if I could use something like `mat.row_iter()` then this would create an iterator of `Doubles` that I can further iterate through. Or alternatively, if I needed to iterate through columns use something like `mat.col_iter()`. It is _very_ difficult to work with matrixes as of today.

brancengregory:
## Summary

Goal is to make it easier to iterate over RMatrix rows or columns.

Iterate over rows or columns:

```
let matrix = RMatrix::new_matrix(3, 2, |r, c| [
    [1., 2., 3.],
    [4., 5., 6.]][c][r]);

for row in matrix.row_iter() {
    println!("{:?}", row.data()); // [1, 4], [2, 5], [3, 6]
}

for col in matrix.col_iter() {
    println!("{:?}", col.data()); // [1, 2, 3], [4, 5, 6]
}
```

Convert rows directly to Doubles and collect. For instance collecting a 2-column matrix into points by row:

```
let points: Vec<Point> = matrix
    .row_iter()
    .map(Doubles::from)
    .map(|row| Point::new(row.elt(0).0, row.elt(1).0))
    .collect();
```

String matrices work the same way:

```
let rows: Vec<Strings> = str_matrix.row_iter().map(Strings::from).collect();
```

Iterate over the elements of each row/column:

```
for row in matrix.row_iter() {
    let sum: f64 = row.iter().sum();
    println!("{sum}");
}
```

Rows and columns are returned as owned RColumn copies.

## Checklist

- [x] format via `just fmt`
- [x] tested with `just test`
- [ ] integration tests added (new features only)
- [x] CHANGELOG.md updated

## Resolves

- Closes #738

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@
 - Added `#write_make_module_wrappers()` to `extendr_module!` macro <https://github.com/extendr/extendr/pull/1092>
   This is part of the process of moving `document.rs` into `document.c` to avoid
   issues with compiling that binary into target, see https://github.com/extendr/extendr/issues/1087.
+- Added `RMatrix::row_iter()` and `RMatrix::col_iter()` to iterate over rows and columns as `RColumn<T>`, plus `RColumn::iter()` and `RColumn::iter_mut()` <https://github.com/extendr/extendr/pull/1104>
+- Added `From<RColumn<T>>` conversions to the typed vector wrappers (`Doubles`, `Integers`, `Logicals`, `Complexes`, `Strings`, `Raw`) for their corresponding scalar types
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


JosiahParry (changes requested):
This is awesome! @brancengregory could you mind making a few doc commentsthat indicate that these methods are experimental for the time being? @CGMossa and thoughts on this? seems pretty inoccuous and its one of the first things i asked for as an exxtendr user

PR #1106, opened 2026-07-20
feat(macros): support item-level `#[cfg(...)]` in `extendr_module!`

fxdmhtt:
`extendr_module!` previously rejected any attribute on its `mod`/`fn`/ `impl`/`use` entries, because the custom parser only accepted those four keywords. This made it impossible to conditionally compile an individual exported function: gating the `fn` definition with `#[cfg(...)]` while leaving the `extendr_module! { fn foo; }` entry unconditional caused the generated metadata call (`meta__foo(&mut functions);`) to reference a function that no longer exists, breaking the build.

Parse optional outer attributes before each entry via `Attribute::parse_outer` and forward them to the generated metadata statement for `fn`/`impl`/`use` entries, so a cfg'd-out item and its metadata call are compiled out together. Attributes on `mod` are rejected, since gating the module name is meaningless.

Add a regression test using `cfg(test)` / `cfg(not(test))` to deterministically verify that a gated `fn` entry is included/excluded in the module metadata without relying on a Cargo feature flag.

## Summary

`extendr_module!` previously rejected any attribute on its `mod` / `fn` /
`impl` / `use` entries: the custom parser only accepted those four keywords, so
a leading `#` produced `expected mod, fn or impl`.

This made it impossible to conditionally compile an individual exported
function. Gating the `fn` definition with `#[cfg(...)]` while leaving the
`extendr_module! { fn foo; }` entry unconditional causes the generated metadata
call (`meta__foo(&mut functions);`) to reference a function that no longer
exists, breaking the build under that cfg.

## Changes

- Parse optional outer attributes before each entry via
  `Attribute::parse_outer`, and forward them to the generated metadata
  statement for `fn` / `impl` / `use` entries. A cfg'd-out item and its
  metadata call are now compiled out together.
- Reject attributes on `mod`, since gating the module name is meaningless.
- Add a regression test using `cfg(test)` / `cfg(not(test))` to
  deterministically verify a gated `fn` entry is included/excluded in the
  module metadata, without relying on a Cargo feature flag.

## Testing

- `cargo test -p extendr-api --features tests --target x86_64-pc-windows-gnu`
  — full suite passes (200+ tests, 0 failed), including the new
  `cfg_gated_fn_entries_respect_attributes` test.
- Existing `mod` / `fn` / `use` entries in `extendr_module_tests.rs` continue
  to compile unchanged (backward compatible).

## Motivation

Enables gating an individual R-exported function behind a feature (e.g. an
AHRS helper that needs an optional `nalgebra` dependency) while keeping the
rest of the module's bindings available when that feature is off.

### DIFF
--- a/extendr-api/tests/extendr_module_tests.rs
+++ b/extendr-api/tests/extendr_module_tests.rs
@@ -39,3 +39,48 @@ mod adjacent_module {
         fn foo;
     }
 }
+
+/// Regression test: `extendr_module!` must accept item-level `#[cfg(...)]`
+/// attributes on `fn` entries, and the generated metadata call must carry the
+/// same attribute so that a cfg'd-out function does not leave a dangling
+/// reference to a non-existent `meta__` helper.
+///
+/// `cfg(test)` is always true here and `cfg(not(test))` always false, so the
+/// assertions are deterministic without relying on a Cargo feature flag.
+mod cfg_gated_module {
+    use super::*;
+
+    #[cfg(test)]
+    #[extendr]
+    fn included_fn() {}
+
+    #[cfg(not(test))]
+    #[extendr]
+    fn excluded_fn() {}
+
+    extendr_module! {
+        mod cfg_gated_module;
+        #[cfg(test)]
+        fn included_fn;
+        #[cfg(not(test))]
+        fn excluded_fn;
+    }
+
+    #[test]
+    fn cfg_gated_fn_entries_respect_attributes() {
+        let names: Vec<&str> = get_cfg_gated_module_metadata()
+            .functions
+            .iter()
+            .map(|f| f.rust_name)
+            .collect();
+
+        assert!(
+            names.contains(&"included_fn"),
+            "cfg(test) fn should be present, got {names:?}"
+        );
+        assert!(
+            !names.contains(&"excluded_fn"),
+            "cfg(not(test)) fn should be compiled out, got {names:?}"
+        );
+    }
+}
--- a/extendr-macros/src/extendr_module.rs
+++ b/extendr-macros/src/extendr_module.rs
@@ -1,7 +1,7 @@
 use crate::wrappers;
 use proc_macro::TokenStream;
 use quote::{format_ident, quote};
-use syn::{parse::ParseStream, parse_macro_input, Ident, Token, Type};
+use syn::{parse::ParseStream, parse_macro_input, Attribute, Ident, Token, Type};
 
 pub fn extendr_module(item: TokenStream) -> TokenStream {
     let module = parse_macro_input!(item as Module);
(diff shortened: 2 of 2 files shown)


extendr-macros/src/extendr_module.rs
```diff
@@ -28,16 +28,25 @@ pub fn extendr_module(item: TokenStream) -> TokenStream {
...
-        .iter()
-        .map(|id| format_ident!("{}{}", wrappers::META_PREFIX, wrappers::type_name(id)));
-    let usemetanames = usenames
-        .iter()
-        .map(|id| format_ident!("get_{}_metadata", id))
-        .collect::<Vec<Ident>>();
+    // Each statement is prefixed with its item's `#[cfg(...)]` (and other)
+    // attributes, so that `cfg`-gating an item in `extendr_module!` compiles out
+    // both the item and the metadata call that references it.
+    let fnmeta_stmts = fnnames.iter().map(|(attrs, id)| {
```
JosiahParry, line 34:
```suggestion
    let fnmeta_stmts = fnnames.iter().map(|(attrs, id)| {
```

extendr-macros/src/extendr_module.rs
```diff
@@ -182,18 +191,25 @@ impl syn::parse::Parse for Module {
             usenames: Vec::new(),
         };
         while !input.is_empty() {
+            // Outer attributes (e.g. `#[cfg(...)]`) preceding an item are
+            // captured and forwarded to the generated metadata call, so that
+            // `cfg`-gating an item here matches the item's own compilation.
```
JosiahParry, line 196:
```suggestion
```

extendr-macros/src/extendr_module.rs
```diff
@@ -166,9 +175,9 @@ pub fn extendr_module(item: TokenStream) -> TokenStream {
 #[derive(Debug)]
 struct Module {
     modname: Option<Ident>,
-    fnnames: Vec<Ident>,
-    implnames: Vec<Type>,
-    usenames: Vec<Ident>,
+    fnnames: Vec<(Vec<Attribute>, Ident)>,
+    implnames: Vec<(Vec<Attribute>, Type)>,
+    usenames: Vec<(Vec<Attribute>, Ident)>,
```
JosiahParry, line 180: This is a code smell. We should never return a vector of tuples. Instead this should be a struct

extendr-macros/src/extendr_module.rs
```diff
@@ -47,12 +56,12 @@ pub fn extendr_module(item: TokenStream) -> TokenStream {
             let mut impls = Vec::new();
 
             // Pushes metadata (eg. extendr_api::metadata::Func) to functions and impl vectors.
-            #( #fnmetanames(&mut functions); )*
-            #( #implmetanames(&mut impls); )*
+            #( #fnmeta_stmts )*
+            #( #implmeta_stmts )*
```
JosiahParry, line 60: I'm worried that this is a regression at this no longer takes mut? or is that handled elsewhere?

extendr-api/tests/extendr_module_tests.rs
```diff
@@ -39,3 +39,48 @@ mod adjacent_module {
...
     }
 }
+
+/// Regression test: `extendr_module!` must accept item-level `#[cfg(...)]`
+/// attributes on `fn` entries, and the generated metadata call must carry the
+/// same attribute so that a cfg'd-out function does not leave a dangling
+/// reference to a non-existent `meta__` helper.
+///
+/// `cfg(test)` is always true here and `cfg(not(test))` always false, so the
+/// assertions are deterministic without relying on a Cargo feature flag.
```
JosiahParry, line 49: Please replace with human written, concise comments here. 7 lines of comments is less useful than 0 lines

JosiahParry (changes requested):
Thank you! This seems like a genuinely useful feature. 

Please follow the PR format next time and put the the AI generated summary within the details. 

For translation, it is entirely okay to use an LLM, we'd still like for your own words to come through <3

CGMossa:
Josiah, please beware that this does not work "all the way through", as the binding generation also needs to incorporate cfg logic before it works.

This is nice work to make `cfg` acceptable in the extendr_module! logic, but it ought to be clear, that there are serious scaffolding work on the R side needed to use this.

fxdmhtt:
Let me start by briefly introducing my background.

I am an algorithm engineer, though I am not particularly familiar with Rust.

I am currently maintaining an algorithm written in Rust; it is not a package created via `rextendr` extensions but exists as a completely independent project. However, I directly accessed Rust's exported functions in R to facilitate my research and analysis.

The entire workflow has presented a few challenges, such as:
1.  **Heavy reliance on environment variables.** I currently have to hardcode specific variables in multiple places just to get `cargo`, `VS Code`, and `rust-analyzer` to compile `extendr` correctly and **independently**.
2.  **Insufficient support for conditional compilation.** This is the issue the PR attempts to address.
3.  **Wrapping Rust DLL calls.** I have to manually invoke `rextendr:::make_wrappers`.
4.  **Pitfalls associated with concurrent calls** to exported Rust DLL functions (pure functions).

And so on...

I'm not sure what the best practices or necessary paths are for other problems, but I am certain that it is reasonable to allow the extendr_module macro to support #[cfg(...)]. That was my motivation for submitting this PR. In my scenario, it worked perfectly, since my Rust was compiled independently.

However, given my limited Rust expertise, I may not be capable of writing code that represents the absolute "best practice" for this requirement. I ask for your understanding.

JosiahParry:
> "Wrapping Rust DLL calls. I have to manually invoke rextendr:::make_wrappers."

if you're doing this something has gone drastically wrong. Typically relying on claude with extendr to fix compilation errors leads to edits in your makevars and other build files which should be avoided at all costs.

I'd recommend joining the discord and asking questions there! 


> "Insufficient support for conditional compilation. This is the issue the PR attempts to address."

yes, this is agreed! what @CGMossa is saying is that making these changes likely require changes not only in extendr rust source but in the rextendr R package as well :)

fxdmhtt:
Thanks for the kind reminder. Actually, when I was writing the rextendr:::make_wrappers call, I was still using old-fashioned programming methods, so this wasn't Claude's fault.

When a raw Rust DLL comes along, I definitely need some mechanism (whether handwritten or automated) to generate R bridging code, and the way I chose to do it is to dig into the specific implementation of rextendr.

PR #1107, opened 2026-07-26
Release 0.8.2: backport non-breaking fixes from the 0.9 line

CGMossa:
**TODO: replace this line with your own framing, in your own voice.**

<details>
<summary><h3>AI-written details</h3></summary>

## Summary
- Cuts a `0.8` maintenance branch from the `extendr-api-v0.8.1` tag and prepares a non-breaking 0.8.2 patch release, as requested in extendr/rextendr#531.
- Backports (cherry-picked with original authorship):
  - #950: `extendr-ffi` builds for `aarch64-pc-windows-gnullvm` (Windows ARM64), the fix Jeroen asked for.
  - #959 + #981: graphics devices report the correct graphics engine version on R >= 4.3 (GE 16) and R >= 4.6 (GE 17), and use `GECreateDevDesc` where available.
  - #973, the panic-hook registration fix, and #1055: R errors no longer carry Rust tracebacks or `unwrap` noise; `EXTENDR_BACKTRACE=1` restores the full traceback.
  - #983: serde deserialization of borrowed strings no longer panics with `unimplemented!()`.
  - #1044: `Rf_isFrame` routed through `extendr-ffi` backports for C API compliance.
  - #965: warning/clippy cleanup on current stable, `rust-version = "1.65"` declared.
  - Test hygiene already staged on `patch_release_0.8.1`: macro snapshot refresh, extendrtests `build.rs` lint fix. (The R-less build CI job and its 4.5.1-default fix were dropped per #1109's closure.)
- Bumps the workspace version to 0.8.2 and pins internal `extendr-ffi`/`extendr-macros` requirements to 0.8.2 so upgrading `extendr-api` guarantees the fixed `extendr-ffi`.
- Restores three public API items the backports had accidentally altered (caught by `cargo semver-checks`): `Robj::from_sexp` and `AltrepImpl::{dataptr, dataptr_or_null}` stay safe fns as in 0.8.1 (#965 had made them `unsafe`), and `extendr_ffi::Rf_isFrame` stays available at its old path with C ABI, now delegating to the `is_data_frame` backport (#1044 had removed it).
- Deliberately not backported: all breaking 0.9 changes (#898, #1000, #1005, #1018, #1020, #1028, #1067, #1069), additive features (#864, #871, #946, #955, #977, #985), deprecations (#952, #971, #972, #982), and the extendrtests Windows build rework (#1042).

## Test plan
[...]

PR #1108, opened 2026-07-26
fix(extendr-ffi): emit consistent cfgs and version exports when R is not found

CGMossa:
This avoids getting cargo lint warnings just because the detection of active R version failed. Does not improve much, but slightly less annoying when starting out.

<details>
<summary><h3>AI-written details</h3></summary>

## Summary
- The no-R fallback in `extendr-ffi/build.rs` set only the `r_4_5` cfg and returned early without emitting `cargo:r_version_major/minor/patch`.
- `extendr-api`'s build script reads `DEP_R_R_VERSION_*` with `.unwrap()`, so any R-less build of `extendr-api` (docs.rs, rust-analyzer on machines without R, CI without R) panicked in the build script. Items gated on `r_4_4` or the graphics engine cfgs were also inconsistent with the 4.5.1 default the fallback claims.
- The version-derived cfg flags and `DEP_R_R_VERSION_*` exports now come from a single helper used by both the detected-R path and the no-R fallback, so the two can never disagree. Note `r_4_4`/`r_4_5` are independent cfg flags, not cumulative: `r_4_5` alone does not enable `#[cfg(r_4_4)]` items such as `SEXPTYPE::OBJSXP`.
- A build without a detected R has nothing to link against, so it only serves type-checking surfaces (docs.rs, rust-analyzer, clippy CI). The fallback therefore passes the newest R version the build script knows about (currently 4.6), enabling every version-gated API rather than an arbitrary older subset.
- Drops the unused `GEfreeDD` import (unused since #981), which warns on every build where `use_r_ge_version_17` is active.
- Same fix as shipped on the 0.8 line in #1107, where the equivalent bug (there it surfaced as `SEXPTYPE::OBJSXP` not found) had the `build-no-r` job failing on every run of `patch_release_0.8.1` since July 2025.

## Test plan
- [x] `cargo build` of `tests/extendrtests/src/rust` with R removed from PATH and `R_HOME` unset reaches the fallback and no longer panics in `extendr-api`'s build script
- [x] Normal with-R `cargo check -p extendr-ffi -p extendr-api` unaffected
- [ ] CI on this PR

---
_Drafted by Claude (claude-fable-5). Reviewed by the author._

</details>

### DIFF
--- a/extendr-ffi/build.rs
+++ b/extendr-ffi/build.rs
@@ -166,6 +166,43 @@ impl InstallationPaths {
     }
 }
 
+/// Emits the R-version-derived build flags: the `rustc-cfg` flags consumed by
+/// this crate and the `DEP_R_R_VERSION_*` exports consumed by dependents such
+/// as extendr-api. Both the detected-R path and the no-R fallback go through
+/// here so the two can never disagree.
+fn emit_r_version_flags(version: &Version) {
+    // used by extendr-api
+    println!("cargo:r_version_major={}", version.major);
+    println!("cargo:r_version_minor={}", version.minor);
+    println!("cargo:r_version_patch={}", version.patch);
+
+    // Set R version specfic config flags
+    // use r_4_4 config
+    if (version.major, version.minor) >= (4, 4) {
+        println!("cargo:rustc-cfg=r_4_4")
+    }
+
+    // use r_4_5 config
+    if (version.major, version.minor) >= (4, 5) {
+        println!("cargo:rustc-cfg=r_4_5")
+    }
+
+    // Graphics engine version 15 was introduced in R 4.2
+    if (version.major, version.minor) >= (4, 2) {
+        println!("cargo:rustc-cfg=use_r_ge_version_15")
+    }
+
+    // Graphics engine version 16 was introduced in R 4.3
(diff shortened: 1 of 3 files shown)


extendr-ffi/build.rs
```diff
@@ -178,7 +178,17 @@ fn main() -> Result<(), Box<dyn Error>> {
         Ok(v) => v,
         Err(_) => {
             warn!("Cannot fetch R version from R. Defaulting to most recent configure flag");
+            // Emit the same cfg set and version exports a real R 4.5.1 would
+            // produce. Dependents like extendr-api read DEP_R_R_VERSION_* and
+            // derive their own cfgs from it, so a partial set here leads to
+            // build-script panics or mismatched cfg gating downstream.
+            println!("cargo:rustc-cfg=r_4_4");
```
JosiahParry: i believe that we dont need r_4_4 since 4_5 is already printed

CGMossa: i think you're totally right.

CGMossa: I am starting to think that the additive API features, that are resolved via R versions, should all be enabled, in the absence of a detected R version. So this might be correct by accident.

PR #1109, opened 2026-07-26
ci: build the Rust crate without R on PATH

CGMossa:
**TODO: replace this line with your own framing, in your own voice.**

<details>
<summary><h3>AI-written details</h3></summary>

## Summary
- Ports the `build-no-r` CI job from the 0.8 release line (#1107) to main: builds `tests/extendrtests/src/rust` on a runner with no R on PATH and no `R_HOME`, so the R-less build path (docs.rs, rust-analyzer without R) keeps coverage.
- Cherry-picks the `tests/extendrtests/src/rust/build.rs` fix from the 0.8 line that defaults the R version to 4.5.1 when R is not found; main's copy still panicked on `R CMD config --version`.
- Path filters cover the four extendr crates, the test crate, and the workflow file itself, since R detection lives in the crates' build scripts.
- Stacked on #1108 (the `extendr-ffi` fallback fix); without it the job cannot pass. Base is set to that branch and will retarget to main when it merges.

## Test plan
- [x] Local reproduction of the job (R stripped from PATH, `R_HOME` unset): `cargo build` of the test crate succeeds with both fixes
- [ ] `build-no-r` green on this PR

---
_Drafted by Claude (claude-fable-5). Reviewed by the author._

</details>

### DIFF
--- a/.github/workflows/test_rust.yml
+++ b/.github/workflows/test_rust.yml
@@ -1,0 +1,50 @@
+name: Rust crate build (no R allowed)
+
+on:
+  push:
+    paths:
+      - "extendr-api/**"
+      - "extendr-engine/**"
+      - "extendr-ffi/**"
+      - "extendr-macros/**"
+      - "tests/extendrtests/src/rust/**"
+      - ".github/workflows/test_rust.yml"
+  pull_request:
+    paths:
+      - "extendr-api/**"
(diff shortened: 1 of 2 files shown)


CGMossa:
I have forgotten why this was useful, but it is useful that rust crates can build without R link being established. I just cannot recall why. Closing.

